PR #58, opened 2015-03-31
build-shotcut.sh improvements

metellius:
This branch adds multiple configuration options and improvements to the build scripts

These eases daily shotcut development, both through using build-shotcut.sh directly or calling make from shotcut src directly.

Here's how you might get started with shotcut development:

Step 1: Set KEEP_BUILD_FILES, then run build-shotcut.sh once to build all dependencies and leave the build-files in place inside Shotcut.app.

Step 2: Code code code

Step 3:
For the qmake route:
invoke make && make install from ~/shotcut/src/shotcut. It will now install directly into Shotcut.app, letting you use the shotcut wrapper script. 

For the build-shotcut.sh route:
ACTION_GET_COMPILE_INSTALL=0 wil skip fetching and resetting before building, which would otherwise just throw away your changes. There is now a check in place that keeps build-shotcut.sh from discarding them if you have made changes and forgotten to set this variable.

Optional, but recommended:
SOURCES_CLEAN=0 will have it only build files that are changed
KEEP_BUILD_FILES=1 is still needed for this route to work
TAR_APP_PACKAGE=0 will skip tar'ing, which you don't really need during your development 
DISABLE_ALL_EXCEPT_SHOTCUT_AND_MLT=1 if you are only developing on shotcut/mlt, something I assume to be very likely.

For both routes:
QTDIR=/some/path/ where /some/path/{lib,include,bin/qmake} is available. This will build with another qt toolchain than the 5.2.0 toolchain that is assumed available by default.

### DIFF
--- a/scripts/build-shotcut.sh
+++ b/scripts/build-shotcut.sh
@@ -20,6 +20,7 @@ SOURCE_DIR="$INSTALL_DIR/src"
 ACTION_GET_COMPILE_INSTALL=1
 ACTION_GET_ONLY=0
 ACTION_COMPILE_INSTALL=1
+CLEANUP=1
 SOURCES_CLEAN=1
 INSTALL_AS_ROOT=0
 CREATE_STARTUP_SCRIPT=1
@@ -1975,7 +1976,7 @@ End-of-desktop-file
     cmd rm -rf Shotcut/Shotcut.app/share/man
   fi
   cmd tar -cjvf "$tarball" Shotcut
-  cmd rm -rf Shotcut
+  test "1" = "$CLEANUP" && cmd rm -rf Shotcut
   popd
 }
 
--- a/scripts/build-shotcut.sh
+++ b/scripts/build-shotcut.sh
@@ -29,6 +29,7 @@ ENABLE_FREI0R=1
 FREI0R_HEAD=1
 FREI0R_REVISION=
 ENABLE_MOVIT=1
+SUBDIRS=
 MOVIT_HEAD=1
 MOVIT_REVISION=
 LIBEPOXY_REVISION=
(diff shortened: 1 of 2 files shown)


metellius:
@ddennedy @pez4brian pinging your names because I saw in another pull request that email notifications might not be enabled by default.

scripts/build-shotcut.sh
```diff
@@ -20,10 +20,13 @@ SOURCE_DIR="$INSTALL_DIR/src"
 ACTION_GET_COMPILE_INSTALL=1
 ACTION_GET_ONLY=0
 ACTION_COMPILE_INSTALL=1
+TAR_APP_PACKAGE=1
```
pez4brian: Suggestion: Consider renaming "TAR_APP_PACKAGE" to "ARCHIVE". It is simpler and means the same thing.

scripts/build-shotcut.sh
```diff
@@ -20,10 +20,13 @@ SOURCE_DIR="$INSTALL_DIR/src"
...
+KEEP_BUILD_FILES=0
```
pez4brian: Suggestion: Consider renaming "KEEP_BUILD_FILES" to "CLEANUP". It is simpler and means the same thing. You will need to invert the logic, obviously.

scripts/build-shotcut.sh
```diff
@@ -20,10 +20,13 @@ SOURCE_DIR="$INSTALL_DIR/src"
...
 SOURCES_CLEAN=1
 INSTALL_AS_ROOT=0
 CREATE_STARTUP_SCRIPT=1
 ENABLE_FREI0R=1
+DISABLE_ALL_EXCEPT_SHOTCUT_AND_MLT=0
```
pez4brian: This seems clunky and oddly specific. How about a variable called "BUILD_PROJECTS"? The logic could be:
* If the value is empty, build all projects.
* Otherwise, the variable can contain a list of project names to build.

In the case of how you are using it now, one would specify:
BUILD_PROJECTS="shotcut melt"

This way, If I need to make changes in frei0r, I could specify:
BUILD_PROJECTS="frei0r"
and skip building everything else.

Also, I could specify:
BUILD_PROJECTS="shotcut"
when I am just developing shotcut.

scripts/build-shotcut.sh
ddennedy: Hmm, right now, TAR_APP_PACKAGE and KEEP_BUILD_FILES would be Linux-only options. This script really has startup-scripts, archiving, and cleanup co-mingled. That could be split up, but that is also a lot to change and test and a separate issue. Also, skipping the "deploy" step is of dubious value on Windows and OS X since you cannot really use them without it. Also, I have felt the current ACTION_ options are rather awkward, and perhaps all this should be:
ACTION_CLEAN_SOURCES
ACTION_GET
ACTION_COMPILE
ACTION_INSTALL
ACTION_ARCHIVE
ACTION_CLEAN_INSTALL

Now, with all that said, I think that set of changes is well outside the scope of metellius' itch. :o) So, I am inclined to accept this as-is hoping to change it soon. IOW, while I kind of agree with Brian about the name, I do not think that is worth changing the commit especially when it is due to be changed with a general cleanup anyways.

scripts/build-shotcut.sh
ddennedy: True, but per my previous comment, the renaming is not worth the effort of redoing the commit when this is going to be changed.

scripts/build-shotcut.sh
ddennedy: I  agree.

scripts/build-shotcut.sh
pez4brian: I don't feel strongly about the renaming. As the commit stands, it is an improvement.

scripts/build-shotcut.sh
metellius: I like this better, but isn't that what SUBDIRS already is? In that case, why not just allow that to be overridden?

scripts/build-shotcut.sh
metellius: I, however, do :) I'll do the renaming.

scripts/build-shotcut.sh
pez4brian: Sure. That would work. But the logic remains: if SUBDIRS is not specified (empty) then figure it out. Else use the projects specified in SUBDIRS.

metellius: Latest version of this branch does this now.

ddennedy:
Thanks, @metellius. After the 15.04 release in a day or two, I will add test these changes on all platforms and then extend some of the changes to the other platform code paths where applicable.

PR #60, opened 2015-04-05
Keyboard shortcut improvements

metellius:
This branch has many improvements centered around keyboard control in shotcut.

- selecting multiple clips is now possible, due to selection being stored in an index array (ie [0,2,3])
   - this functionality is not yet exposed in the UI
- global keyboard shortcut code is moved to MainWindow
- the playlist eating some keypresses is fixed
- the current selection can now be moved with Shift+Arrow keys
     - As a part of this, empty spaces can now also be selected, so that they may be removed with X without reaching for the mouse. This might be the most risky change, and I know of at least one unfixed bug so far (the blank clip may be opened by pressing Enter, and then appending that clip does weird stuff)
- clips may be removed using X without selecting them first
- pressing Esc clears selection

### DIFF
--- a/src/docks/filtersdock.cpp
+++ b/src/docks/filtersdock.cpp
@@ -34,7 +34,7 @@
 
 FiltersDock::FiltersDock(MetadataModel* metadataModel, AttachedFiltersModel* attachedModel, QWidget *parent) :
     QDockWidget(tr("Filters"), parent),
-    m_qview()
+    m_qview(this)
 {
     qDebug() << "begin";
     setObjectName("FiltersDock");
--- a/src/docks/filtersdock.h
+++ b/src/docks/filtersdock.h
@@ -23,6 +23,7 @@
 #include <QDockWidget>
 #include <QObject>
 #include <QQuickView>
+#include "forwardingquickviewworkaround.h"
 
 class QmlFilter;
 class QmlMetadata;
@@ -53,7 +54,7 @@ private slots:
     void resetQview();
 
 private:
-    QQuickView m_qview;
+    ForwardingQuickViewWorkaround m_qview;
 };
 
 #endif // FILTERSDOCK_H
--- a/src/docks/playlistdock.ui
+++ b/src/docks/playlistdock.ui
@@ -84,7 +84,7 @@ p, li { white-space: pre-wrap; }
          <number>0</number>
         </property>
         <item>
-         <widget class="QTableView" name="tableView">
+         <widget class="PlaylistTable" name="tableView">
           <property name="contextMenuPolicy">
            <enum>Qt::CustomContextMenu</enum>
           </property>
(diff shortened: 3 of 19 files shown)


metellius:
To be honest, I didn't spend too much time actually using these new mechanisms until submitting this pull request - I'm sure there are still tweaking to be done on the behavior. For example I'm wondering if the playhead and the selection is conflicting a bit with these changes.

ddennedy:
I am going on holiday for a week starting tomorrow and unable to test this until I return. In the meantime, please make sure your changes work correctly with undo and redo. Otherwise, I do not object to anything from the description provided.

metellius:
Thanks, good to know. I'm back to work again the day after tomorrow, and I already know about bugs that need fixing, so there's no rush.

metellius:
For the sake of reviewability, I chose to keep things in many smaller commits, and as such I recommend viewing the branch in git log -p or similar (I can't seem to find an equivalant in github, and viewing the pull request all squashed together makes it appear as the result of a very long unstructured nighttime hacking session )

ddennedy:
Do not be discourage by my comments. There are many nice things in here. Also, please add a comment header to the top of new files. Thanks

metellius:
> please add a comment header to the top of new files

What should it say? Something like this?

```
/*
 * Copyright (c) 2013-2014 Meltytech, LLC
 * Author: Harald Hvaal <harald.hvaal@gmail.com>
 *
...
```

ddennedy:
Regarding the comment header, yes, but use "2015" instead of "2013-2014" for the date since the file was added in 2015. When I modify a file, I try to remember to bump up the year if needed. I do not make a pass across the entire code base on new years day. Usually, I am trying to enjoy my holiday. :-)

ddennedy:
Another regression is that shift plus mouse-x over timeline no longer skims (scrubs) the player.

metellius:
All the regressions should be addressed now.

ddennedy:
Something is eating clicks on the blank areas of the timeline and timeline ruler and as result not seeking as expected. Drags work, but not single-clicks.

ddennedy:
About currentTrack, currentClipTrack, and resetting selection upon changing current track; I initially reacted to the change in behavior. Now, I am wondering if there is much value in keeping that behavior. I guess it's not that important and we just need to take care of the actions in TimelineToolbar.qml to use currentTrack and selection to prevent triggering asserts.

ddennedy:
I accepted the behavior change with respect to and clip selection and current track, fixed the toolbar action errors, fixed click-to-seek, and merged.

metellius:
I'm not surprised I broke the behavior because I didn't even notice it was like that originally until you explained it now.

There is still refactoring to be done on the way selections are done, but I think the first round of commits here make it less of a challenging task. Specifically selecting multiple clips should be possible, and in order to do this each selected item in the current-selection array also needs to be coupled with which track it's on. I guess it would be easier if there was an unique persistent ID (I get that pointers to producers are not suitable for this) for the clips instead of the clipIndex+trackIndex combination that is so prevalent in the function signatures.

It sounds like going down that route will again make your comments here relevant due to current track and current selection being disconnected again. Do you think it's a bad idea?

PR #61, opened 2015-04-06
Dependency module version tracking

metellius:
The exact revisions of the dependency modules are now recorded in
build-shotcut.sh, so that we can see exactly what commit went into what
release, and what changed between them.

See the comment in build-shotcut.sh for the syntax. To update these listings,
just run update-revs.sh and it will fetch, compare, update the file and commit
(but not push, of course)

After this, we can periodically run update-rev.sh, like so:

```
$ ./update-revs.sh 

libvpx: moving to v1.4.0-103-gb2baaa2 (94 new commits)
    8ea7bafdaa5785bb916106a5a47105828c7836ad..b2baaa215bb0fe3072b1088e9b2ac65a88c50d30
Inspect with: git -C /home/harald/shotcut/src/libvpx log 8ea7bafdaa5785bb916106a5a47105828c7836ad..b2baaa215bb0fe3072b1088e9b2ac65a88c50d30
[fixedRevs ddf1a2b] Pulled in new commits from dependency repoes
 1 file changed, 1 insertion(+), 1 deletion(-)

$ git show
commit ddf1a2b90b7b3f6fc7e838a9e38d7a0781217f1e
Author: Harald Hvaal <harald.hvaal@gmail.com>
Date:   Mon Apr 6 20:16:01 2015

    Pulled in new commits from dependency repoes
    
    libvpx: moving to v1.4.0-103-gb2baaa2 (94 new commits)
        8ea7bafdaa5785bb916106a5a47105828c7836ad..b2baaa215bb0fe3072b1088e9b2ac65a88c50d30

diff --git a/scripts/build-shotcut.sh b/scripts/build-shotcut.sh
index d167035..49bf4ef 100755
--- a/scripts/build-shotcut.sh
+++ b/scripts/build-shotcut.sh
@@ -409,7 +409,7 @@ function set_globals {
   git git://github.com/mltframework/mlt.git 9359202c362bef3bedcd36d3d020d92586bca8f4 FETCHFROM:origin/master v0.9.6-16-g9359202
   git git://github.com/ddennedy/frei0r.git baa08d27193d758b5139dfa503dbedc881298f1c FETCHFROM:origin/master v1.4-25-gbaa08d2
   git git://repo.or.cz/x264.git 121396c71b4907ca82301d1a529795d98daab5f8 FETCHFROM:origin/master unknown-tag
-  git http://git.chromium.org/webm/libvpx.git 8ea7bafdaa5785bb916106a5a47105828c7836ad FETCHFROM:origin/master v1.3.0-5787-g8ea7baf
+  git http://git.chromium.org/webm/libvpx.git b2baaa215bb0fe3072b1088e9b2ac65a88c50d30 FETCHFROM:origin/master v1.4.0-103-gb2baaa2
   git http://git.sesse.net/movit/ b757b4f1ca57c6dcabfe4ad696a2b887544d1a71 FETCHFROM:origin/master 1.1.3
   http-tgz http://downloads.sourceforge.net/project/lame/lame/3.99/lame-3.99.5.tar.gz lame-3.99.5
   git git://github.com/mltframework/shotcut.git a1622c54fa72413bdb5ac115af1ef2b2b7ddba16 FETCHFROM:origin/master v15.04-9-ga1622c5
```

I need some input as to what the title for these generated commits will be; I'm not satisfied with  "Pulled in new commits from dependency repoes".

@pez4brian @ddennedy

### DIFF
--- a/scripts/build-shotcut.sh
+++ b/scripts/build-shotcut.sh
@@ -26,25 +26,10 @@ SOURCES_CLEAN=1
 INSTALL_AS_ROOT=0
 CREATE_STARTUP_SCRIPT=1
 ENABLE_FREI0R=1
-FREI0R_HEAD=1
-FREI0R_REVISION=
 ENABLE_MOVIT=1
-SUBDIRS=
-MOVIT_HEAD=1
-MOVIT_REVISION=
-LIBEPOXY_REVISION=
-X264_HEAD=1
-X264_REVISION=
-X265_HEAD=1
-X265_REVISION=
-LIBVPX_HEAD=1
-LIBVPX_REVISION=
 ENABLE_LAME=1
-LIBOPUS_HEAD=1
-LIBOPUS_REVISION=
 ENABLE_SWH_PLUGINS=1
-FFMPEG_HEAD=0
-FFMPEG_REVISION="origin/release/2.3"
+SUBDIRS=
 FFMPEG_SUPPORT_H264=1
 FFMPEG_SUPPORT_H265=1
 FFMPEG_SUPPORT_LIBVPX=1
@@ -54,21 +39,41 @@ FFMPEG_SUPPORT_FAAC=0
 FFMPEG_SUPPORT_OPUS=1
 FFMPEG_ADDITIONAL_OPTIONS=
 ENABLE_VIDSTAB=1
-VIDSTAB_HEAD=1
-VIDSTAB_REVISION=
-MLT_HEAD=1
-MLT_REVISION=
-LOG_COLORS=0
-SHOTCUT_HEAD=1
-SHOTCUT_REVISION=
 ENABLE_WEBVFX=1
-WEBVFX_HEAD=1
-WEBVFX_REVISION=
+LOG_COLORS=0
 # QT_INCLUDE_DIR="$(pkg-config --variable=prefix QtCore)/include"
 QT_INCLUDE_DIR=${QTDIR:+${QTDIR}/include}
 # QT_LIB_DIR="$(pkg-config --variable=prefix QtCore)/lib"
 QT_LIB_DIR=${QTDIR:+${QTDIR}/lib}
 MLT_DISABLE_SOX=0
+
+# Revisions section
+# if XX_HEAD is set to 1, update-revs.sh will resolve the refspec
+# from FETCHFROM and assign it to XX_REVISION
+FREI0R_HEAD=1
+FREI0R_REVISION=baa08d27193d758b5139dfa503dbedc881298f1c #v1.4-25-gbaa08d2 FETCHFROM:origin/master
+MOVIT_HEAD=1
+MOVIT_REVISION=b757b4f1ca57c6dcabfe4ad696a2b887544d1a71 #1.1.3 FETCHFROM:origin/master
+LIBEPOXY_HEAD=1
+LIBEPOXY_REVISION=20062c25e7612cab023cdef44d3277ba1bd0b2de #v1.2-13-g20062c2 FETCHFROM:origin/master
+X264_HEAD=1
+X264_REVISION=121396c71b4907ca82301d1a529795d98daab5f8 #unknown-tag FETCHFROM:origin/master
+X265_HEAD=1
+X265_REVISION=5c5574c5a98f15ddf545479d80296765398f1f09 #1.6-154-g5c5574c FETCHFROM:origin/master
+LIBVPX_HEAD=1
+LIBVPX_REVISION=5e300cac2ca3c7cc628a05a41e1e270387c5a690 #v1.4.0-111-g5e300ca FETCHFROM:origin/master
+LIBOPUS_HEAD=1
+LIBOPUS_REVISION=36e0445e619a2814b05d4e48dbaee785b70aac4d #draft-ietf-codec-oggopus-06-21-g36e0445 FETCHFROM:origin/master
+VIDSTAB_HEAD=1
+VIDSTAB_REVISION=4ec5be1ba92325de8154106a1470a49242459125 #release-0.98a-14-g4ec5be1 FETCHFROM:origin/master
+MLT_HEAD=1
+MLT_REVISION=e8257fb4cea1700b6211ac28542f0d5be37427a4 #v0.9.6-17-ge8257fb FETCHFROM:origin/master
+SHOTCUT_HEAD=1
+SHOTCUT_REVISION=a1622c54fa72413bdb5ac115af1ef2b2b7ddba16 #v15.04-9-ga1622c5 FETCHFROM:origin/master
+WEBVFX_HEAD=1
+WEBVFX_REVISION=a54b0938757d53033424f411a0292926249313ba #0.4.4-44-ga54b093 FETCHFROM:origin/master
+FFMPEG_HEAD=0
+FFMPEG_REVISION="origin/release/2.3"
 
 ################################################################################
 # Location of config file - if not overriden on command line
@@ -390,7 +395,6 @@ function set_globals {
   REPOLOCS[0]="git://github.com/FFmpeg/FFmpeg.git"
   REPOLOCS[1]="git://github.com/mltframework/mlt.git"
   REPOLOCS[2]="git://github.com/ddennedy/frei0r.git"
-#  REPOLOCS[3]="git://git.videolan.org/x264.git"
   REPOLOCS[3]="git://repo.or.cz/x264.git"
   REPOLOCS[4]="http://git.chromium.org/webm/libvpx.git"
   REPOLOCS[5]="http://git.sesse.net/movit/"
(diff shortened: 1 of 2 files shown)


pez4brian:
I like the idea of being able to recreate a build with the exact same dependencies it was created with the first time. But we lose something here - and that is knowing when the latest version of a dependency has broken our build. I wonder if we should have a switch that allows one to specify that the latest version should be used for all dependencies (or maybe a particular dependency). That way I can set up an automated build that builds against the latest of each dependency every day (or week).

Another thing that we lose is the ability to specify the branch/version in the config script. It would be nice if you hadn't taken that out.

I'll let @ddennedy  make the final call on this one.

metellius:
>  I wonder if we should have a switch that allows one to specify that the latest version should be used for all dependencies (or maybe a particular dependency).

For this scenario, I then propose having the build server just run update-revs.sh && build-shotcut.sh, and then throwing away the resulting commit (alternatively making the commiting optional, if that's an issue). With a process like this it would be easy to get a build result coupled with a dependency listing showing the SHA's used, and then comparing to see which change made it fail. But to be honest I'd like to ask: is the kind of bugs you have mainly been facing the kind that a nightly build bot will discover? Wouldn't it be more realistic to brace for impact when updating all depencies, rather than never being quite sure if a buggy commit suddenly popped up as a part of that final build-shotcut.sh invocation that results in a final package? 

> Another thing that we lose is the ability to specify the branch/version in the config script. It would be nice if you hadn't taken that out.
It doesn't sound like something I took out on purpose; what are you referring to? Could you comment on it in the actually commit above? It doesn't sound like something that would conflict too much with this proposal.

I think the biggest advantage is getting complete certainty over what has changed between two releases. I had this thought early on looking at build-shotcut.sh, but what triggered this was this recent G+ post: 

https://plus.google.com/115512205788937968595/posts/1We16F8u2hG

Where the guy says that 15.04 crashes, but 15.03 works fine. The reply is "There is very little that changed that could have caused this for you.", but how can you be sure of that without actually haven gone through all the repositories, compared the git logs side by side to estimate what commits might have made it into 15.04 and what didn't? With a set of SHAs, you could generate a list of all commits that are new in the final package since last release, then walk quickly through them and select some for inclusion in the changelog.

ddennedy:
On Mon, Apr 6, 2015 at 1:29 PM, metellius <notifications@github.com> wrote:

> I wonder if we should have a switch that allows one to specify that the
> latest version should be used for all dependencies (or maybe a particular
> dependency).
>
> For this scenario, I then propose having the build server just run
> update-revs.sh && build-shotcut.sh, and then throwing away the resulting
> commit (alternatively making the commiting optional, if that's an issue).
> With a process like this it would be easy to get a build result coupled
> with a dependency listing showing the SHA's used, and then comparing to see
> which change made it fail. But to be honest I'd like to ask: is the kind of
> bugs you have mainly been facing the kind that a nightly build bot will
> discover?
>
Most of the build-related problems are detected by the MLT-MELT build.


> Wouldn't it be more realistic to brace for impact when updating all
> depencies, rather than never being quite sure if a buggy commit suddenly
> popped up as a part of that final build-shotcut.sh invocation that results
> in a final package?
>
> Another thing that we lose is the ability to specify the branch/version in
> the config script. It would be nice if you hadn't taken that out.
> It doesn't sound like something I took out on purpose; what are you
> referring to? Could you comment on it in the actually commit above? It
> doesn't sound like something that would conflict too much with this
> proposal.
>
> Please double-check this. We must be able to pin certain versions of
projects.

> I think the biggest advantage is getting complete certainty over what has
> changed between two releases.
>
Agreed


> I had this thought early on looking at build-shotcut.sh, but what
> triggered this was this recent G+ post:
>
> https://plus.google.com/115512205788937968595/posts/1We16F8u2hG
>
> Where the guy says that 15.04 crashes, but 15.03 works fine. The reply is
> "There is very little that changed that could have caused this for you.",
> but how can you be sure of that without actually haven gone through all the
> repositories, compared the git logs side by side to estimate what commits
> might have made it into 15.04 and what didn't?
>

First of all, "very little" is a subject claim. I formed that opinion based
on my review of the git logs from March for MLT and Shotcut and
understanding those changes relative to other releases. Also, I know
FFmpeg, another component in the critical path of this problem is pinned at
v2.3. Sure, something may have creeped in from somewhere else I did not
look that has a strange side effect, but I am not willing to give my most
diligent analysis to every problem report, let alone ones I cannot
reproduce. If you have a more accurate and complete list of commits, what
are you going to do with that for this problem? I am not saying this
accurate, comprehensive list of commits is not useful. I am just saying it
will not help me with this particular problem. I already have a hunch about
the problem, but I am not sure what to do with that.

metellius:
> Please double-check this. We must be able to pin certain versions of
> projects.

I'm still not seeing why this is gone at all - in my solution a dependency specification can either have a fixed SHA, in which it will never be touched, or it can be told to resolve a certain refspec each time update-revs.sh is called. This refspec can be anything - be it origin/master or certain release: origin/10.2. The big point is that there will never be uncertainty about what revisions build-shotcut.sh will  build as long as we avoid pointing to moving branches. The revisions will _only_ change when update-revs.sh is called (or manual modifications to build-shotcut.sh).

Let's take movit as an example: The way it stands now it says

```
git http://git.sesse.net/movit/ b757b4f1ca57c6dcabfe4ad696a2b887544d1a71 FETCHFROM:origin/master 1.1.3
```

It will always build b757b4f1ca57c6dcabfe4ad696a2b887544d1a71, but when update-revs.sh it will update this sha to whatever origin/master is pointing to. If we were to lock it to for example 1.1.2, we could either set FETCHFROM to this branch:

```
git http://git.sesse.net/movit/ b757b4f1ca57c6dcabfe4ad696a2b887544d1a71 FETCHFROM:origin/1.1.2 1.1.3
```
And when update-revs.sh is called next time, it would update the SHA accordingly:
```
git http://git.sesse.net/movit/ abcdabcdabcdabcdabcdabcd FETCHFROM:origin/1.1.2 1.1.2
```
The alternative is to remove FETCHFROM completely and set it like so:
```
git http://git.sesse.net/movit/ abcdabcdabcdabcdabcdabcd
```

Let me know if this addresses your needs.

scripts/build-shotcut.sh
```diff
@@ -44,7 +44,6 @@ LIBOPUS_HEAD=1
 LIBOPUS_REVISION=
 ENABLE_SWH_PLUGINS=1
 FFMPEG_HEAD=0
-FFMPEG_REVISION="origin/release/2.3"
```
pez4brian: I would like to have this back so that it matches the others above (FREI0R_REVISION, LIBEPOXY_REVISION, X264_REVISION,...)

scripts/build-shotcut.sh
```diff
@@ -386,96 +385,81 @@ function set_globals {
...
+  git git://github.com/mltframework/shotcut.git a1622c54fa72413bdb5ac115af1ef2b2b7ddba16 FETCHFROM:origin/master v15.04-9-ga1622c5
+  http-tgz http://ftp.us.debian.org/debian/pool/main/s/swh-plugins/swh-plugins_0.4.15+1.orig.tar.gz swh-plugins-0.4.15+1
+  git git://github.com/mltframework/webvfx.git a54b0938757d53033424f411a0292926249313ba FETCHFROM:origin/master 0.4.4-44-ga54b093
+  git git://github.com/georgmartius/vid.stab.git 4ec5be1ba92325de8154106a1470a49242459125 FETCHFROM:origin/master release-0.98a-14-g4ec5be1
+  git git://github.com/anholt/libepoxy.git 20062c25e7612cab023cdef44d3277ba1bd0b2de FETCHFROM:origin/master v1.2-13-g20062c2
+  git git://git.opus-codec.org/opus.git 36e0445e619a2814b05d4e48dbaee785b70aac4d FETCHFROM:origin/master draft-ietf-codec-oggopus-06-21-g36e0445
+  git https://github.com/videolan/x265 47b970f9f675addb74080c1073ef7035c8fd6e66 FETCHFROM:origin/master 1.6-126-g47b970f
+  http-tgz http://bitbucket.org/eigen/eigen/get/3.2.4.tar.gz eigen-eigen-10219c95fe65
+EOF
+
```
pez4brian: This is messy and difficult to read. Why not just populate the "XXX_REVISION" fields above. It would be cleaner and easier to read. I guess it might require some regex magic in your script to update the revisions.

scripts/build-shotcut.sh
```diff
@@ -44,7 +44,6 @@ LIBOPUS_HEAD=1
 LIBOPUS_REVISION=
```
pez4brian: In my opinion, all you would need to do to accomplish your goal would be to preload the existing "XXX_REVISION" variables. Everything else in this script could stay the same.

scripts/build-shotcut.sh
metellius: One of my hopes was to reduce redundancy by having these lines in one place; I was secretly hoping to some day get rid of the 1 to 14 magic indices spread throughout the script and use for example subfolders as keys instead.
If we were to go this way, these magic numbers would have to spread to update-revs.sh in order to fetch the repository uri (in order to find "git://github.com/mltframework/mlt.gitgit://github.com/mltframework/mlt.git" I, the script, would have to just know to grep for REPOLOCS[1]).
But I'll give it a try and we can have another look afterwards. I do want to find something we can all agree on.

metellius:
update-revs.sh will now fill COMPONENT_REVISION if COMPONENT_HEAD is set to 1. Otherwise, it will leave it alone. Running update-revs.sh now does mostly the same thing as before, but does more error checking due to the less obvious assumptions it makes about build-shotcut.sh.

metellius:
What do you think of the latest revision?

ddennedy:
This presents some workflow challenges. Because update-revs.sh:chdirToMatchingRepo() looks for git repos in ../.., it can only be run once you have first run build-shotcut.sh to create the source tree and run it from there. However, I do not tend to work out of the build script's source tree structure. Nevertheless, if I go to ~/shotcut/src/shotcut/scripts and run update-revs.sh, build-shotcut.sh now fails because shotcut's HEAD is ahead of origin. I have to set ACTION_GET_COMPILE_INSTALL=0 to proceed, but then I am not getting the new revs! I have to do some juggling to make things work. Also, the build server agents today start from a clean directory, fetch only build-shotcut.sh, and then run it. Thus, the suggestion to update-revs.sh && build-shotcut.sh will not work.

metellius:
I see. I guess we'll have to make some adjustmens/exceptions for the building-directly-from-origin/master-without-help-from-update-revs.sh case. Something like
```
if [ "$PULL_FROM_MASTER_BRANCH" = "1" -a "${MODULENAME}_HEAD" = "1" ]; then
    ${MODULENAME}_REVISION="origin/master"
fi
````

With the way update-revs.sh works it will never make sense without the whole set of repositories cloned and ready, and I don't see it as too controversial that you're required to have to run build-shotcut.sh once with at least ACTION_GET_ONLY before running update-revs.sh. But if it's a significant issue I'm sure it can also be arranged to optionally work with an explicit prefix, something like REPOSITORY_PREFIX=~/shotcut-dependencies/.

**Comment by pez4brian** [comment]
While I generally agree that it would be good to lock the build to a specific revision rather than "HEAD", this hasn't been a huge problem for us so far. I feel like update-revs.sh is an overly complicated solution relative to the small benefit that we get.

How about this for a proposal:

1. Put the script back to the way it was
2. Use the *_REVISION parameters at the top of build-shotcut.sh to specify a specific revision for each project.
3. Each month, after the release, we can manually update the revisions and make a commit.

Here is what I like about this:

1. The revisions can still be easily overridden by the user in build-shotcut.conf
2. Updating the revisions is a simple operation that can happen at the same time that we increment the version number after a release.
3. We will all spend the next one month development cycle testing the updated revisions which will increase the chances that we detect an incompatibility.

I still like the ability to specify a project to build so that I can use the script more easily for development. So I would like to see that stay.

**Comment by metellius** [comment]
> How about this for a proposal:
> 
> Put the script back to the way it was
> Use the *_REVISION parameters at the top of build-shotcut.sh to specify a specific revision for each project.
> Each month, after the release, we can manually update the revisions and make a commit.

What you're proposing here is what it already does; update-revs.sh is just a convenient way of automating this monthly update. Would you rather go manually through the 12 repositories and copypaste the SHAs into build-shotcut.sh every four weeks?

The significant changes in this pull request as of now can be summarized as:
* All the revision lines are now commited filled in
* update-revs.sh helps you update these fields, but you're free to do it manually. 
* get_subproject does some checking to see if you're tree is dirty (this could be moved to another commit) or if you're checking out a branch and not a SHA (we can remove this check if you feel it is too strict)

> I still like the ability to specify a project to build so that I can use the script more easily for development. So I would like to see that stay.

When was it away?

**Comment by pez4brian** [comment]
Ok. I see it your way now. I somewhat disagree with Dan's workflow comments because, personally, I would never use update-revs.sh for my workflow. I think update-revs should be run on some kind of routine, but not frequently by every developer. And I think that build-shotcut.sh could be committed routinely with the updated repositories. We *could* even make a robot do that (or TeamCity). For example, we could have a builder in Teamcity that runs once every week. When it runs, it would get the latest for all projects, compile, and if they all succeed, it would commit an updated build-shotcut.sh. However, I would still prefer a once-per-month system that aligns with releases.

**Comment by ddennedy** [comment]
Brian, it does not have to be a frequently used workflow for it to be a workflow. The idea to use update-revs.sh on the nightly build nodes simply does not work. To run it only periodically as part of release process has complications when trying to confirm the build prior to pushing the commit it generates. The idea to only periodically update revisions does not work for me. I require a nightly build of MLT and Shotcut because I sometimes test these interim builds after some changes, which is more convenient than rolling my own build (my time is limited, I support all the platforms, and I am the lead). I will not have MLT and Shotcut pinned for any period longer than a day. Neither do I want a nightly commit to clutter the git history. Neither will I pin revs for some short period prior to release because Shotcut translations are committed on the last day to give translators enough time prior to release (English translation is updated one week end of month).

I value agility and frequent releases more than the rigor of the repeatability of old things. Being able to download a previous build and run it is good enough for me at this time. It is not too difficult to review the git logs from relevant repos between releases. I don't want to keep going round and round discussing this. This is not really solving a problem for me.

**Comment by pez4brian** [comment]
I don't disagree. As I said previously, this hasn't been a big problem for us. The only thing I really wanted was a way to specify which project (if not all of them) should be built - since this would improve my daily workflow. 

@metellius you might want to consider submitting a separate pull request for the subprojects change. We can probably move that forward while we let the dependencies conversation simmer.

**Comment by metellius** [comment]
I'm a bit surprised at how much I failed to pick up your stance on this, I'll stop working on the revision pinning.

For the record, my main reason for wanting this was not just being able to rebuilt a certain revision, it was more about the risk introduced by throughout the month never being guaranteed to be testing the exact same thing, and an otherwise stable shotcut release could potentially be spoiled by an unforeseen buggy commit in (for example) ffmpeg on the day before you build the release.

As I'm writing this I have one final suggestion for how we can at least record how a release was built: build-shotcut.sh will include in the package a textfile containing the revisions at the time of building, something like this:
```
LIBEPOXY_REVISION=20062c25e7612cab023cdef44d3277ba1bd0b2de
X264_REVISION=121396c71b4907ca82301d1a529795d98daab5f8
X265_REVISION=5c5574c5a98f15ddf545479d80296765398f1f09
LIBVPX_REVISION=5e300cac2ca3c7cc628a05a41e1e270387c5a690
```
Which at least lets us conveniently paste that directly into the local build config if a rare bug that is only frequent in a certain release should need debugging.

metellius:
> The only thing I really wanted was a way to specify which project (if not all of them) should be built - since this would improve my daily workflow.

Isn't this already in as of ee97c21a144b35cee8ff910435ff3341f2670e19?

pez4brian:
Yes. That covers it. I am a happy camper :)

ddennedy:
To be honest, I was not sure of my stance until I had a chance to actually review it and try it out. It can be nice to pin versions of most dependencies except movit, webvfx, frei0r, mlt, and shotcut (pretty long list) - projects where I am fairly active. Then, other projects like x265 and libvpx VP9, which are still rapidly evolving, I do not want to ping and to take some risk there to give people the absolute latest. lame, ffmpeg, and swh-plugins are already pinned. That leaves vid.stab, liibepoxy, opus, and eigen. Maybe I'll just start manually managing those now.
Yes, I will consider a patch to create a git sha1 manifest. It could go into the src archive only or put into share/ of all archives.

PR #62, opened 2015-04-12
Support multitrack ripple insert

metellius:
ATM ripple insert will not touch the other tracks in the timeline, causing the tracks to misalign. This commit handles that by walking through the other tracks and
- if a blank is found, it is extended by the length of the inserted clip
- if a clip is found, it is split in two and a blank is inserted between them.

I expect the latter case to be rare. Is this how other editors do it? I have never tried ripple insert in another video editor.

Undo support is not handled in this commit, so it's obviously not ready yet. The challenge is the amount of playlist modifications that have to be undone after a ripple insert, so I would appreciate some input on how to best store the previous state in the InsertCommand class.

@pez4brian @ddennedy


ddennedy:
This behavior will need to toggled on using an option in the timeline menu like "Ripple all tracks", which must be saved in Settings across sessions. It must be accompanied by also supporting ripple delete - both with support for undo and redo. One way to save state is by using MLT XML. Obviously, snap-shotting the entire multitrack and restoring it is gross and heavy, but maybe it is the only safe way to handle a complex operation. Otherwise, massive engineering may be required, which is likely to introduce bugs.

metellius:
> One way to save state is by using MLT XML. Obviously, snap-shotting the entire multitrack and restoring it is gross and heavy, but maybe it is the only safe way to handle a complex operation. Otherwise, massive engineering may be required, which is likely to introduce bugs.

For the former case in my comment (which I'm still guessing will be more common) the only extension to what we already do today is adjusting the lengths of the blanks on the other tracks. We could say that this is supported, but for the remaining complex cases we simply store the xml.




> This behavior will need to toggled on using an option in the timeline menu like "Ripple all tracks"

Of course, but asking out of curiosity: when would you want to intentionally offset one tracks from the other ones? 

> It must be accompanied by also supporting ripple delete

Only because of having "ripple all tracks" as a separate option right? Or is there any other reason why it's a must?

ddennedy:
On Sun, Apr 12, 2015 at 1:37 PM, metellius <notifications@github.com> wrote:

> This behavior will need to toggled on using an option in the timeline menu
> like "Ripple all tracks"
>
> Of course, but asking out of curiosity: when would you want to
> intentionally offset one tracks from the other ones?
>

I have been uncertain about which way to go on this, but I have been
thinking about it for a while. I was leaning towards an option because it
is different than what we do today and because some beginners will get an
unexpected behavior. However, it can also be argued that today's behavior
is unexpected for a beginner, but I tend to disagree. I suppose one idea is
that a beginner will create unexpected damage, and they generally prefer
the footprint of that damage to be smaller than larger. Also, we do not
need to cater entirely to beginners, but it might be nice to be
considerate. Feedback welcome.


>  It must be accompanied by also supporting ripple delete
>
> Only because of having "ripple all tracks" as a separate option right? Or
> is there any other reason why it's a must?
>

They logically belong together, but they do not need to be in the same
commit. I just want someone to commit to doing delete soon after doing it
for insert, preferably for the same release. OK, I will skip this
requirement because if you don't get around to it, I will.

ddennedy:
About the use case for not wanting to ripple all tracks. It is common to add a music bed track with one ore more and then later to insert some other footage on the main video track. You do not want to split and shift the music. It can be argued to use a separate video track for B-roll cut-aways, but that is a technique that is not always obvious to people. Same thing could apply if someone makes a video track to composite a graphic across the entire video.
Also, if we do not offer to disable ripple all tracks, then we should have a join operation, but a join function should probably check that producers are equivalent (otherwise, it is should be grouped if we had that function). Also, we would need to figure out a policy for how to handle the filters attached to each producer (drop the second producer's filters and use only first producer's?).
If we add an option, should it be in the timeline menu or toolbar? I am leaning towards the menu.

tin2tin:
To ripple insert/delete over several tracks is very useful if you're working with video in a video track and  audio in an audio track in sync. In that case you would want to ripple insert new clips before that clip, so all clips in all tracks (or all selected tracks) are pushed to the right in order to keep the clips in sync. But this would demand the option to only operate in multiple, but selected tracks only for all editing operations, which might be a huge undertaking. 

An alternative route is to work with linked or grouped clips with an option under the settings to add new clips to seperate tracks eg. video into a video track and audio into an audio track and possibly empty space/blank clips in the rest of the tracks, and then have the clips grouped by that setting. That way videoclips would never loose sync with audio clips. Shooting audio separate from the video is a very common pratice with ex. DSLR camaras thanks to the poor quality of the audio recordig on the cameras. 

Here's a description of the grouping/linking/locking function of kdenlive: https://userbase.kde.org/Kdenlive/Manual/Timeline/Grouping

ddennedy:
Thanks @tin2tin. We are already very familiar with Kdenlive. Within Shotcut code base, grouping is actually more difficult/work than ripple on all tracks, and with this commit someone has actually contributed code for it. However, your message does remind me that I wanted to add track locking in conjunction with this feature.

metellius:
Newest revision now includes track locking and undohelper as well. These features, although split into separate sequential commits, make sense together, so I'm including them all in this pull request.

ddennedy:
I just realized we already have a ripple toggle button for the toolbar. I will not have 2 similar looking and sounding ripple buttons! I am sorry I led you astray by telling you to add a button. Let me think about if all ripple operations should be toggled by this single button, or whether drop should have its own insert/overwrite mode switch. Feel free to think about that too and comment.

metellius:
How about by default rippling all unlocked tracks? For the ripple single track, we could have a solo button to the track head that temporarily locks all other tracks.

ddennedy:
I will reconsider the option to always ripple all unlocked tracks. In the meantime, here is what I am thinking. Solo is mainly used for audio, even in video editors. I was thinking of adding a solo function, but I don't want to add another button to the track header. I was thinking of making a shift-click on mute, mute all other tracks (and unmute clicked track if muted). We can do the same for the lock button.
Next, regarding the ripple buttons, they really are different ripple toggles. However, I find that for the drop operation, one often needs to change the mode while in the middle of dragging (and seeing the words Overwrite or Insert appear). We can remove the ripple-drop toggle from the toolbar and instead let a keyboard modifier change the drop behavior. People are used to this when dragging an icon in a file manager to choose between copy and move. Then, the existing ripple icon would be repurposed for the ripple-all-tracks toggle. Alternatively, the ripple button toggles a more global ripple option that affects a consolidated insert/overwrite, a consolidated remove/lift, drag-n-drop, and ripple all tracks.

ddennedy:
I did some testing of the undo behavior, and I ran into some problems, but I verified at least one of these is an old bug in the current release. This cross-checking takes time, and I need to give it more testing. Also, I need to study more the UndoHelper because I need to really want to understand it now instead of later.

metellius:
> Alternatively, the ripple button toggles a more global ripple option that affects a consolidated insert/overwrite, a consolidated remove/lift, drag-n-drop

I like this option, we get to have a lot less buttons on the toolbar, and the incentive for the user for learning the keyboard modifier is greater. We could even extend it with a double keyboard modifier for the power users (but discoverable since the ripple-all-tracks button would be flashing when pressed). Ctrl: insert, Ctrl+Shift: insert all tracks

ddennedy:
That option would not require a keyboard modifier for the drop operation because the operation would be determined by the state of the ripple toggle. However, a keyboard modifier could be added as a convenient override.

ddennedy:
If you feel stalled while I continue to review the undoHelper and decide what to do about ripple toggle, then you can fix the track lock feature to suppress all actions. The new auto-select, keyboard shortcuts, and toolbar buttons are still able to affect locked tracks.

ddennedy:
I have refactored UndoHelper and fixed the memory leaks for both UndoHelper and Mlt objects in my working copy. No need for you to do it.

I have also decided on the ripple all tracks toggle button based on a review of some other tools. I am going to remove it, there will be no toggle for ripple all tracks, and instead ripple will affect all unlocked tracks. The existing remove, lift, insert, overwrite buttons will not be consolidated and toggled based on the state of the existing ripple button. When ripple support is added to trimming and moving clips, then the existing ripple button will toggle their behavior.

ddennedy:
Portions of this have been merged now plus my fixed and changes on top. You should review my commits to see what I did.

metellius:
Awesome!
Now that you know undohelper, I'd like to hear if you have any thoughts about 'temporary_undo_id'... Is it a hack? Is there a better way to identify clips before and after some operation?

PR #66, opened 2015-04-22
Show in the filterdock what is currently selected

metellius:
This is shown in a label directly above the list of filters.

If nothing is selected, "Nothing selected" is shown on top of the filters list.
This makes it less obvious why there is an opaque black box with two disabled
buttons.

Clips just show "Anonymous submission" so far, and I'm not sure what would be the best thing to show.

### DIFF
--- a/src/models/attachedfiltersmodel.cpp
+++ b/src/models/attachedfiltersmodel.cpp
@@ -80,6 +80,19 @@ void AttachedFiltersModel::setProducer(Mlt::Producer* producer)
     }
 }
 
+QString AttachedFiltersModel::trackTitle() const
+{
+    if (m_producer.isNull())
+        return QString();
+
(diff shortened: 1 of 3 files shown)


ddennedy:
Do not use the title property. The reason why the xml consumer generates it is a little complicated story and not important to discuss here. I want this label invisible when a producer other than a track head is selected. I definitely like something that says "Nothing selected," and it is educational to say "Track: ____" when a track head is selected. However, I do not want to display the name of the producer and its resource here. We show that in the properties panel, playlist item, and timeline shot. We don't need to repeat it everywhere possible! Therefore, ignore the title property and set the QML visible property based on empty string.

metellius:
Is there an easy way to see if a producer is a track? The attachedfilters model does not know that as far as I can see.

ddennedy:
shotcut:name, which you are already checking, is only used for tracks.

PR #74, opened 2015-05-24
Reject locked track modifications

metellius:


ddennedy:
Shortcuts c, v, & b (append, insert, overwrite) are not blocked.

PR #78, opened 2015-06-01
Webvfx in out point fix

### LINKED ISSUE #77: Non-fadein/out filter in/out points are not updated
Trying to use the time argument in webvfx I noticed that the in/out points were not updated when the clip was resized. I eventually tracked it down to MultitrackModel::trimClipIn (among others)
```
        // Adjust all filters that have an explicit duration.
        int n = info->producer->filter_count();
        for (int j = 0; j < n; j++) {
            Mlt::Filter* filter = info->producer->filter(j);
            if (filter && filter->is_valid() && filter->get_length() > 0) {
                if (QString(filter->get(kShotcutFilterProperty)).startsWith("fadeIn"))
                    filter->set_in_and_out(in, in + filter->get_length() - 1);
            }
            delete filter;
        }
```
This startsWith test was added in 1456ff84
So in webvfx's case, the filter in/out is only set once, in Component.onCompleted in the filter ui, and never again.

I have not yet had time to look at possible solutions, but my initial impression is that QString(filter->get(kShotcutFilterProperty)).startsWith("fadeIn") looks hackish.

metellius:
Fixes #77

### DIFF
--- a/src/mltcontroller.cpp
+++ b/src/mltcontroller.cpp
@@ -567,8 +567,10 @@ void Controller::setIn(int in)
         for (int i = 0; i < n; i++) {
             Filter* filter = m_producer->filter(i);
             if (filter && filter->is_valid() && filter->get_length() > 0) {
-                if (QString(filter->get(kShotcutFilterProperty)).startsWith("fadeIn"))
+                if (QString(filter->get(kShotcutFilterProperty)).startsWith("fadeIn")
+                        || QString(filter->get("mlt_service")) == "webvfx") {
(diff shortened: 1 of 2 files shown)


ddennedy:
Check the property mlt_service=webvfx as I asked previously. Do not add objectName to overlay HTML filter as that is the fallback for any webvfx-based filter that does qualify itself with a "shotcut:filter"

metellius:
Closed? Should I post a new pull request for the correct fixes?

ddennedy:
sorry, re-opened

metellius:
Is there anything missing for this to get merged?

PR #81, opened 2015-06-15
Fix shotcut wrapper not supporting relative paths

metellius:
This was broken when 'cd $INSTALL_DIR' was added as a workaround for another
bug.

Thus, calling something like Shotcut.app/shotcut ../../Documents/project.mlt wouldn't work.


ddennedy:
Sorry to reject this, but the change that broke it was really just a work around a bug in Qt that is fixed in newer releases. I am opposed to band-aids on band-aids. Soon, we will upgrade Qt, and we can revert the previous change and not resort to this ugliness.

PR #82, opened 2015-06-19
Fix CuteLogger target path

nfnty:


ddennedy:
Shotcut statically links CuteLogger. The install target for this sub-project should be removed.

PR #87, opened 2015-06-21
Multi-clip selection

metellius:
Ctrl-clicking a clip while another is selected will append the clip to the
current selection.

- Selections are no longer just a clipIndex, but a variantmap containing both
  trackIndex and clipIndex.
- A clipindex helper class is provided for setting and getting the trackIndex
  and clipIndex properties from the variant
- Functions operating on the selection will now have to consider multiple clips
  and multiple tracks. You can no longer make any assumptions about current
  track and current selection.
- The first clip in the selection list will always be the first clip that was
  clicked on, and if the operation only operates on a single clip, the first
  item in this list should be chosen.

Shift-clicking for selecting ranges is on my TODO list.

### DIFF
--- a/src/docks/timelinedock.cpp
+++ b/src/docks/timelinedock.cpp
@@ -246,27 +246,59 @@ void TimelineDock::makeTracksTaller()
     QMetaObject::invokeMethod(m_quickView.rootObject(), "makeTracksTaller");
 }
 
-void TimelineDock::setSelection(QList<int> newSelection)
+void TimelineDock::clearSelection()
+{
+    setSelection(QVariantList());
+}
+
+void TimelineDock::setSelection(const QVariantList & newSelection)
 {
     qDebug() << "Setting selection to" << newSelection;
     if (newSelection == selection())
         return;
 
-    QVariantList list;
-    foreach (int idx, newSelection)
-        list << QVariant::fromValue(idx);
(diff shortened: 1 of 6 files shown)


metellius:
Just realized deselecting should really be a part of this commit, will update the pull request once I get that done.

metellius:
Deselection now also included.

ddennedy:
Am I correct to understand that does not yet modify any operations to handle multi-selection? I am very busy and might take me a while to get around to reviewing this. I have not decided that I actually want to introduce this. Before going much further, please pause to give some thought about the use-cases for multi-select and possibly document them. What are some tricky things that arise? For example, think about moving multiple clips. How does it complicate the implementation for QUndoCommands? Seems like this is likely to make things complicated and difficult to keep bug-free.

metellius:
> I am very busy and might take me a while to get around to reviewing this
No worries, I'll be AFK most of July anyway so I'm not expecting anything soon.

> Am I correct to understand that does not yet modify any operations to handle multi-selection?
As of earlier commits, selection was already a list of clip indices, the difference is that now they are coupled with a trackIndex.
Remove and lift already did the selection handling correctly by iterating over the selection instead of just choosing one, and in most other cases the action is either aborted if multiple items are selected, or the first one is chosen. I have not done any significant work yet on the timeline operations yet.

I'll work on the uses cases etc when I get home.

PR #97, opened 2015-07-20
Webvfx filter ui: custom properties

metellius:
This adds a new part to the html overlay filter ui that lets the user set custom
properties, accessible from a webvfx enabled html filter. The new UI is only
visible when the webvfx javascript extension checkbox is enabled.
The list of property names is stored in a special property called
"customPropertynames", which is a comma-separated list of names.

Here's a short youtube video that shows how it may be used:

[IMAGE ALT TEXT HERE](https://youtu.be/KYyXdFfYsas)

### DIFF
--- a/src/qml/filters/webvfx/ui.qml
+++ b/src/qml/filters/webvfx/ui.qml
@@ -31,6 +31,21 @@ Rectangle {
 
     SystemPalette { id: activePalette; colorGroup: SystemPalette.Active }
     Shotcut.File { id: htmlFile }
+    Shotcut.File {
+        id: customUiFile
+        url: {
+            if (!htmlFile.url || !htmlFile.exists())
+                return "";
+            var uiFile = htmlFile.url.toString();
+            return uiFile.substr(0, uiFile.lastIndexOf(".")) + "_ui.qml";
+        }
+        onUrlChanged: {
+            if (exists())
+                customUILoader.source = url;
+            else
+                customUILoader.source = "";
+        }
+    }
 
     Component.onCompleted: {
         var resource = filter.get('resource')
@@ -214,6 +229,12 @@ Rectangle {
             Layout.columnSpan: 2
             Layout.fillWidth: true
             visible: editButton.visible
+        }
+
+        Loader {
+            id: customUILoader
+            Layout.columnSpan: 4
+            Layout.fillWidth: true
         }
 
         Item {
--- a/src/qml/modules/Shotcut/Controls/SimplePropertyUI.qml
+++ b/src/qml/modules/Shotcut/Controls/SimplePropertyUI.qml
@@ -1,0 +1,40 @@
+import QtQuick 2.1
+import QtQuick.Layouts 1.0
+import QtQuick.Controls 1.1
+
+Column {
+    id: root
+    Layout.columnSpan: 4
+    Layout.fillWidth: true
+    spacing: 4
+
+    property var properties: []
+
+    Label {
+        text: qsTr("Custom Properties")
+        font.bold: true
+    }
+
+    Repeater {
+        model: root.properties
+
+        RowLayout {
+            width: parent.width
+            height: propField.height
+            Label {
+                Layout.preferredWidth: editButton.width
+                Layout.minimumWidth: implicitWidth
+                anchors {
+                    verticalCenter: propField.verticalCenter
+                }
+                text: modelData
+            }
+            TextField {
+                id: propField
+                Layout.fillWidth: true
+                text: filter.get(modelData)
+                onTextChanged: filter.set(modelData, text)
+            }
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


ddennedy:
Just make your webvfx-based filter a separate filter. I don't want to make generic things like this.

metellius:
How do I introduce shotcut to a user-made filter? Can I make it appear in filter selection, or do I need to edit xml for each clip I want to use this filter? I hope there is a way to do this without requiring the user to start shotcut with a special environment variable.

ddennedy:
Create a folder in share/shotcut/qml/filters containing metadata and UI QML files.

metellius:
But that's in the install folder right (Shotcut.app/share/shotcut/qml/filters)? Or can I make such a path relative to the mlt project file? It makes sense to bundle the filter together with the project, or at least be able to have something like ~/.share/shotcut/qml/filters/mycustomfilter so I wouldn't have to copy stuff around for every new shotcut release.

ddennedy:
Yes, that is in the install folder. I don't care if you have to do a little work for your custom, non-shared filters. But, you can submit a patch that looks in an additional location.

Now, the other use case you raise is to add parameters to HTML that is fairly specific to the project. That is getting rather esoteric, and you are likely the only one who will use it. I believe many people do not use the HTML filter not realizing an editor is provided or simply unsure about what it is and how to use it. In any case, I can see from the video, this implementation is not good enough to merge. I guess we should discuss whether it has a chance to be merged to help you decide whether to continue working on it.

From the video, it looks like the user needs to know the names of their custom properties and enter them instead of having the UI present them. We know that with webvfx, JavaScript cannot set a MLT property to be read by this filter UI. One idea is to extend webvfx to support that. Then, you could read the property names to present a better UI or even define a QML property that contains the UI. Without that, how about a convention to look for a file with the same name but ending with .qml? That QML could define either a complete UI or, for a more lightweight solution, a special QML property name containing the MLT property names. If you just want to continue on the current direction, then for the UI why not a 3 column table with a delete column and always an empty row to add the next?

I have been thinking about a way to extend the HTML editor to work with the forthcoming keyframes timeline. It would be nice in the editor to select something, give it an id, and then the editor adds webvfx code to look for a property with the same name as its id. The generated code's render function will get the element by id and set its value from the property. The editor could inform the keyframes UI about the property when the id is set; however, what about when loading an existing file? It should generate some code that can set a special MLT property containing information about the animatable properties. There is some crossover here.

metellius:
Out of your suggestions, I think the same-name-but-qml-ending custom filter UI feature is the most promising (at least for my user facing custom html parameterization usecase). The qml file could then be a custom component that is loaded and shown as a part of the webvfx filter UI, which either defines all the fields and labels, or even simply uses a convenience inheritable item that would make the least amount of work needed a file named something like myfilter_ui.qml: "import QtQuick 2.0; SimpleCustomPropertyUI { properties: [ 'text', 'subtext', 'color' ] }"

> From the video, it looks like the user needs to know the names of their custom properties and enter them instead of having the UI present them. We know that with webvfx, JavaScript cannot set a MLT property to be read by this filter UI. One idea is to extend webvfx to support that. Then, you could read the property names to present a better UI or even define a QML property that contains the UI.

You might remember I tried this a while back by allowing webvfx to set metadata properties through the provided qobject interface. You pointed me towards MLT metadata, but I could never find out how to find a proper pointer to the metadata struct in order to share it between webvfx and shotcut. I'm not even sure that was the correct way to go about it.

The overlap you refer to is interesting: as long as the only data resource for the HTML editor/keyframe timeline/filter UI is the html file itself, metadata for it will have to be readable somehow. Maybe it doesn't even have to be through javascript at all, but rather a pragma style html comment with metadata.

metellius:
I just pushed the filtername_ui.qml implementation, see the commit message for details. I'm quite satistifed with the simplicity here - most of the complexity from the former solution was from the add/remove property buttons.

As of this commit this is a completely undiscoverable feature, how do you feel about adding a label or a tooltip that (when the user has selected and loaded myownfilter.html) explains how you can also make myownfilter_ui.qml.

ddennedy:
I get the impression from some comments people are confused about this filter, in general. I will add some very brief text that lets the user know that Shotcut includes a HTML editor, but they need to click New and create a file name to get started. Along with that text, I will include a link to the HowTo section of the site, and we can add a page about this feature - How to add parameters to your HTML. How does that sound?

I can give you a wiki account to create and edit the page, but not only must you learn the ways of this wiki but you must remember to only access it as https://shotcut.org/ to workaround the CDN. Alternatively, you can just draft something somewhere else, and I will convert it. You can include screenshots.

metellius:
> I can give you a wiki account to create and edit the page

Sure, I can type something up. The webvfx filter is very powerful, but when it involves coding outside the HTML editor I think a text-based tutorial on the wiki with examples makes sense. As long as I can rely on you looking over my grammar.

ddennedy:
When it comes to writing your article, please remember this page exists:
http://www.mltframework.org/doxygen/webvfx/effects_authoring.html
You might want to link to that page for more info except I have not been able to get the toImageData() to work. I worked at one point on Qt 4.7, I think, but I think it broke at some point in QtWebKitBridget. Maybe, I'm wrong: https://bugreports.qt.io/browse/QTBUG-38809

PR #102, opened 2015-08-06
Hint at the driver to use discrete GPU on windows

imerr:
Quoting your FAQ: "If your system is using a hybrid GPU such as NVIDIA Optimus, then you must use the control panel to force Shotcut to use the discrete GPU."
This change should inform the driver to do that automatically

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -17,6 +17,7 @@
  */
 
 #include <QtWidgets>
+#include <QtGlobal>
 #include "mainwindow.h"
 #include "settings.h"
 #include <Logger.h>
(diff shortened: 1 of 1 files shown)


ddennedy:
Thanks, but I think perhaps we only want to do this when GPU processing is enabled. Is it possible to do this "if (Settings.playerGPU())" after constructing the Application object on line 207?

imerr:
Don't think it is, the driver is probably going to load the hint as soon as the application starts
Keep in mind this is only a hint and can be overridden with the respective setting apps (and might not even work in energy saving mode - I havent tried)

PR #113, opened 2015-09-05
QTBUG47714 Workaround

metellius:


ddennedy:
This is not working on Windows. It actually makes it crash. It is at least in part due to the new signal/slot connect syntax you are using. That does not work in our Windows builds. I do not know if it is related to using MinGW or what.  In any case you will see that we are generally not using it. It gives errors in debug log such as "signal not found in QQuickView." Then, if you do not make it declare the slots as "private slots:" it will give errors such as "No such slot." After making those changes, I get a link error: "undefined reference to vtable." Adding a Q_OBJECT macro alone did not help. I had to split declare the class in the header and include the macro.

ddennedy:
Also works for me on Linux and OS X.

metellius:
If you want to have a Q_OBJECT class without including it in the header, you need to add #include "qmlview.moc" to the bottom in order to have qmake pick it up and invoke moc on it.

Too bad we have to stick to the old style connects, it's hard to notice when they fail and just print a warning. At work we use a special assert around the connect statements that upon invalid signatures asserts in debug but just leave the code as it is in release builds.

PR #132, opened 2015-10-26
Simpler focus workaround

metellius:
Because the focusObjectChanged is not emitted correctly for qquickwidget, this
    commit throws away the ForwardingQuickViewWorkaround hack and instead does it
    through a single event filter installed on the QWidgetWindow (ie. native handle

The event print mechanism that I have been using a lot for debugging where key events go is also included in this PR.


metellius:
This is broken for me in master, and fixed by this workaround:
Click a clip in timeline, then click the middle of the glwidget. Pressing left/right will not move the clip forward/backward.

Things to test:
- Click timeline (clip in qquickwidget gets focus), then click up/down (keypress should be ignored, and reach all the way up to mainwindow where it changes active track)
- Focus timeline, press up/down and check same as above. Then have shotcut lose focus by clicking on another application, then reactivate it by clicking window bar only, and check that up/down still works.

Issue #151, opened 2015-12-03
Crash when closing and restoring timeline dock


jmcgvr:
To reproduce
- open any existing mlt file that has any kind of track in the timeline. Just opening shotcut with an untitled project then adding a track doesn't seem to ever cause this.
- close the timeline dock, reopen it, crash
- this isn't consistent (but very frequent for me). Shotcut seems to be able to "settle" into a state where no number of closing and reopening the timeline dock causes a crash, including loading other mlt files.
- If it doesn't crash, the audio waveforms are not redrawn in the timeline
```
Problem signature:
  Problem Event Name:	APPCRASH
  Application Name:	shotcut.exe
  Application Version:	0.0.0.0
  Application Timestamp:	565fbaaa
  Fault Module Name:	Qt5Quick.dll
  Fault Module Version:	5.5.0.0
  Fault Module Timestamp:	559bf2b5
  Exception Code:	c0000005
  Exception Offset:	0000000000059e25
  OS Version:	6.1.7601.2.1.0.256.1
  Locale ID:	1033
  Additional Information 1:	1ef3
  Additional Information 2:	1ef3fde1435c37b31f094ed73efc4e41
  Additional Information 3:	74a4
  Additional Information 4:	74a4a12429940639514f8911e1d99c14
```

ddennedy:
I reproduced it on Windows and Linux.

ddennedy:
Crashes in QTBUG47714WorkaroundRenderListener::afterSync() on `delete oldTexture`.

metellius:
Currently working on this.

ddennedy:
Fixed in 1dd6491.
The other linked commit 27f58a3 does not apply (commit message changed just after noticing the mistake.)

PR #159, opened 2015-12-09
Move canvas painting to QQuickPaintedItem

metellius:
This commit takes a different approach on the qt >=5.5.0 canvas item bug
    that has proven very hard to workaround reliably. Instead of doing the drawing
    in canvas with javascript, the drawing code is moved to a small set of
    QQuickPaintedItem classes that do the same drawing with QPainter (which
    provides a very similar api).
    
    In order to speed up and avoid detaching, this commit also includes a change to
    how audio levels are passed along - now the pointer to the levels data is made
    available through a QVariant, and the waveform painter will access it using
    in/out points and read directly from that data without deep copying.

It wasn't my intention initially to touch the audio levels part, but I quickly realized how inefficient it was to do a subcopy of the array, send it to javascript, do data processing there, then send it back to c++ for drawing.

This addresses #151

### DIFF
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -126,13 +126,11 @@ QVariant MultitrackModel::data(const QModelIndex &index, int role) const
                 return info->fps;
             case IsAudioRole:
                 return m_trackList[index.internalId()].type == AudioTrackType;
-            case AudioLevelsRole: {
-                QVariantList* levels = (QVariantList*) info->producer->get_data(kAudioLevelsProperty);
-                int channels = 2;
-                if (info->producer && info->producer->is_valid() && info->producer->get_data(kAudioLevelsProperty))
-                    return levels->mid(info->frame_in * channels, info->frame_count * channels);
-                break;
-            }
+            case AudioLevelsRole:
+                if (info->producer->get_data(kAudioLevelsProperty))
+                    return QVariant::fromValue(*((QVariantList*) info->producer->get_data(kAudioLevelsProperty)));
+                else
+                    return QVariant();
             case FadeInRole: {
                 QScopedPointer<Mlt::Filter> filter(getFilter("fadeInVolume", info->producer));
                 if (!filter || !filter->is_valid())
--- a/src/qml/timeline/Clip.qml
+++ b/src/qml/timeline/Clip.qml
@@ -18,6 +18,7 @@
 
 import QtQuick 2.2
 import QtQuick.Controls 1.0
+import Shotcut.Controls 1.0
 import QtGraphicalEffects 1.0
 import QtQml.Models 2.2
 
@@ -88,7 +89,7 @@ Rectangle {
         // Model as a property expression is not working in all cases.
         waveformRepeater.model = Math.ceil(waveform.innerWidth / waveform.maxWidth)
         for (var i = 0; i < waveformRepeater.count; i++)
-            waveformRepeater.itemAt(0).requestPaint()
+            waveformRepeater.itemAt(0).update()
     }
 
     onAudioLevelsChanged: generateWaveform()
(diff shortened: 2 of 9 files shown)


ddennedy:
Yet another solution is to just comment out/remove the call to delete oldTexture and just leak until we can upgrade Qt instead of this heavy change.

ddennedy:
Moving UI elements from QML to C++ is counter to the goal of moving the project toward a C++ runtime and QML UI.

bmatherly:
My observation of the patch is that he is moving some of the heavy lifting related to graphical rendering to C++, but the UI is still built in QML. This isn't unlike other QML components that require heavy lifting (colorwheelitem, etc). QT does the very same thing - simple controls are built with QML while advanced controls are implemented in C++.

We could maybe debate the abstraction/naming. For example, should "painteditems.h/cpp" be renamed to "TimelineItems" or something more specific. Otherwise, the file could become a black hole for all future painteditems.

ddennedy:
Per the e-mail I just sent out, the movement of UI code from QML to C++ is no longer antithetical. I will test this on all platforms. I am a little behind on OS X dev/test environment at the moment, and that might cause some delay.

metellius:
> Yet another solution is to just comment out/remove the call to delete oldTexture and just leak until we can upgrade Qt instead of this heavy change.

The choice of which is the lesser evil is yours - I had mixed feelings moving the short and sweet onPaint code from qml to c++, but at least I felt safer knowing that it's public API.

> Tested on Linux and Windows so far. I had a couple of crashes on Windows when moving clips between tracks that seems to be new.

Maybe this is the one thing that I was unsure about - lifetime and ownership management of the audio levels QVariantList during deletion. If this is the case, two solutions that come to mind are (1) somehow transferring the ownership of the data to QML so that they are cleaned up later, or (2) using shared pointers.

metellius:
> We could maybe debate the abstraction/naming. For example, should "painteditems.h/cpp" be renamed to "TimelineItems" or something more specific. Otherwise, the file could become a black hole for all future painteditems.

I was hoping we wouldn't have future paintedItems at all really, QQuickPaintedItem is often slower than implementing custom QQuickItems with scene graph nodes (i'm curious whether rendering the waveform graph is an exception or not here).
But otherwise, yes, something like TimelineItems would make sense. The names of the actual classes and their qml registered names is also very much debatable.

ddennedy:
> Yet another solution is to just comment out/remove the call to delete oldTexture and just leak 

After I saw how frequently that delete is called, and the object is wraps a GL texture; I did not like that solution.

OK, here is a stack of the crash moving clips between tracks on Windows - deep in Qt. :(
image

metellius:
I would start commenting out stuff to see what is causing it. some things to try:
- have AudioLevels return an empty QVariant. That way the levels data never passes through qml
- return immediately in the paint methods of the painted items
- remove the painteditems from the qml entirely

bmatherly:
> I was hoping we wouldn't have future paintedItems at all really, QQuickPaintedItem
> is often slower than implementing custom QQuickItems with scene graph nodes 
> (i'm curious whether rendering the waveform graph is an exception or not here).
> But otherwise, yes, something like TimelineItems would make sense. The names 
> of the actual classes and their qml registered names is also very much debatable. 

Well, we already have one: ColorWheelItem
I agree that QQuickItems are preferred over QQuickPaintedItems, but you never know what the future holds.

My advice would be to make the file name and class names more specific.

metellius:
> My advice would be to make the file name and class names more specific.

OK, I will follow this advice. Here's a suggestion:
* painteditems is renamed to timelineitems
* classes are renamed to TimelineTransition, TimelinePlayhead, TimelineMuteTriangle and TimelineWaveform.
* The items will be used from QML by the same names.

I'll make this change unless I hear something else from you within a day.

bmatherly:
Sounds good to me.

ddennedy:
Hey, the crash on Windows appears in master as well and is unrelated to the levels data being passed. It is just much easier to reproduce in the debugger. I was comparing the latest release running outside the debugger to this branch running in the debugger and misled. I think this is good after making the requested changes. I am not providing input on the naming discussion; what you guys decided is fine.

Later today, I am getting an OS X machine and in should be ready for dev&test by the weekend.

ddennedy:
I am only waiting on the renaming to pull this.

metellius:
pushed

PR #209, opened 2016-03-14
build-shotcut.sh flags for debug/asan builds

metellius:
DEBUG_BUILD will enable debug builds for qmake and most configure-based
subprojects.
ASAN_BUILD will make it compile with address sanitizer, but in this case only
for mlt, shotcut and webvfx.

### DIFF
--- a/scripts/build-shotcut.sh
+++ b/scripts/build-shotcut.sh
@@ -24,6 +24,8 @@ CLEANUP=1
 ARCHIVE=1
 SOURCES_CLEAN=1
 INSTALL_AS_ROOT=0
+DEBUG_BUILD=0
+ASAN_BUILD=0
 CREATE_STARTUP_SCRIPT=1
 ENABLE_FREI0R=1
 FREI0R_HEAD=1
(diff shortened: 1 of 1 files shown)


scripts/build-shotcut.sh
```diff
@@ -384,6 +386,17 @@ function set_globals {
     fi
   fi
 
+  if [ "$DEBUG_BUILD" = "1" ]; then
+    CONFIGURE_DEBUG_FLAG="--enable-debug"
```
bmatherly: Consider initializing these variables so that we don't end up with uninitialized variables when the flag is not set. Same for ASAN below.

metellius: Done :)

bmatherly:
I think this is a great idea - and is good to go in my opinion. While you are at it with a "debug" flag, you might consider also adding a call to GDB in the shotcut script (which is generated by the build-shotcut.sh script).

I have done this before for debugging MLT. It looks something like this:

gdb `find src/mlt/ -type d -printf '-d %p '` -ex run --args "$INSTALL_DIR/bin/melt" $@

The use of GDB could be predicated by the debug flag.

Thoughts?

metellius:
Why all the -d arguments? For me gdb has always automatically found the source without any extra arguments, but maybe it is because I'm using a local debug-compiled qt toolchain?
Initially I just modified the script and added something like $PRE in there, so that when I needed to debug I just did
`PRE="gdb --args" ./shotcut /path/of/test.mlt`
But recently I'm instead just using source-me in a subshell:
`( cd ~/shotcut/Shotcut/Shotcut.app; source source-me; bin/shotcut /path/of/test.mlt )`
But are you suggesting it always run shotcut in gdb with that flag turned on? It takes ages to start shotcut in gdb for me (lots of debug symbols from qtwebkit etc to load), so I personally would rarely want to always have it run in gdb...

bmatherly:
I'm no GDB expert. For whatever reason, I had to provide the path to source when i used that.

Good point about it taking a long time to start. I only do that when I'm debugging a particularly difficult issue. So maybe having it always use GDB is not a good idea. Just throwing it out there. All I know is that every time I want to use GDB I have to google it to remind myself how to launch it. It would be handy for me if it was just a switch somehow. Probably not worth polluting this patch with my GDB idea.

metellius:
Yeah, I'll leave it for another patch. Thanks for reviewing.

PR #274, opened 2016-07-09
stabilize filter: Add a SliderSpinner to enable adjusting the 'smooth…

chmorgan:
…ing' property of the stabilize filter

The smoothing property controls how many frames are used for the low pass filtering of motion. A
larger value means that more camera movement is transformed away.

The 'smoothing' property is an integer. Add QmlFilter::getInt() to match the type of the property.

### DIFF
--- a/src/qml/filters/stabilize/ui.qml
+++ b/src/qml/filters/stabilize/ui.qml
@@ -166,6 +166,25 @@ Item {
             onClicked: zoomSlider.value = 0
         }
 
+        Label {
+            text: qsTr('Smoothing')
+            Layout.alignment: Qt.AlignRight
+        }
+        SliderSpinner {
+            id: smoothingSlider
+            minimumValue: 0
+            maximumValue: 100
+            tickmarksEnabled: true
+            stepSize: 1
+            value: filter.get('smoothing')
+            onValueChanged: {
+                filter.set('smoothing', value)
+            }
+        }
+        UndoButton {
+            onClicked: smoothingSlider.value = 15
+        }
+
         Button {
             id: button
             text: qsTr('Analyze')


bmatherly:
Thanks for your contribution. This looks good to me. I think that "smoothing" is a good option to add.

When considering your changes, it occurred to me that it might not be obvious to some users what the options do. Since you are saturated on this, would you be willing to add tool tips to all the Stabilize filter options to help people know how to use them? It could be a new pull request and would not have to block this one.

ddennedy:
Remove the getInt(); it is not needed. How do you think we got along this far without it?

chmorgan:
@bmatherly would you know of a tooltip example I could copy from? Would you know if those have access to the descriptions provided in the yaml on the properties that are defined in the mlt filter?

chmorgan:
@ddennedy thats a really good question. It looked missing and double didn't match up with the mlt property. The variable is defined as an int by mlt. Should it be retrieved via getDouble() even in this case? Would that result in any aliasing for a value that has to be a whole number?

ddennedy:
Just use get() and set() for an int. Both MLT properties and QML variables are variant types. In the beginning, I tried to get by without getDouble() but ran into some numeric locale issues with decimal point handling.

bmatherly:
Here is an example using ToolTip:
https://github.com/mltframework/shotcut/blob/master/src/qml/filters/audio_normalize_1p/ui.qml#L25
https://github.com/mltframework/shotcut/blob/master/src/qml/filters/audio_normalize_1p/ui.qml#L99

We had to make our own because QML does not offer a ToolTip control:
https://github.com/mltframework/shotcut/blob/master/src/qml/modules/Shotcut/Controls/ToolTip.qml

ddennedy:
Some Qt Quick Controls offer a tooltip property that should be used where available. Then, use ours when not available.

bmatherly:
Button is the only control that offers tooltip. When using a button, use the built-in tooltip.

chmorgan:
@ddennedy, @bmatherly repushed with just the changes to add the smoothing property. If I can't get this in without tooltips I'll add them but was hoping to start small in terms of contributions as the time spent refactoring scales with the size of the changes. I can make the tooltips changes next, was going to look at pulling from the properties documentation in mlt for those so I didn't have to copy the text over, but that is only in english so maybe that isn't a good idea.

bmatherly:
+1 from me.

I recommend putting tooltips in a different pull request. Regarding pulling from properties... We have avoided that in the past because it makes translation difficult. Also, in general, some of the property documentation is written from the perspective of melt or a programmer and not necessarily from the perspective of Shotcut. My advice would be to copy/paste text from the properties if it is appropriate and then modify/embellish it to make it as user friendly as possible within Shotcut.

ddennedy:
I just realize that this new parameter has no unit of measure and goes from 0 - 100 with no indication of whether 0 or 100 is smoothest. Looking at the docs:
https://www.mltframework.org/bin/view/MLT/FilterVidstab#smoothing
Is this parameter N in the (2N + 1)?
Is too inaccurate or misleading to simply give this parameter a unit of measure of "frames"?

chmorgan:
I'm thinking that frames would be a reasonable units. Were you saying that it would make sense to set the 'suffix' of that property to be 'frames'?

An aside, the width of the value field of the slider spinner doesn't appear to resize correctly. For instance in the stabilize filter the zoom field shows x.y% even if the percentage is greater than 10.

bmatherly:
I agree: "Frames" is a good unit of measure.

When you go to make tooltips, I would leave out the (2N +1) equation as it might confuse people. I assume that the equation means: current frame + N frames before + N frames after. I'm not sure users need to know or care about that distinction.

ddennedy:
>  'suffix' of that property to be 'frames'?

`qsTr(' frames')` qsTr() to make it translatable. We do not always make suffix translatable if it has a commonly used symbol or abbreviation, but perhaps we should and will because even those can vary. Also, the controls do not put a space between the value and the suffix, and I like to have a space there; so, there is an intentional space at the beginning of this string.

> the width of the value field of the slider spinner doesn't appear to resize correctly

It is not resize automatic. It must be changed from its default with the spinnerWidth property. With that said, it looks fine for me on Windows 10 using 125% text size. It is not independent of factors such as OS, and system font settings. However, it does automatically adjust for whole jumps in DPI. For example, Apple Retina systems show a devicePixelRatio within Qt as 2.0, but on Windows it will not show  1.5 for 150%. It only shows 2.0 when you go to 200%. So, largely, we have to just increase the width of this property as-needed.

PR #275, opened 2016-07-09
stabilize filter ui: Group the 'Analyze' button and analyze parameter…

chmorgan:
…s togehter using a GroupBox

Analyze settings and button work together and are separate from the filter options.

GroupBox's title does not support text attributes so remove the <b></b> tags around the translation string
template and all of the translated strings in the .ts files.

### DIFF
--- a/src/qml/filters/stabilize/ui.qml
+++ b/src/qml/filters/stabilize/ui.qml
@@ -102,43 +102,69 @@ Item {
         anchors.fill: parent
         anchors.margins: 8
 
-        Label {
-            text: qsTr('<b>Analyze Options</b>')
+        GroupBox {
+            title: qsTr('Analyze Options')
+            anchors.left: parent.left
+            anchors.right: parent.right
             Layout.columnSpan: 3
-        }
 
-        Label {
-            text: qsTr('Shakiness')
-            Layout.alignment: Qt.AlignRight
-        }
-        SliderSpinner {
-            id: shakinessSlider
-            minimumValue: 1
-            maximumValue: 10
-            tickmarksEnabled: true
-            stepSize: 1
-            value: filter.getDouble('shakiness')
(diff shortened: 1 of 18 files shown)


ddennedy:
GroupBoxes are ugly, and I refuse to use them. Why the do you think I made bold labels for a heading and indentation?!

Never include changes to translation files in your pull requests. That process is handled separately so we do not nag the transifex.com translators on a daily basis whenever a string change occurs.

chmorgan:
@ddennedy hmm. Seemed like they looked ok to me but I'm not fixated on using them. Its my first use of qml.

Ok with moving the button up above the label as to separate the analysis section from the filter properties section? If one were to change/alter the source translation string what happens in the case where you'd have orphaned strings in the .ts files?

Also, do you normally edit the qml in qtcreator? I tried going to the 'design' view after opening the project and it had errors, reported perfectly good properties like 'title' as being invalid, and told me to fix the qml. Maybe qtcreator can't edit qml unless its in a particular format? Or my version from qt 5.7 is too new?

ddennedy:
For filter UIs, I tend to prefer action-style buttons at the bottom like the convention for a dialog window - unless, perhaps, if there are many items in the window and it becomes cluttered and confusing. If you disagree for this filter, you can move the button, and I will try it out.

The english .ts file is picked up by transifex.com from git whenever it changes. Then, translators get a notification and make the changes on transifex.com. I update this (english .ts) in a git a week before release. Then, the night before release, I download all translations (except english) from transifex.com, update them with lupdate one last time, and upload them to git.

QML Designer is worthless to Shotcut. I think it only works with pure and simple Qt QML (no extensions). It can be handy to mockup animations, but we make very little use of them. You should just edit the QML source and preview it in Shotcut. You can setup Qt Creator to run Shotcut without building it. That way you can make a change, and run it from within there and see messages in its panel.

PR #276, opened 2016-07-10
stabilize filter ui: Move the 'Analyze' button up under the 'Analyze …

chmorgan:
…Options' to make it clearer that this button is only related to the analyze properties

Imo this makes it clearer to the user that the analyze goes with the analyze settings and that the other settings don't require you to press the analyze button again.

It would be even neater if we could auto-analyze when the user enabled the filter, but that would require some hysteresis and would have to handle the case where the user didn't wait for the analysis to complete etc.

Another idea is tracking what settings were used to indicate to the user when analyze had to be run again. That might be a simpler change that better fits with how the current filter uis are designed.

### DIFF
--- a/src/qml/filters/stabilize/ui.qml
+++ b/src/qml/filters/stabilize/ui.qml
@@ -28,18 +28,28 @@ Item {
     height: 150
     property string settingsSavePath: settings.savePath
 
+    function hasFilterCompleted() {
+        return (filter.get("results").length > 0 &&
+                filter.get("filename").indexOf(filter.get("results")) !== -1)
+    }
+
     function setStatus( inProgress ) {
         if (inProgress) {
             status.text = qsTr('Analyzing...')
         }
-        else if (filter.get("results").length > 0 && 
-                 filter.get("filename").indexOf(filter.get("results")) !== -1) {
+        else if (hasFilterCompleted()) {
             status.text = qsTr('Analysis complete.')
         }
         else
         {
             status.text = qsTr('Click Analyze to use this filter.')
         }
+    }
+
+    function analyzeValueChanged() {
+        button.enabled = true
+        status.text = qsTr('Analysis required')
+        filter.set("results", '')
     }
 
     // This signal is used to workaround context properties not available in
@@ -57,7 +67,6 @@ Item {
         onAnalyzeFinished: {
             filter.set("reload", 1);
             setStatus(false)
-            button.enabled = true
         }
     }
     
@@ -118,7 +127,10 @@ Item {
             tickmarksEnabled: true
             stepSize: 1
             value: filter.getDouble('shakiness')
-            onValueChanged: filter.set('shakiness', value)
+            onValueChanged: {
+                filter.set('shakiness', value)
+                analyzeValueChanged()
+            }
         }
         UndoButton {
             onClicked: shakinessSlider.value = 4
@@ -135,7 +147,10 @@ Item {
             tickmarksEnabled: true
             stepSize: 1
             value: filter.getDouble('accuracy')
-            onValueChanged: filter.set('accuracy', value)
+            onValueChanged: {
+                filter.set('accuracy', value)
+                analyzeValueChanged()
+            }
         }
         UndoButton {
             onClicked: accuracySlider.value = 4
@@ -145,6 +160,7 @@ Item {
             id: button
             text: qsTr('Analyze')
             Layout.alignment: Qt.AlignRight
+            enabled: !hasFilterCompleted()
             onClicked: {
                 button.enabled = false
                 fileDialog.folder = settings.savePath
--- a/src/qml/filters/stabilize/ui.qml
+++ b/src/qml/filters/stabilize/ui.qml
@@ -141,6 +141,24 @@ Item {
             onClicked: accuracySlider.value = 4
         }
 
+        Button {
+            id: button
+            text: qsTr('Analyze')
+            Layout.alignment: Qt.AlignRight
+            onClicked: {
+                button.enabled = false
+                fileDialog.folder = settings.savePath
+                fileDialog.open()
+            }
+        }
+        Label {
+            id: status
+            Layout.columnSpan: 2
+            Component.onCompleted: {
+                setStatus(false)
+            }
+        }
+
         Label {
             text: qsTr('<b>Filter Options</b>')
             Layout.columnSpan: 3
@@ -183,24 +201,6 @@ Item {
         }
         UndoButton {
             onClicked: smoothingSlider.value = 15
-        }
-
-        Button {
-            id: button
-            text: qsTr('Analyze')
-            Layout.alignment: Qt.AlignRight
-            onClicked: {
-                button.enabled = false
-                fileDialog.folder = settings.savePath
-                fileDialog.open()
-            }
-        }
-        Label {
-            id: status
-            Layout.columnSpan: 2
-            Component.onCompleted: {
-                setStatus(false)
-            }
         }
 
         Item {


bmatherly:
Thanks for your work on this. I believe that polishing these kinds of UI issues can make a better experience for users in the long run.

I am kind of torn on this one. On the one hand, I can appreciate your desire to group the analysis things together. On the other hand, I'm not sure if this change reduces confusion.

Also, we should give some consideration to keeping symmetry with the "Normalize: two pass" filter. The difference is that the Normalize filter does not have any "Analyze" options.

Maybe the Stabilize filter should be designed to have two "steps". The first step would be to configure and run the analysis - with the filter options either disabled or hidden. Then, after analysis is complete, the filter options are available - with an option to "Re-Analyze" (or go back to step 1, essentially).

ddennedy:
I do not really like the appearance of the location of the new button. Maybe because it appears in the label column. Here is a variation that places it in the control column, which I like better:
```javascript
--- src/qml/filters/stabilize/ui.qml
+++ src/qml/filters/stabilize/ui.qml
@@ -141,6 +141,28 @@ Item {
             onClicked: accuracySlider.value = 4
         }
 
+        Item {
+            Layout.fillWidth: true
+        }
+        RowLayout {
+            Layout.columnSpan: 2
+            Button {
+                id: button
+                text: qsTr('Analyze')
+                onClicked: {
+                    button.enabled = false
+                    fileDialog.folder = settings.savePath
+                    fileDialog.open()
+                }
+            }
+            Label {
+                id: status
+                Component.onCompleted: {
+                    setStatus(false)
+                }
+            }
+        }
+
         Label {
             text: qsTr('<b>Filter Options</b>')
             Layout.columnSpan: 3
@@ -185,24 +207,6 @@ Item {
             onClicked: smoothingSlider.value = 15
         }
 
-        Button {
-            id: button
-            text: qsTr('Analyze')
-            Layout.alignment: Qt.AlignRight
-            onClicked: {
-                button.enabled = false
-                fileDialog.folder = settings.savePath
-                fileDialog.open()
-            }
-        }
-        Label {
-            id: status
-            Layout.columnSpan: 2
-            Component.onCompleted: {
-                setStatus(false)
-            }
-        }
-
         Item {
             Layout.fillHeight: true;
         }
```

ddennedy:
Regarding Brian's last comment, I agree it can be nice to disable the Filter Options until after Analyze is complete. However, always keep the Analyze Options available. Upon existence of valid analysis file, disable the Analyze button and hide the help text beside it. Then, whenever someone changes one of the Analyze Options, enable the Analyze button and show a new help text beside it: You must click Analyze to use the new option values. Except in this case, since we already have a .stab file, we do not need to prompt for it again. There is no need to keep track of the settings; simply set and clear a boolean pending results state.

I am not very keen to make it automatically start analysis, find and kill a running job whenever an Analyze Option changes, and start a new job. I think that is unnecessarily complicated and takes some control away from the user. Keep it simple, but add more feedback and context sensitivity.

ddennedy:
Chris, send an email to dan@dennedy.org if you want a free membership to the patron forum for a couple of months for your contribution. With a more substantial contribution, I will give you a year. If you become a somewhat regular contributor, you will become a part of the team and therefore get a lifetime membership.

chmorgan:
@ddennedy thoughts about multiple cuts from the same source where each is using the stabilize filter? That would seem to necessitate a .stab file for each and a button to click for each.

I like those improvements. Let me see if I can figure out how to cleanly do that in qml.

ddennedy:
You can make a project consisting of only a full clip and its .stab file and then open that MLT XML as a virtual clip to use in other projects making cuts from it. Yes, it would be nice to select a .stab file in the stabilize filter, but I know people are going to shoot their self in the foot with that. ;-)

chmorgan:
Yeah. i was wondering how to manage reference counts in the case where we might have the same file used by multiple filters. What if they changed an analyze property etc?

Had no idea you could open a project as a virtual clip like that. That sounds like something I need for a video I'm working on right now.



Thoughts on selecting a filename for the user? filter hash + filename + .stab in the /tmp folder (or whatever cross platform thing qt provides for that kind of thing)?

chmorgan:
@ddennedy, @bmatherly  added some logic to control the state of buttons/controls based on the analysis status. Didn't do any hiding/showing of the filter options yet, thought that could be another commit or pull request.

src/qml/filters/stabilize/ui.qml
```diff
@@ -29,19 +29,36 @@ Item {
     property string settingsSavePath: settings.savePath
 
     function setStatus( inProgress ) {
```
bmatherly: Nit: extra space after (

src/qml/filters/stabilize/ui.qml
```diff
@@ -29,19 +29,36 @@ Item {
...
         }
-        else
+        else // analysis failed
         {
             status.text = qsTr('Click Analyze to use this filter.')
+            button.enabled = true
         }
     }
 
+    function analyzeValueChanged( ) {
```
bmatherly: Nit: extra space between ()

ddennedy:
No to saving the stab file in tmp or an app data folder because that breaks portable and archivable projects. It is possible to make projects with relative paths.

chmorgan:
@ddennedy hmm. I was trying to think of a way to make it more automatic for the user if you can think of one. Was surprised I had to pick a filename the first time it came up, then worry about not deleting it from the filesystem etc.

src/qml/filters/stabilize/ui.qml
chmorgan: @bmatherly  I fixed the if() whitespace by altering my commit and remove the whitespace in setStatus() with a separate commit.

bmatherly:
I think it is important to force the user to pick a filename so that they are aware that it exists and they manage it just like they should manage their project file and media. I would like it if the user were able to view the file name/path somehow - maybe a tooltip. Also, if I choose analyze a second time, I wish the file chooser dialog would default to the current file that was chosen last time.

chmorgan:
@bmatherly that's not too bad to implement and I think makes a lot of sense. If these changes look good I can do that a near future PR.

ddennedy:
On the road map is "project management." For this, you would set a project folder, and it will save things there automatically and have options to copy or move media into it. It might include other things as well.

ddennedy:
[Warning] <> file:///C:/Projects/Shotcut/share/shotcut/qml/filters/stabilize/ui.qml:229:13: id is not unique

ddennedy:
After successfully completing Analyze and after the filter loses focus (click on another filter or clip) and you return to stabilize, it says "Analysis complete" but the Analyze button is enabled. It would be nice to disable it until an option is changed just like after analysis is done. Not required, but since you have the above fix to make, you might consider it. Thanks!

chmorgan:
@ddennedy doh. Same happens if you switch filters. It looks like the .qml is re-instantiated each time the filter is shown/hidden in the menu. If we can't save state there I looked at QmlFilter. I can't see anything in the QmlFilter class that would let me determine if there was an active analyze running. The job is created, queued and the delegate created but otherwise no state appears to be saved in QmlFilter. Suggestions for direction are welcome.

ddennedy:
No there is no way to save state other than to set a custom filter property, but that will not help you here. I was referring to after analysis is complete, not in the middle of running.

cmo-hms:
@ddennedy it sounds like a similar issue. The qml or filter isn't keeping state of whether the analysis had run or completed so it has no way of knowing what its state is when it is reloaded.

Initial implementation thought was to have QmlFilter attach to its own analyzeFinished slot and use that to track whether analysis had completed or not. Trouble is that then we wouldn't be tracking the dirty status when the filter parameters changed.

Other thought, subclass QmlFilter -> StabilizeQmlFilter and overload 'set' for the appropriate parameter changes and set an "analysis needed" boolean upon change.

Other thoughts?

chmorgan:
And I keep posting from th@ddennedy it sounds like a similar issue. The qml or filter isn't keeping state of whether the analysis had run or completed so it has no way of knowing what its state is when it is reloaded.

Initial implementation thought was to have QmlFilter attach to its own analyzeFinished slot and use that to track whether analysis had completed or not. Trouble is that then we wouldn't be tracking the dirty status when the filter parameters changed.

Other thought, subclass QmlFilter -> StabilizeQmlFilter and overload 'set' for the appropriate parameter changes and set an "analysis needed" boolean upon change.

Other thoughts?

ddennedy:
After running analyze once, letting it complete, leaving the filter and returning, it says "Analysis complete." So, there is **some** state. How does it know that? (I know; no need to answer. I am just trying to help you understand.) All I am saying is that the Analyze button at this point is enabled when it should be disabled. Do not try anything more complicated than fixing that at this point. Also, you still have not fixed the loading error I mentioned above, which is an easy fix.

chmorgan:
@ddennedy I get where you are going. Also, fixed the other issue but it wasn't in the last push. Will push tomorrow with something that I think can accomplish much of this. Thanks for the pointer and patience.

chmorgan:
I think this is a little better than the last pull request but I'm guessing it still has some holes in its approach. Feedback welcome of course.

src/qml/filters/stabilize/ui.qml
```diff
@@ -42,6 +46,12 @@ Item {
         }
     }
 
+    function analyzeValueChanged() {
+        button.enabled = true
+        status.text = qsTr('Analysis required')
+        filter.set("results", '')
```
ddennedy: Now, after analysis is complete and you leave the filter and return to it, it says Analysis required. That is because the option values are loaded when the UI loads, which triggers this function and clears the results. Clearing results really does make analysis required. If you remove this line there is stil a race condition that typically causes this function to be called and give the superficial impression that it is. The solution is to make a top-level Component.onCompleted script block that controls the order of setting option values, the button state, and label. I did that and also refactored some things as well.

ddennedy:
See this commit 5a5bce6d79a96cc2621444466cd286c8b6a606dd to see how I fixed it.

In the future, the first line of git commit messages should be short followed by a blank line and more information.

chmorgan:
@ddennedy ahh I see what was going on and how the change you made corrected it.

PR #281, opened 2016-07-23
Paint playlist items using a custom delegate

metellius:
This uses the screen space more efficiently because we are no longer
constrained to a tabular layout with one cell for each piece of information.

Short summary of the changes so far:

- A class inheriting from QStyledItemDelegate is introduced, which handles painting of items
- PlaylistModel is modified so the different pieces of information are available as different roles, and not columns. This lets the delegate painter retrieve all it wants even though it's just painting a single cell.

This pull request is not ready for merge, because there is still one issue remaining that will affect how to proceed: compact playlists without thumbnails.
For the thumbnail case, there is enough space next to the thumbnail to present clip in/duration/start, like this:

But, when choosing the compact style AKA "hidden", the row height is too small to fit all of this. For reference, here's how it looks in the latest release:

Here's two suggested solutions:

(1) We set the row height to font lineheight * 2, and wrap the time fields into two lines. Pros: still no horizontal scrollbars. Cons: Less compact than before

(2) When choosing "hidden", the model goes back to COLUMN_COUNT like today, and shows the in/duration/start fields in columns. For thumbnail modes, there is only one column and painting is done with the delegate painter. Pros: Rows are just as compact as today. Cons: more code complexity, horizontal scroll bars

I'm leaning towards (1), but I'd like your opinion before I do any more because going for (2) would need more refactoring.

### DIFF
--- a/src/docks/playlistdock.cpp
+++ b/src/docks/playlistdock.cpp
@@ -22,9 +22,106 @@
 #include "mainwindow.h"
 #include "settings.h"
 #include "shotcut_mlt_properties.h"
+#include "widgets/playlisticonview.h"
 #include <commands/playlistcommands.h>
-#include <QMenu>
 #include <Logger.h>
+
+#include <QMenu>
+#include <QStyledItemDelegate>
+#include <QPainter>
+#include <QDebug>
+#include <QHeaderView>
+
+class TiledItemDelegate : public QStyledItemDelegate
+{
+    Q_OBJECT
+public:
+    TiledItemDelegate(QAbstractItemView * view, QWidget *parent = 0)
+         : QStyledItemDelegate(parent),
+         m_view(view)
+    {
+        connect(&Settings, SIGNAL(playlistThumbnailsChanged()),
+                SLOT(emitSizeHintChanged()));
+    }
+
+    void paint(QPainter *painter,
+               const QStyleOptionViewItem &option, const QModelIndex &index) const
+    {
+        const QImage thumb = index.data(Qt::DecorationRole).value<QImage>();
+        const QString setting = Settings.playlistThumbnails();
+        const int lineHeight = painter->fontMetrics().height();
+        const bool roomEnoughForAllDetails = lineHeight * 4 < thumb.height();
+        const QFont oldFont = painter->font();
+        QFont boldFont(oldFont);
+        boldFont.setBold(true);
+
+        if (option.state & QStyle::State_Selected) {
+            painter->fillRect(option.rect, option.palette.highlight().color());
+        } else {
+            if (option.features & QStyleOptionViewItem::Alternate)
+                painter->fillRect(option.rect, option.palette.alternateBase());
+        }
+
+        QRect thumbRect(QPoint(), thumb.size());
+        thumbRect.moveCenter(option.rect.center());
+        thumbRect.moveLeft(0);
+        painter->drawImage(thumbRect, thumb);
+
+        const QPoint indexPos = option.rect.topLeft() + QPoint(5, 20);
+        const QString indexStr = "#" + index.data(PlaylistModel::FIELD_INDEX).toString();
+        painter->setFont(boldFont);
+        painter->setPen(option.palette.color(QPalette::Dark).darker());
+        painter->drawText(indexPos, indexStr);
+        painter->setPen(option.palette.color(QPalette::WindowText));
+        painter->drawText(indexPos - QPoint(1, 1), indexStr);
+        painter->setFont(oldFont);
+
+        QRect centeredTextRect = option.rect;
+        centeredTextRect.setHeight(lineHeight * (roomEnoughForAllDetails ? 4 : 2));
+        centeredTextRect.moveCenter(option.rect.center());
+
+        QRect textRect = centeredTextRect;
+        textRect.setLeft(thumb.width() + 10);
+
+        QPoint textPoint = textRect.topLeft();
+        textPoint.setY(textPoint.y() + lineHeight);
+        painter->setFont(boldFont);
+        painter->drawText(textPoint,
+                painter->fontMetrics().elidedText(index.data(Qt::DisplayRole).toString(), Qt::ElideMiddle, textRect.width()));
+        painter->setFont(oldFont);
+
+        textPoint.setY(textPoint.y() + lineHeight);
+        painter->drawText(textPoint, tr("Duration: %1").arg(index.data(PlaylistModel::FIELD_DURATION).toString()));
+        if (roomEnoughForAllDetails) {
+            textPoint.setY(textPoint.y() + lineHeight);
+            painter->drawText(textPoint, tr("In: %1").arg(index.data(PlaylistModel::FIELD_IN).toString()));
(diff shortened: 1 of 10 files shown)


ddennedy:
I require any change here to still require a grid view. So, 2. Why is it more complex? We should be able to switch between using a delegate and not, and let the model support both. With that said, just like today there is an option for no thumbnail, maybe what I will call the new "tile view" can have an option for no thumbnail. In the future, there could be additional columns and column options for the details view.

I also want any change here to include an icon view where there is only a name displayed under the icon. If you are not up for that, I will work on it, but that will delay it some, naturally.

metellius:
Most of the complexity would be with the data function sometimes following columns and sometimes only roles, and emitting model change signals correctly when the table changes. But it is doable.

Either way I didn't think about a tile view, and in this case there is no way around having multiple columns (by this I initially thought you meant an file manager sort of icon view, but as far as I know there's no standard component for this in Qt, so we're better off using the tableview). It sounds like a useful view mode, so I'll take the task.

ddennedy:
I understand what you mean about the model. Even if the data() method can respond to both column and role, I suppose it still must have a mode to generate QModelIndex correctly for the change signals.

As for tile view, I meant that I call your new view the tile view - a vertical list of tiles. It too might be configurable to not show a thumbnail. Just imagine your first screenshot above without the thumbnail.

As for an icon view, see QListView::IconMode.

Three modes: detail (today's QTableView), tile, and icon. (Except, in icon view mode, I think it makes little sense to not show a thumbnail.)

metellius:
Aha, didn't know about IconMode. Great, then we have all we need.

metellius:
The icon mode is showing to be quite the hassle when it comes to playlist reordering. The problem starts with QListView **insisting** on handling drag'n'drop by moving the icons internally, and the only (as far as I can see) API you get to respond to this interaction is the _indexesMoved_ signal, which just gives you a list of QModelIndexes and nothing more. Trying to figure out what row in the model one of these modelindexes belong to seems impossible.

I was able to work around this somewhat by subclassing QListView and going straight to QAbstractItemVIew::dropEvent for the icons mode, but this is a hack that only got me a little further: dropping between icons just appends them at the end (the normal listview doesn't have any concept of spacing). The drop indicator only shows up on the top and bottom of icons depending on y-position, and x-position is ignored.

All in all icons mode in qlistview is lacking when it comes to DND, and I suggest we just do setMovement(QListView::Static) and maybe instead add a sort action and/or have the option of dropped items just be inserted alphabetically. Horizontally flowing icons is a somewhat awkward representation of an ordered playlist after all.

ddennedy:
I see. I think icon view is more useful when using Playlist as a bin, which is probably a more popular use case than using it as a playlist. In that case, DND is perhaps not as important, and certainly sorting by name is useful. OTOH some people might like the ability to control ordering - e.g. while reviewing, moving more interesting clips to the beginning. I think making it static creates inconsistent usability. From user perspective, simply changing the presentation should not affect behavior so much. In that case, please leave out icon view for now. I am sorry if you put a lot of time into it simply because I asked for it.

metellius:
Rather than the amount of work I put into it, I'm more focused on the usefulness of the feature. After trying it out I feel it's compact and nice when I have lots of clips in the bin to choose from, even if it's missing reordering.

Here's what it looks like with large thumbs:


Mode menu



It's not good with a situation where we have to tell the users to switch view mode whenever they want to reorder clips, but it could be marked as experimental and then be improved later by subclassing QAbstractItemView and doing the icons mode ourselves.

I'm going to push the incomplete branch to this PR as it stands with a disclaimer that there are unrelated changes, debug prints and glitchy signal emissions that should overlooked.

ddennedy:
Great. I like the idea of marking it "experimental." I will check it out.

metellius:
Did some experiments with a custom view like I mentioned, and was pleasantly surprised to see how much QAbstractItemView takes care of things for you while you mostly only have to focus on painting and layout. Here's a screenshot of what it looks like so far:



Drag'n'drop from the view to the mainwindow just works without much effort, and for DND into the view it just appends it to the end of the playlist. I suppose there's some more virtual functions that needs implementation for it to insert into the playlist. I'm also quite pleased with how much control of the look and interaction I get by rolling my own view, as opposed to reverse-engineering and fighting with the stock QListView.

ddennedy:
I did a quick initial review. The tiles mode has some rendering issues on Windows where I am testing it. There is always a horizontal scroll bar. If I resize the panel via splitter, then scroll bar increases (shows more to scroll) after I return to roughly same place/size. Then, when I scroll there are text rendering artifacts. Also, in this view mode, drag-n-drop to reorder clips does not work. Did you know about that?

metellius:
Thanks for the review. There's still glitches and artifacts all around because I haven't done a lot of bugfixing yet. The horizontal scrollbar is because I haven't done the itemdelegate sizehint function properly yet. Going back to the commit I pushed, I can confirm that the DND reordering was indeed broken, but it has been fixed in my current HEAD. I'm not going to spend time figuring out exactly why since it's in a such early state.

Once I smooth out a few more issues I will now push my current branch HEAD which includes the custom icon view.

metellius:
Pushed a new revision now. The icon view is shaping up nicely, and I actually see myself using that more than the tiled view. The biggest missing feature that I'm aware of now is missing save/load of view mode.

I threw in a sort function today as well. It's available from the menu button, but only when in multitrack mode.

ddennedy:
> when in multitrack mode

You can not assume the playlist is used as a media bin simply because the project is multitrack. See Export > From > Playlist. You can still use the playlist as a true playlist. Or, another use case is to use it to stage a track and choose "Add All to Timeline." So, the question is: does the user expect the sorted order to affect the sequence of those operations? Probably so, but it will not with a simple model sort. Therefore, sorting must reorder the MLT playlist.

metellius:
> So, the question is: does the user expect the sorted order to affect the sequence of those operations? Probably so, but it will not with a simple model sort. Therefore, sorting must reorder the MLT playlist.

What do you mean by a simple model sort? The sort function in the commit I pushed today resides in PlaylistModel and looks at the resource column in the playlist model, sorts it and calls Mlt::Playlist::move to move items around. Is that not the right way?

ddennedy:
Nevermind. I did not look at the code and thought you were using QSortFilterProxyModel instead of Mlt::Playlist::move. Sounds good the way it is.

metellius:
Think I smoothed out most issues now, and moved the sort functionality to another branch for a later pull request (it needs a bit of rework to work with undo). Let me know if you see those rendering artifacts again (I haven't seen any, a screenshot would be helpful).

ddennedy:
This looks really good now! I did not experience any artifacts as before. I think the only complaint I have is that in Icons view mode, the thumbnails are too small. This is especially noticeable on the "In and Out - Left/Right" mode, but in general the size displayed is smaller than other view modes. What do you think?

ddennedy:
I will pull this, and then you can directly commit tweaks and fixes. I gave you write access to shotcut repo so you can do that. Of course, it is nice to have major changes still reviewed separately.

metellius:
Ok, will do.

PR #310, opened 2016-09-13
Merge adjacent clips feature

metellius:
Available through the clip context menu for now, unsure if there should be a keyboard shortcut for this or not.

### DIFF
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -2648,6 +2648,43 @@ void MultitrackModel::insertOrAdjustBlankAt(QList<int> tracks, int position, int
     }
 }
 
+bool MultitrackModel::mergeClipWithNext(int trackIndex, int clipIndex, bool dryrun)
+{
+    int i = m_trackList.at(trackIndex).mlt_index;
+    QScopedPointer<Mlt::Producer> track(m_tractor->track(i));
+    if (!track)
+        return false;
+    Mlt::Playlist playlist(*track);
+    if (clipIndex >= playlist.count() + 1)
+        return false;
+    Mlt::ClipInfo clip1;
+    Mlt::ClipInfo clip2;
+    playlist.clip_info(clipIndex, &clip1);
+    playlist.clip_info(clipIndex + 1, &clip2);
+
+    if (QString::fromUtf8(clip1.resource) != QString::fromUtf8(clip2.resource))
+        return false;
+
+    if (clip1.frame_out + 1 != clip2.frame_in)
+        return false;
+
+    if (dryrun)
+        return true;
+
+    playlist.resize_clip(clipIndex, clip1.frame_in, clip1.frame_out + clip2.frame_count);
+    QModelIndex modelIndex = createIndex(clipIndex, trackIndex);
+    QVector<int> roles;
+    roles << DurationRole;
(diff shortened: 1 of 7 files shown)


ddennedy:
Does it let you merge adjacent clips from different sources?

metellius:
It only allows it if the resource property is exactly the same if thats what you mean. Btw, i wondered if the join clips function that I found in mlt_playlist was something I should use for this, but when I tried it just made this weird <playlist> clip instead...

ddennedy:
It is good to use MLT APIs where you can, but sometimes they are macro-like and make undo difficult to implement. It all depends. I will review your change soon. As far as keyboard shortcut, we can wait. If you want to be proactive about that, some times I see what some other tools such Premiere or Avid use for that, if available, and see if there is no conflict and makes sense in the context of our other shortcuts.

ddennedy:
Oh, one more thing to discuss. What happens with the filters when merging?

src/models/multitrackmodel.cpp
```diff
@@ -2648,6 +2648,43 @@ void MultitrackModel::insertOrAdjustBlankAt(QList<int> tracks, int position, int
...
+        return false;
+    Mlt::Playlist playlist(*track);
+    if (clipIndex >= playlist.count() + 1)
+        return false;
+    Mlt::ClipInfo clip1;
+    Mlt::ClipInfo clip2;
+    playlist.clip_info(clipIndex, &clip1);
+    playlist.clip_info(clipIndex + 1, &clip2);
+
+    if (QString::fromUtf8(clip1.resource) != QString::fromUtf8(clip2.resource))
```
ddennedy: I prefer to use qstrcmp() when comparing C strings.

metellius:
> What happens with the filters when merging?

It does not take filters into account yet. The easiest way out is just to refuse merging if they have filters applied (maybe show an error message), otherwise we'd maybe try comparing filter xml or something like that and I'm unsure how that would work out?

ddennedy:
We now have copy/paste filters, but there is something broken at the moment on it for timeline clips. I will get that working and then I will apply it on top of the change to preserve clip A's filters only. Then, I will see about making out filters on clip B special case handling.

tin2tin:
Way, way back I did an icon for this function. If you want to add a button, maybe it could be used as a placeholder(until someone with talents for making icons comes along). It's called "splice": [tin2tin_icons8 (1).zip](https://github.com/mltframework/shotcut/files/473500/tin2tin_icons8.1.zip)

ddennedy:
I just tested it and already filters on A clip are preserved.

AlexisWilke:
It did not work correctly for me. I am splitting to be able to mute some part of the audio and when I merge back, the audio is back... I suppose I have to apply the audio changes before the merge. But I wanted to mention that. Testing further.

ddennedy:
@AlexisWilke This is a merge, not a grouping function. Merging cannot keep different filters on the different sections of footage.

AlexisWilke:
Just an idea (and probably requires another issue?)... maybe you could warn the user if the list of filters is different in the clips being merged.

Hritik14:
Has this feature been removed ? Couldn't find it in the "Clip" context menu.

bmatherly:
It is under the "More" submenu:
image

PR #320, opened 2016-10-08
Add setting and command line option to set path for settings, presets…

2wayne:
…, and database files for so-called portableapp.

It's work in progress. I'd like to get some feedback.

### DIFF
--- a/src/autosavefile.cpp
+++ b/src/autosavefile.cpp
@@ -27,6 +27,8 @@
 static const QLatin1String subdir("/autosave");
 static const QLatin1String extension(".mlt");
 
+static QString g_AutoSavePath = "";
+
 static QString hashName(const QString &name)
 {
     return QString::fromLatin1(QCryptographicHash::hash(name.toUtf8(), QCryptographicHash::Md5).toHex());
@@ -75,8 +77,13 @@ bool AutoSaveFile::open(OpenMode openmode)
 AutoSaveFile* AutoSaveFile::getFile(const QString &filename)
 {
     AutoSaveFile* result = 0;
+    QFileInfo info;
     QDir appDir(path());
-    QFileInfo info(appDir.absolutePath(), hashName(filename) + extension);
+    if (!g_AutoSavePath.isEmpty()) {
+        info(appDir.absolutePath(), hashName(filename) + extension);
+    } else {
+        info(appDir.absolutePath(), hashName(filename) + extension);
+    }
 
     if (info.exists()) {
         result = new AutoSaveFile(filename);
@@ -87,7 +94,16 @@ AutoSaveFile* AutoSaveFile::getFile(const QString &filename)
     return result;
 }
 
+void AutoSaveFile::overrideAutoSaveFilePath(QString path)
+{
+    g_AutoSavePath = path;
+}
+
 QString AutoSaveFile::path()
 {
-    return QStandardPaths::writableLocation(QStandardPaths::DataLocation) + subdir;
+    if (!g_AutoSavePath.isEmpty()) {
+        return g_AutoSavePath + subdir;
+    } else {
+        return QStandardPaths::writableLocation(QStandardPaths::DataLocation) + subdir;
+    }
 }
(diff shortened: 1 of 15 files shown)


metellius:
Just a couple of comments on the code: Passing pointers to implicitly shared objects like QDir/QString is unidiomatic as they can be passed directly by value without loss of performance (see http://doc.qt.io/qt-5/implicit-sharing.html for a list of implicitly shared classes). Most of the time you will want to pass const-refs into functions (ie. const QString&).

NULL is just an alias for 0, and then using it for comparisons with class instances like you for example do with AutoSavePath becomes a comparison between an int and QString type, which is likely not what you want (I'm a bit surprised it even works, but I guess it implicitly invokes one of the QString constructors that accept pointers... C++ will at times do a scary amount of implicit conversions without giving you any heads-up)
In addition, NULL is mostly used for C, and in C++ you just use the '0' character instead, or nullptr when using C++11.

src/autosavefile.cpp
```diff
@@ -75,8 +77,14 @@ bool AutoSaveFile::open(OpenMode openmode)
 AutoSaveFile* AutoSaveFile::getFile(const QString &filename)
 {
     AutoSaveFile* result = 0;
-    QDir appDir(path());
-    QFileInfo info(appDir.absolutePath(), hashName(filename) + extension);
+    QFileInfo info;
+
+    if (AutoSavePath != NULL) {
```
ddennedy: Never compare with NULL. use `if (!AutoSavePath.isEmpty.isEmpty())` instead.

src/autosavefile.cpp
```diff
@@ -27,6 +27,8 @@
 static const QLatin1String subdir("/autosave");
 static const QLatin1String extension(".mlt");
 
+QString AutoSavePath = NULL;
```
ddennedy: Global should static, start with g_ prefix, and initializing QString will NULL is non-sensel as QString is not a pointer.

src/autosavefile.cpp
```diff
@@ -87,7 +95,16 @@ AutoSaveFile* AutoSaveFile::getFile(const QString &filename)
...
 
+void AutoSaveFile::overrideAutoSaveFilePath(QString *path)
+{
+    AutoSavePath = path;
+}
+
 QString AutoSaveFile::path()
 {
-    return QStandardPaths::writableLocation(QStandardPaths::DataLocation) + subdir;
+    if (AutoSavePath != NULL) {
```
ddennedy: use if `(!AutoSavePath.isEmpty.isEmpty())` instead

src/database.cpp
```diff
@@ -169,12 +170,23 @@ void Database::deleteOldThumbnails()
...
 {
     connect(&MAIN, SIGNAL(aboutToShutDown()),
             this, SLOT(shutdown()), Qt::DirectConnection);
 
-    QDir dir(QStandardPaths::standardLocations(QStandardPaths::DataLocation).first());
+    QDir dir;
+    if (DatabaseDir != NULL) {
+        dir = DatabaseDir;
+    } else {
+        dir = QStandardPaths::standardLocations(QStandardPaths::DataLocation).first();
```
ddennedy: I am thinking we should invert this commit. Instead of storing several globals, setting them, and comparing against them; we should make our DataDir somewhere that is initialized from QStandardPaths, set by command line option, and then read it wherever today we read QStandardPaths::DataLocation.

metellius:
Could this be perhaps also be accomplished through some qt.conf trickery?

ddennedy:
@metellius Even if it is possible to use qt.conf, that will not work for me because I am also going to add a menu item in Settings to choose the AppData folder. This is in addition to a command line option. Settings will switch to the INI format and save/load in there. That means there will still a system-specific setting to at least define this location.

2wayne:
I made the suggested changes. I hope it is much better now. Am I missing something? Unfortunately I can't test it myself because I can't get the dependencies to work.

ddennedy:
I added this in a different manner in commit f4e1b0. The command line argument will be `--appdata` but also includes a shotcut.ini file for the settings. It only tries to move/copy data from the old folder to a new folder when you use Settings > Data Directory > Set... and not from the command line option.

PR #336, opened 2016-11-04
Small fixes

ochilan:
9cc7cdb avoids the situation where the player would switch back and forth between two producers after having opened at least two media files and successively clicking on the "Source" tab.
19b9cfb enables repairing of MLT XMLs which were produced by an instance running on a system with another locale (in my case Windows with German locale and Linux with English locale). In this case the whole file would have decimal points which are inconsistent with the locale everywhere. Previously, repairing would only be offered if both commas and periods were found within the XML.

### DIFF
--- a/src/docks/encodedock.cpp
+++ b/src/docks/encodedock.cpp
@@ -153,22 +153,18 @@ void EncodeDock::loadPresetFromProperties(Mlt::Properties& preset)
             ui->bFramesSpinner->setValue(preset.get_int("bf"));
         else if (name == "deinterlace") {
             ui->scanModeCombo->setCurrentIndex(preset.get_int("deinterlace"));
-            ui->scanModeCombo->setEnabled(false);
         }
         else if (name == "progressive") {
             ui->scanModeCombo->setCurrentIndex(preset.get_int("progressive"));
-            ui->scanModeCombo->setEnabled(false);
         }
         else if (name == "top_field_first") {
             ui->fieldOrderCombo->setCurrentIndex(preset.get_int("top_field_first"));
         }
         else if (name == "width") {
             ui->widthSpinner->setValue(preset.get_int("width"));
-            ui->widthSpinner->setEnabled(false);
         }
         else if (name == "height") {
             ui->heightSpinner->setValue(preset.get_int("height"));
-            ui->heightSpinner->setEnabled(false);
         }
         else if (name == "aspect") {
             double dar = preset.get_double("aspect");
@@ -190,12 +186,9 @@ void EncodeDock::loadPresetFromProperties(Mlt::Properties& preset)
                 ui->aspectDenSpinner->setValue(1000);
                 break;
             }
-            ui->aspectNumSpinner->setEnabled(false);
-            ui->aspectDenSpinner->setEnabled(false);
         }
         else if (name == "r") {
             ui->fpsSpinner->setValue(preset.get_double("r"));
-            ui->fpsSpinner->setEnabled(false);
         }
         else if (name == "pix_fmt") {
             QString pix_fmt(preset.get("pix_fmt"));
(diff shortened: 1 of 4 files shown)


ddennedy:
In the future, please ensure the first line of your git commit message is short. This is a general rule for git projects and not just Shotcut. I prefer <= 70 chars but git etiquette says even less. In any case, leave a more detailed message after a blank line following the short message. Thank you for contributing.

ddennedy:
So, really, the reason for the XML problem you suffered is that POSIX (and QLocale) uses ISO codes whereas MinGW setlocale() on WIndows outputs something quite different, and there is no mapping between the two. :-( I try to be conservative about repairing the XML because it can be easy to get wrong and foul things up. So, I am concerned about simply requiring the XML conform to the current system locale; however, that might actually make things work better overall. Let me think about this... Maybe, on Windows, I can require this new behavior, but on non-Windows only do this if the LC_NUMERIC attribute is not parseable by QLocale (since it was probably created on a Windows system).

ddennedy:
Let QLocale try to parse the LC_NUMERIC locale attribute, and if it fails initialize new member variable m_locale to system locale. Then, require all numeric strings with decimals to conform to m_locale. If not, fix them according to m_locale, and output new XML with LC_NUMERIC based on m_locale. What do you think? Can you try that approach and test it with your documents?

ochilan:
I haven't realized the issue with the LC_NUMERIC property within the file before, I'll think about this. However, am I correct that the code as it was before would never offer to repair the file unless it contained both commas and periods? The point is, I don't see when this would happen regardless of the locale used unless the file has been edited manually.

ddennedy:
> However, am I correct that the code as it was before would never offer to repair the file unless it contained both commas and periods?

You are correct.

> I don't see when this would happen regardless of the locale used unless the file has been edited manually.

There have been bugs in the past, particularly on Windows, where Shotcut and/or MLT could cause it to output a mixture, and this area of XML correction feature was added in direct response immediately  after some fixes.

ochilan:
I thought about this some more and did some research, and it shows that it's a mess like it always is when things like locales and different standards and pseudo-standards are involved.

The problem I see with having QLocale try to parse the LC_NUMERIC property is that on Windows, it will never be able to parse the value when it has been written by the Windows version of shotcut/MLT (where it will be something like "German_Germany.1252" in my case). So we would always "fall back" to the system locale anyway. If I'm not mistaken this will effectively make the LC_NUMERIC declaration in the XML useless and we will force it to conform to the system locale in any case.

Not sure how to fix this. I guess using some other way to determine a proper LC_NUMERIC string when writing out the XML in the first place is not an option? Surely, something like QT should not be a dependency of MLT.

ochilan:
Or was this the idea, that it should always fall back to system locale on Windows? Not exactly a nice way to handle it, but I don't currently see a better way either. Except maybe to force some locale defined by us when writing out the XML on Windows instead of using the system locale.

ochilan:
Oops, I didn't know new commits would automatically be added to an already existing pull request, I didn't really mean to add it here. But anyway, now that it is here: Note that the way the widgets are enabled/disabled is not really consistent. For instance, if you select another Video Mode from the settings menu, the resolution, FPS, etc. are entered into the Video tab of the encoding dock and are editable. If you then select a Preset from the list, they become disabled. If you then once again select another Video Mode, the values are changed but the fields are still disabled. Are they meant to be re-enabled in this case, i.e. when another Video Mode is selected?

ochilan:
Regarding the "know what you're doing": That is true. Would it probably help to e.g. de-select the chosen preset from the list once some manual change is being done? Or some other way to let the user know that the values now deviate from the chosen preset?

ddennedy:
I cherry-picked 2 of the commits and pushed a modified version of the change to MltXmlChecker. See also MLT commit https://github.com/mltframework/mlt/commit/dce7c5bf85043822d10c477640a2862d57ace3d8

Now, it should save **new** XML with LC_NUMERIC using ISO 639 and 3166 codes.

PR #367, opened 2016-12-29
Translation for Traditional Chinese, Taiwan (zh-TW)

rschiang:
This is the still-working translation of *Traditional Chinese* (zh-Hant), specifically *Taiwan* (zh-TW).

Unlike English, technical terms in Traditional Chinese differs too much from zh-CN ones, hence the importance of independent language file;

I've translated about 90% of the user interface, and will finish the remaining (mostly image processing jargons) in the following holiday.

(**Update**: I've noticed that Shotcut uses Transifex on collaboration. Please accept the language request, and I'll upload the .ts file there)

### DIFF
--- a/translations/shotcut_zh_TW.ts
+++ b/translations/shotcut_zh_TW.ts
@@ -1,0 +1,7945 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="zh_TW">
+<context>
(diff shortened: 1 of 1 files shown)


ddennedy:
I approved on transifex. I do not use direct git pull requests to update translations but will pull through tx, lupdate, and then push to git.

rschiang:
I've done all the translations on Transifex. Thanks again, and let us push Shotcut forward together! 🍺

PR #377, opened 2017-01-17
Fix #376 Mute status is more visible

### LINKED ISSUE #376: Muted/Unmuted Volume Clarity
In either of the Fusion themes, it is nearly impossible to tell whether the volume is muted or not. Besides, there isn't any indication on the volume button or elsewhere showing whether the volume is muted or not at first sight to help with it.

mutedornot

[Windows 10 Pro; Shotcut 17.01.02]

ldennedy:


src/player.cpp
```diff
@@ -151,7 +151,7 @@ Player::Player(QWidget *parent)
     m_muteButton->setFocusPolicy(Qt::NoFocus);
     m_muteButton->setObjectName(QString::fromUtf8("muteButton"));
     m_muteButton->setIcon(QIcon::fromTheme("dialog-cancel", QIcon(":/icons/oxygen/32x32/actions/dialog-cancel.png")));
-    m_muteButton->setToolTip(tr("Silence the audio"));
+    m_muteButton->setToolTip(tr("Mute/Unmute"));
```
ddennedy: Change the tooltip based on the current state: Mute or Unmute - not hyphenated.

src/player.cpp
```diff
@@ -868,7 +868,7 @@ void Player::on_actionVolume_triggered()
 {
     int x = (m_volumePopup->width() - m_volumeWidget->width()) / 2;
     x = mapToParent(m_volumeWidget->geometry().bottomLeft()).x() - x;
-    int y = 8 - m_volumePopup->height();
+    int y = 32 - m_volumePopup->height();
```
ddennedy: Use m_scrubber->geometry().height() instead of hard-coded 32, and keep the 8.

PR #387, opened 2017-02-18
Simplified Chinese translated.

Kongse:
The translation is done by using the current Simplified Chinese untranslated ts file as framework, and the Traditional Chinese translation file as the phasebook. Lots of words have been modified to fit in Simplified Chinese style. Tranlations have been checked on application.

### DIFF
--- a/translations/shotcut_zh_CN.ts
+++ b/translations/shotcut_zh_CN.ts
@@ -6,17 +6,17 @@
     <message>
         <location filename="../src/dialogs/addencodepresetdialog.ui" line="17"/>
         <source>Dialog</source>
-        <translation type="unfinished"></translation>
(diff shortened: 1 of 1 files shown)


ddennedy:
You must use transifex.com for this. I do not accept translation updates any other way at the moment because I would have to figure out a workflow that works with other translators using that service.

Kongse:
I've not known that there is a translation colabration site. It's a nice workflow, ok I will finish the zh_CN translation on it.

Kongse:
Remaining zh_CN tranlations on transifex finished.

ddennedy:
Thank you. It is nice to get this translation updated.

Kongse:
You guys made a great video editing software, very handy, thank you too.

PR #393, opened 2017-02-26
Add .desktop file for snappy package

nikalon:
Added a shotcut.desktop file for snappy package. Snapcraft can work with a `gui` directory, where you can put a .desktop file and the app's icon. If the icon is already included in the `gui` directory, there is no need to declare it in snapcraft.yaml file.

Snappy will take care of transforming the variables inside the shotcut.desktop file during installation, meaning that `${SNAP}/meta/gui/shotcut-logo-64.png` will be changed to a valid route after install.


ddennedy:
Thank you. I was just working on this, and I was making a part for the desktop file installation. It has not been clear to me from the docs about the magic of the gui subdirectory and the purpose of the icon declared in snapcraft.yaml.

ddennedy:
Ugh, this did not work!
```
ddennedy@MS-7593:~/src/shotcut/snap$ ls -l /snap/shotcut/
total 0
drwxrwxr-x 10 root root 236 Feb 12 14:14 3
lrwxrwxrwx  1 root root   2 Feb 26 11:44 current -> x1
drwxrwxr-x 10 root root 236 Feb 26 11:38 x1
ddennedy@MS-7593:~/src/shotcut/snap$ ls -l /snap/shotcut/x1/meta/gui
total 0
```

ddennedy:
I figured out the subdirectory needs to be setup/gui and fixed in commit 3a5f8dfdaccda87fab8451140690124065d9483b

PR #482, opened 2018-01-01
Add Audio Spectrum Visualization filter.

bmatherly:
Dan,
I wrote this last April. See the discussion here:
https://forum.shotcut.org/t/basic-audio-visualizateing/1332

Thoughts on releasing it?


ddennedy:
Thanks! I checked it out, and it works even in GPU mode.

PR #484, opened 2018-01-14
Avformat properties improvements.

bmatherly:
* Restore the previous tab after changes are made.
* Clear out fields that do not apply (e.g. when "none" is selcted).
* Add "All" option for audio track selection to combine all audio
  tracks into one.


bmatherly:
This is a proposed solution for:
https://forum.shotcut.org/t/audio-channel-monitoring-ch1-left-ch2-right/3846

For these applications, one can choose "All" audio tracks to get all mono tracks into Shotcut.

PR #490, opened 2018-01-30
Multichannel

bmatherly:
Adds configuration of number of channels for the player and export.

### DIFF
--- a/src/glwidget.cpp
+++ b/src/glwidget.cpp
@@ -639,6 +639,7 @@ int GLWidget::reconfigure(bool isMulti)
             m_consumer->set("0.prefill", 1);
             if (property("keyer").isValid())
                 m_consumer->set("0.keyer", property("keyer").toInt());
+            m_consumer->set("1.channels", property("audio_channels").toInt());
         }
         else {
             if (!profile().progressive())
@@ -649,6 +650,7 @@ int GLWidget::reconfigure(bool isMulti)
             m_consumer->set("prefill", 1);
             if (property("keyer").isValid())
                 m_consumer->set("keyer", property("keyer").toInt());
+            m_consumer->set("channels", property("audio_channels").toInt());
         }
         if (m_glslManager) {
             if (!m_threadStartEvent)
--- a/src/docks/encodedock.cpp
+++ b/src/docks/encodedock.cpp
@@ -62,6 +62,7 @@ EncodeDock::EncodeDock(QWidget *parent) :
 
     connect(ui->videoBitrateCombo, SIGNAL(currentIndexChanged(int)), this, SLOT(on_videoBufferDurationChanged()));
     connect(ui->videoBufferSizeSpinner, SIGNAL(valueChanged(double)), this, SLOT(on_videoBufferDurationChanged()));
+    connect(&Settings, SIGNAL(playerAudioChannelsChanged(int)), ui->channelsSpinner, SLOT(setValue(int)));
 
     m_presetsModel.setSourceModel(new QStandardItemModel(this));
     m_presetsModel.setFilterCaseSensitivity(Qt::CaseInsensitive);
@@ -152,6 +153,8 @@ void EncodeDock::loadPresetFromProperties(Mlt::Properties& preset)
                 if (ui->videoCodecCombo->itemText(i) == preset.get("vcodec"))
                     ui->videoCodecCombo->setCurrentIndex(i);
         }
+        else if (name == "channels")
+            ui->channelsSpinner->setValue(preset.get_int("channels"));
         else if (name == "ar")
             ui->sampleRateCombo->lineEdit()->setText(preset.get("ar"));
         else if (name == "ab")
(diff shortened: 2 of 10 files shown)


src/glwidget.cpp
```diff
@@ -639,6 +639,7 @@ int GLWidget::reconfigure(bool isMulti)
             m_consumer->set("0.prefill", 1);
             if (property("keyer").isValid())
                 m_consumer->set("0.keyer", property("keyer").toInt());
+            m_consumer->set("1.channels", property("audio_channels").toInt());
```
ddennedy: The multi consumer is used when using decklink for monitoring. Unfortunately, the DeckLink API only accepts the values 2, 8, or 16 which does not match our multi-channel use cases well at all. I found out because 6 was failing to initialize and looked it up. I was able to hack it to use 8, but the audio was very distorted, and my receiver could not understand it (even though it is 7.1).

Setting "1.channels" here to target decklink does not work because it is overridden by the multi consumer (all the consumers need to be in sync on this parameter). Really, for now, we need to hide the Audio Channels in the Settings menu when using a decklink for external monitoring.

bmatherly: Oh, yes. That is true. I think a more comprehensive solution (in the future) would be to extend the Decklink consumer to configure the device for as many channels as necessary to fit the requested channels - and then fill the rest of the channels with silence. This is a method I have used in the past to get around the Decklink limitations.

Your suggestion is more appropriate in the short-term. I'll look into it.

bmatherly:
I agree that channel labels would be good. I'll look into adding that.

With my MLT patch, the following channel layouts are supported:
https://github.com/bmatherly/mlt/blob/surround_convert/src/modules/core/filter_audiochannels.c#L55
I understand that there are other channel layouts in use, but these are the layouts we get from FFmpeg when we use get_default_channel_layout. So I think we would get a long way to support these and keep using get_default_channel_layout() in the producer/consumer.

I didn't think about the pan filters, etc. I should look into that to make sure there isn't a conflict.

ddennedy:
The libavutil default channel layout for 4 channels is 4.0 and not quadraphonic; yours matches quadraphonic:
https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/channel_layout.c#L84
https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/channel_layout.h#L91

Also, I do not want to expose >6 channels at this time as some important encoders, particularly ac3, do not support it:
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/ac3.h#L31

ddennedy:
When trying to test this branch on Windows against the new sdl2_audio consumer that was just merged, I only get silence with 2 or 6 channels in settings. With 2 channels in settings I get
> [Info   ] <MLT> [consumer sdl2_audio] Unable to output 2 channels. Change to 6
> [Info   ] <MLT> [consumer sdl2_audio] Audio Opened: driver=wasapi channels=6 frequency=48000

When testing with melt at command line with `-verbose` and channels=2 or 6, I get sound and the same warnings. So, I guess these APIs require to open with all channels and use which channels you need.
I do not have external monitoring through decklink enabled.

ddennedy:
Changing GLWidget::reconfigure() to prefer rtaudio over sdl2_audio, the sound works with both 2 and 6 channels. However, pause and resume makes it crash. Did you know a debug build of Shotcut integrates DrMingw, which upon crash appends to a file named "shotcut.RPT" in the same dir as the exe? Here is the dump from it for this crash:
```
Error occured on Wednesday, March 28, 2018 at 21:13:53.

shotcut.exe caused an Access Violation at location 0000000000000000 DEP violation at location 0000000000000000.

AddrPC           Params
0000000000000000 000000007A964820 000000003837A5CE 000000007E2AB130
0000000049281F8F 00000000081E9EF0 0000000100000000 0000000000000000  libmltrtaudio.dll! ??   [C:/Projects/Shotcut/src/mlt/src/modules/rtaudio/RtAudio.cpp @ 215]
   213: {
   214:   if ( rtapi_ )
>  215:     delete rtapi_;
   216: }
   217: 
0000000049294DB6 000000005F9FA5A0 0000000000000006 000000000000BB80  libmltrtaudio.dll!find_and_create_rtaudio  [C:/Projects/Shotcut/src/mlt/src/modules/rtaudio/consumer_rtaudio.cpp @ 79]
    77: 
    78: 	mlt_consumer getConsumer()
>   79: 		{ return &consumer; }
    80: 
    81: 	RtAudioConsumer()
0000000049293AB1 000000005F9FA5A0 000000007E2AB130 0000000000000001  libmltrtaudio.dll!play_audio  [C:/Projects/Shotcut/src/mlt/src/modules/rtaudio/consumer_rtaudio.cpp @ 890]
   888: }
   889: 
>  890: } // extern C
   891: 
0000000049294A05 000000005F9FA5A0 0000000000000000 0000000000000000  libmltrtaudio.dll!consumer_thread  [C:/Projects/Shotcut/src/mlt/src/modules/rtaudio/consumer_rtaudio.cpp @ 79]
    77: 
    78: 	mlt_consumer getConsumer()
>   79: 		{ return &consumer; }
    80: 
    81: 	RtAudioConsumer()
000000004928167C 000000005F9FA5A0 0000000075E003B0 0000000000000000  libmltrtaudio.dll!consumer_thread_proxy  [C:/Projects/Shotcut/src/mlt/src/modules/rtaudio/consumer_rtaudio.cpp @ 783]
   781: {
   782: 	RtAudioConsumer* rtaudio = (RtAudioConsumer*) arg;
>  783: 	rtaudio->consumer_thread();
   784: 	return NULL;
   785: }
```

bmatherly:
@ddennedy thanks for testing and reporting. I will try to fix the pause crash in the next couple of days. If I don't get it fixed before the time you need to make a release, feel free to revert the RtAudio commit from MLT.

I need to make other changes to this review before it will be ready to merge:
* Remove 4 channel configurations (only support mono, stereo and 5.1)
* Set the appropriate channel layout on the consumer (for playout and export).
* Add channel labels to the audio meter

bmatherly:
This PR is superseded by #534 #535 & #536

PR #494, opened 2018-02-06
snap build expose more shipped binaries "untested pull"

walterav1984:
Shotcut snap comes with its own melt/ffmpeg/ffprobe/etc binaries, I’d like these to become available as a global bash command to the whole system instead of only snap shotcut gui for ease in testing a current/different melt/ffmpeg version. For example:

$ shotcut #runs shotcut gui version which is already working
$ shotcut.melt #runs the melt version shipped with shotcut snap
$ shotcut.ffmpeg #runs the ffmpeg version shipped with shotcut snap

I think its not that hard to add, however I don’t know if others would like this feature?

https://forum.shotcut.org/t/shotcut-snap-could-include-commands-for-shipped-melt-ffmpeg-ffprobe-binaries/4054

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -13,6 +13,16 @@ confinement: classic
 apps:
   shotcut:
     command: bin/desktop-launch $SNAP/Shotcut.app/shotcut
+  ffmpeg:
+    command: $SNAP/Shotcut.app/ffmpeg
+  ffplay:
+    command: $SNAP/Shotcut.app/ffplay
+  ffprobe:
+    command: $SNAP/Shotcut.app/ffprobe
+  melt:
+    command: $SNAP/Shotcut.app/melt
+  qmelt:
+    command: $SNAP/Shotcut.app/qmelt
 
 parts:
   shotcut:


ddennedy:
Yes, that is a great addition to have! Is the "shotcut." prefix a snap convention that is provided automatically, because I do not see the prefixes in your patch?

walterav1984:
> Is the "shotcut." prefix a snap convention that is provided automatically,

I guess the prefix is automatically, since I didn't need to do anything special for my first dvswitch snapcraft.yaml (thanks to @diddledan)  to let "dvgrab" or "dvsource-file" be called by "dvswitch.dvgrab / dvswitch.dvsource-file"  using the same syntax as this pull request. But I haven't tested/confirmed this for shotcut.

https://github.com/walterav1984/dvswitch/blob/ubuntu-16.04-master/snap/snapcraft.yaml

ddennedy:
From the [docs](https://docs.snapcraft.io/build-snaps/syntax#apps):

> If the app name matches the snap name, the app will be executed when the name of the snap is called. Otherwise, the executable will have the snap name and a dot prefixed to the app name.

walterav1984:
This snap/command name matching confirms may assumptions for the prefix thanks for pointing to the docs.

As far as snap "classic" confinement I'm looking forward to check if $shotcut.ffmpeg will work in current future ubuntu/ other linux distro's without complaining about missing libs.

ddennedy:
This is back in and working, but it is only in the candidate, beta, and edge channels where 18.05 resides.

PR #496, opened 2018-02-10
Allow finer grained control of color grading filter.

bmatherly:
See the discussion here:
https://forum.shotcut.org/t/color-grading-filter-improvement/3800


src/qml/filters/color/ui.qml
```diff
@@ -28,19 +28,16 @@ Item {
     width: 450
     height: 250
```
ddennedy: The height needs to increase to make canvas larger for proper scrolling when the filter dock is small and shows scroll bars.

bmatherly: Fixed .

PR #512, opened 2018-03-04
Add Mosaic filter.

bmatherly:
As requested here:
https://forum.shotcut.org/t/is-there-a-mosaic-filter/4134

### DIFF
--- a/src/qml/filters/mosaic/meta.qml
+++ b/src/qml/filters/mosaic/meta.qml
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2018 Meltytech, LLC
+ * Author: Brian Matherly <code@brianmatherly.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+ 
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+    type: Metadata.Filter
+    name: qsTr("Mosaic")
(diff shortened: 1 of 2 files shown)


src/qml/filters/mosaic/meta.qml
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+ 
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+    type: Metadata.Filter
+    name: qsTr("Mosaic")
```
ddennedy: Why not Pixelize instead of Mosaic? People may expect something more artistic from Mosaic.

src/qml/filters/mosaic/meta.qml
```diff
@@ -0,0 +1,27 @@
...
+    mlt_service: "frei0r.pixeliz0r"
```
ddennedy: The obscure filter is similar and has a rectangle parameter but lacks size parameters. Hmm

bmatherly:
Good questions. I hadn't considered that some people might think of those video/picture mosaics where many videos/images combine to make a larger picture.

I was just thinking about the Premiere "Mosaic" effect:
https://www.youtube.com/watch?v=bCVNk_9470Y

I'm mostly interested in the mosaic filter as an artistic effect. But people can combine it with a mask to blur faces if they want (as demonstrated in the Premiere video).

I hesitate to make a "face hiding" filter because the first thing people will ask for is that it track faces.

I'd like to think about this a little more before we commit the "Mosaic" name to this filter. Maybe I'll write something from scratch.

ddennedy:
Why write from scratch? Unless it is being ported to Movit.
It is good to know Premiere calls it mosaic. I see Avid does as well. Then, I am fine with it.

bmatherly:
> Why write from scratch? Unless it is being ported to Movit.

The only reason to write from scratch would be to combine the two features: ability to specify "pixel" size and the ability to specify a region. After a little more research, I don't feel motivated to do that. The mosaic + mask filter should be a good workflow for hiding faces - plus, people can use the mask with blur if they want blurred faces instead of pixelized.

> It is good to know Premiere calls it mosaic. I see Avid does as well. Then, I am fine with it.

Gimp has a very similar filter called "Pixelize". I you prefer "Pixelize", I would be happy to change it. If you prefer "Mosaic", then go ahead and merge this as is. I don't feel strongly either way.

ddennedy:
If we call it Pixelize, surely someone will complain that it should be Pixelate. :-S

PR #515, opened 2018-03-10
Added Russian translation to the shortcut

TotalCaesar659:


ddennedy:
I only like to take translations through transifex.com. All other translations are received through that. It facilitates collaboration. Does that work ok for you?

TotalCaesar659:
There was no Transifex in readme, so here the fix for it (https://github.com/mltframework/shotcut/pull/517).
P.S. You're right about common translation strings, but as I know, Transifex doesn't support .desktop files. So the authors of many projects that I've translated had accepted my PRs to them, even if Transifex is available to their project.

ddennedy:
Yes, of course, regarding desktop files. I got the two pull requests mixed up when reviewing them on my phone. :)

PR #516, opened 2018-03-10
Updated Russian translation

TotalCaesar659:


ddennedy:
Please submit them through transifex so they are not clobbered the next time I pull from transifex.

PR #525, opened 2018-03-26
Build Shotcut with Travis, also build shotcut appimage

gntech:
This pull request automates the build of Shotcut from the docker image.

It builds both the default .tar.bz2-archive plus an additional AppImage which is very similar to the tar.bz2 but with added functionality. It uses a squash-filesystem internally and does not need to be unpacked. Just set the exectuable bit and double-click. Very convinient. :)
More info here [AppImage](https://appimage.org/)

When a new release is tagged on Github, Travis automatically uploads the .tar.bz2 and .AppImage to the corresponding releases section on Github.

What is needed for this to work:
* A token needs to be created on github and uploaded to travis-ci.org to allow the upload.

Further work:
* Initially I tried to automate also automatically build snap package and push to snap repo but I have decided to leave that out of the scope for now. I think this could be added quite easily but I don't have the time right now to do the research.
* Right now Travis builds from the docker image, however there is no correlation between the git commit that Travis started the build from and the actual commit used in the build. The docker image pulls shotcut from github ~3 minutes after the build has started. This means that if there are several commits in a row there is a possibility that Travis builds a later commit. This could be fixed if the dockerimage was set to build the local source that is automatically pulled by Travis instead of pulling the source in the docker image.
* Windows build files could also be built and uploaded to the tagged release.
* You could possible split the build load between AppVeyor Circle-CI and Travis.
    * AppVeyor can be used to build in a windows environment.
    * Circle CI can build in a MacOS environment.
* Consider to use Qt 5.6.2 instead of 5.6.1 then you could use prebuilt libraries from the above providers.

See, https://www.appveyor.com/docs/build-environment/#qt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,39 @@
+sudo: required
+language: generic
+dist: trusty
+addons:
+  apt:
+    packages:
+    - squashfs-tools
+    - desktop-file-utils
+
+services:
+  - docker
+
+before_install:
+- wget --no-check-certificate https://raw.githubusercontent.com/mltframework/mlt-scripts/master/docker/shotcut-build/Dockerfile
+- mkdir shotcut-build
+- mv Dockerfile shotcut-build/
+- docker build --rm -t ddennedy/shotcut-build shotcut-build
+- mkdir work; cd work
+- wget --no-check-certificate https://raw.githubusercontent.com/mltframework/shotcut/master/scripts/build-shotcut.sh
+- wget --no-check-certificate https://raw.githubusercontent.com/mltframework/shotcut/master/scripts/shotcut.nsi
+- wget --no-check-certificate https://raw.githubusercontent.com/AppImage/AppImages/master/pkg2appimage
+
+script:
+- docker run -it --rm -v $PWD:/root/shotcut ddennedy/shotcut-build
+- bash -ex pkg2appimage ../scripts/shotcut-recipe.yml
+- mv shotcut.tar.bz2 shotcut-${TRAVIS_TAG}.tar.bz2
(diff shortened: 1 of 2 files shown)


ddennedy:
We already have automated nightly build of the portable tarball. All this does is change some tools and automate upload. Uploading is trivial. 

The advantage of a tarball over AppImage is that it exposes some command line utilities without having to figure out the AppImage mount. Still, AppImage would be a nice addition.

Our Qt builds require QtWebKit, and Windows uses 32- and 64-bit mingw-w64 per the toolchain we use for both Windows development and cross-compiling. Attempting to switch to msvc and msvcrt is a huge change to some of our dependencies, compiler compatibility, toolchains for dev and debug, and our [SDK and its instructions](https://shotcut.com/notes/windowsdev/). There is a lot to consider. It does not appear that AppVeyor satisfies those requirements.

I appreciate your intention, but you need to understand this from my perspective. You seem to want to change the build system without taking ownership. It is considerable effort to change and detracts from the work on bugs and features. Still, I admit, I have thought about similar things, and if I decide to overhaul the build system in the future, I will come back to this.

gntech:
Thanks for your response. I totally see your point about your build toolchain. :)

Glad you liked the appimage though, the recipe is slightly modified from the one in the official github.com/appimage/appimages repo. Appimages are more and more used and I think they are very convinient for the users. Kdenlive, openshot, freecad, digikam, krita, etc all uses appimages.

If you want to be able to call CLI tools integrated to the appimages it can be solved.

Freecad uses the following solution to call a secondary cli-tool
`./freecad.AppImage -c`
See https://forum.freecadweb.org/viewtopic.php?f=22&t=24719

OpenToonz uses the following solution to exectue an integrated tool called tconverter:
`./OpenToonz.appimage --appimage-exec tconverter ARGUMENTS`
See https://morevnaproject.org/2016/11/27/use-opentoonz-cli/

PR #551, opened 2018-05-08
Additional GPU checking and clarifications.

bmatherly:
* Use "GPU Effects" instead of "GPU Processing" so people don't assume
  it refers to encoding.
* Add additional information about the limitations of "GPU Effects"
  and possible side-effects in the information dialog.
* Check for projects that use filters that are incompatible with GPU
  effects when GPU Effects are enabled.


ddennedy:
I might do another release in a day or two and hold off on this non-critical change that will break a few translated strings.

bmatherly:
Makes sense.

I've noticed a lot of issues on the Forum being solved by people turning off "GPU Processing". So I thought this might help out. Feel free to suggest wording changes.

PR #560, opened 2018-05-14
improve packaging

eszlari:
- new tree structure for packaging: linux/macos/windows
- merge .desktop files in 3 different places into 1
- qmake: install linux specific meta files
- linux: add mimetype file
- appdata: shorten summary for better display on flathub.org website
- linux: meta data naming according to AppStream specification:
https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html


packaging/linux/snap/package.mak
```diff
@@ -0,0 +1,10 @@
+snap:
+	snapcraft prime
+	cp ../../../icons/shotcut-logo-64.png gui
+	cp ../org.shotcut.Shotcut.desktop gui/shotcut.desktop
+	desktop-file-edit --set-key=Icon --set-value='${SNAP}/meta/gui/shotcut-logo-64.png'
+	snapcraft snap
```
ddennedy: I have to use [cleanbuild](https://docs.snapcraft.io/reference/snapcraft-command#cleanbuild). Is there a way to support these additional commands? I could alter my process to use a docker container instead of cleanbuild. I will give that a try before deciding.

src/src.pro
```diff
@@ -284,13 +284,16 @@ OTHER_FILES += \
     shotcut.rc \
     ../scripts/build-shotcut.sh \
     ../icons/shotcut.icns \
-    ../scripts/shotcut.nsi \
-    ../Info.plist \
+    ../packaging/windows/shotcut.nsi \
+    ../packaging/macos/Info.plist \
```
ddennedy: QMAKE_INFO_PLIST in src.pro needs to change for this.

packaging/linux/org.shotcut.Shotcut.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<mime-info xmlns='http://www.freedesktop.org/standards/shared-mime-info'>
+  <mime-type type="application/vnd.mlt-project">
```
ddennedy: Shotcut uses "application/mlt+xml" to identify some things placed on the clipboard. And then, for whatever reason - probably lack of detailed scrutiny - Info.plist uses "text/mlt+xml". Let's switch everything over to "application/vnd.mlt+xml". I will take care of the Shotcut source code changes for that. "+xml" is preferred over "-xml" because it is a [suffix](https://en.wikipedia.org/wiki/Media_type#Suffix).

packaging/linux/org.shotcut.Shotcut.xml
```diff
@@ -0,0 +1,8 @@
...
+    <comment>MLT project</comment>
```
ddennedy: I consistently use "MLT XML" as the standard name for these files in Shotcut, Info.plist, and both the Shotcut and MLT websites.

eszlari:
I must confess that I have no clue about snap and just copied this from the Internet as a starting point.

packaging/linux/snap/package.mak
```diff
@@ -0,0 +1,10 @@
+snap:
+	snapcraft prime
+	cp ../../../icons/shotcut-logo-64.png gui
+	cp ../org.shotcut.Shotcut.desktop gui/shotcut.desktop
+	desktop-file-edit --set-key=Icon --set-value='${SNAP}/meta/gui/shotcut-logo-64.png' gui/shotcut.desktop
+	snapcraft snap
+
+snap-clean:
+	snapcraft clean
+	rm -rf snap
```
ddennedy: desktop-file-edit fails with an error:
```
snap/gui/shotcut.desktop: error: value "${SNAP}/meta/gui/shotcut-logo-64.png" for key "Icon" in group "Desktop Entry" looks like a relative path, instead of being an absolute path to an icon or an icon name
```

ddennedy: This approach with snapcraft prime and snap works for the icon (if you target the directory prime/meta/gui), but snapcraft is deleting the desktop file from both the prime dir and the installed snap! Otherwise, the ideal place for these commands is in an [override-stage](https://docs.snapcraft.io/build-snaps/scriptlets) key in the part.

However, I am struggling to get that to work. What does work is moving package.mak into the parent directory with this:
```
snap:
	mkdir -p snap/meta/gui
	cp ../../icons/shotcut-logo-64.png snap/gui
	cp org.shotcut.Shotcut.desktop snap/gui/shotcut.desktop
	sed -i 's|Icon=.*|Icon=${SNAP}/meta/gui/shotcut-logo-64.png|' snap/gui/shotcut.desktop
	snapcraft
```
Nice thing about that is I can add a target for the appimage as well.

PR #565, opened 2018-05-22
Use YAML for filter presets.

bmatherly:
This is an alternative proposal to https://github.com/mltframework/mlt/pull/340

This logic could also be moved into the Mlt::Properties save() and load() functions since it maintains backwards compatibility.

### DIFF
--- a/src/qmltypes/qmlfilter.cpp
+++ b/src/qmltypes/qmlfilter.cpp
@@ -259,7 +259,17 @@ int QmlFilter::savePreset(const QStringList &propertyNames, const QString &name)
             dir.cd(objectNameOrService());
     }
     const QString preset = name.isEmpty()? tr("(defaults)") : name;
-    properties.save(dir.filePath(preset).toUtf8().constData());
+    // Convert properties to YAML string.
+    char* yamlStr = properties.serialise_yaml();
+    QString yaml = yamlStr;
+    free(yamlStr);
+    // Save YAML to file
+    QFile yamlFile(dir.filePath(preset));
+    if(!yamlFile.open(QIODevice::WriteOnly)) {
+        LOG_ERROR() << "Failed to save preset: " << dir.filePath(preset);
+    }
+    yamlFile.write(yaml.toUtf8());
+    yamlFile.close();
     loadPresets();
     return m_presets.indexOf(name);
 }
@@ -458,7 +468,26 @@ void QmlFilter::preset(const QString &name)
 
     if (!dir.cd("presets") || !dir.cd(objectNameOrService()))
         return;
-    m_filter.load(dir.filePath(name).toUtf8().constData());
+
+    // Detect the preset file format
+    bool isYaml = false;
+    QFile presetFile(dir.filePath(name));
+    if(presetFile.open(QIODevice::ReadOnly)) {
+        if(presetFile.readLine(4) == "---") {
+            isYaml = true;
+        }
+        presetFile.close();
+    }
+
+    if(isYaml) {
+        // Load from YAML file.
+        QScopedPointer<Mlt::Properties> properties(Mlt::Properties::parse_yaml(dir.filePath(name).toUtf8().constData()));
+        m_filter.inherit(*properties);
+    } else {
+        // Load from legacy preset file
+        m_filter.load(dir.filePath(name).toUtf8().constData());
+    }
+
     emit changed();
 }
 


ddennedy:
> This logic could also be moved into the Mlt::Properties save() and load() functions since it maintains backwards compatibility.

I do not like that idea because I prefer mlt++ to be a simple wrapper over the C API.

src/qmltypes/qmlfilter.cpp
```diff
@@ -259,7 +259,17 @@ int QmlFilter::savePreset(const QStringList &propertyNames, const QString &name)
...
     const QString preset = name.isEmpty()? tr("(defaults)") : name;
-    properties.save(dir.filePath(preset).toUtf8().constData());
+    // Convert properties to YAML string.
+    char* yamlStr = properties.serialise_yaml();
+    QString yaml = yamlStr;
+    free(yamlStr);
+    // Save YAML to file
+    QFile yamlFile(dir.filePath(preset));
+    if(!yamlFile.open(QIODevice::WriteOnly)) {
+        qDebug() << "Failed to save preset: " << dir.filePath(preset);
```
ddennedy: Use LOG_ERROR() instead of qDebug() as that uses our logging framework.
It would be nice at this point to let the user know, but from this point in the code that is difficult. :-(

src/qmltypes/qmlfilter.cpp
```diff
@@ -458,7 +468,27 @@ void QmlFilter::preset(const QString &name)
...
+        if(presetFile.readLine(4) == "---") {
+            isYaml = true;
+        }
+        presetFile.close();
+    }
+
+    if(isYaml) {
+        // Load from YAML file.
+        Mlt::Properties* properties = Mlt::Properties::parse_yaml(dir.filePath(name).toUtf8().constData());
+        m_filter.pass_values(*properties, "");
```
ddennedy: m_filter.inherit() is preferred here. Yeah, I know the name "inherit" kinda sucks.

src/qmltypes/qmlfilter.cpp
```diff
@@ -458,7 +468,27 @@ void QmlFilter::preset(const QString &name)
...
+        delete properties;
```
ddennedy: Preference now is to use smart pointers:
`QScopedPointer<Mlt::Properties> properties(Mlt::Properties::parse_yaml(dir.filePath(name).toUtf8().constData()));`

bmatherly:
>> This logic could also be moved into the Mlt::Properties save() and load() functions since it maintains backwards compatibility.
> 
> I do not like that idea because I prefer mlt++ to be a simple wrapper over the C API.

I agree with that comment. My main point is that is the old key=value style is deprecated and YAML is preferred going forward, the API is not convenient to support that (just look how much extra code it takes to use YAML). So that will not encourage applications to move from the deprecated style to YAML.

I could also move it into mlt_properties_load() and mlt_properties_save(). Or create new "yaml" version of those functions. Just thinking out loud here. I don't feel strongly about it.

src/qmltypes/qmlfilter.cpp
bmatherly: done.

src/qmltypes/qmlfilter.cpp
bmatherly: done

src/qmltypes/qmlfilter.cpp
bmatherly: done

ddennedy:
Waiting a little to merge while I continue to think about your proposals. Some good points but I am also a little anxious about change - not so much the load side but on save. I worry that some app is using save and then parsing the results in their own code, e.g. custom preset handling.

bmatherly:
> Waiting a little to merge while I continue to think about your proposals.

That is fine. No rush. We could also bake this code in Shotcut for a while and then migrate it into MLT some day in the future (if we remember to do that).

ddennedy:
Do not merge this yet. I found during testing that it breaks the new Size and Position presets with keyframes.

bmatherly:
> I found during testing that it breaks the new Size and Position presets with keyframes.

The problem is caused by keys that have a colon in them (like "shotcut:animIn"). See my proposed solution here:
https://github.com/mltframework/mlt/pull/343

I also considered quoting the key string, but that would involve more complex changes.

PR #566, opened 2018-05-22
Add Hue filter

vhodde:
Did a copy of glow (blur) filter and modified it to call hue functionalities (frei0r only)
Tested on Windows 7 only

### DIFF
--- a/src/qml/filters/hue/ui_frei0r.qml
+++ b/src/qml/filters/hue/ui_frei0r.qml
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2014-2015 Meltytech, LLC
+ * Author: Brian Matherly <pez4brian@yahoo.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.1
+import QtQuick.Controls 1.1
+import QtQuick.Layouts 1.0
+import Shotcut.Controls 1.0
+
+Item {
+    property string paramHue: '0'
+    property var defaultParameters: [paramHue]
+    width: 350
+    height: 50
+    Component.onCompleted: {
+        filter.set('start', 1)
+        if (filter.isNew) {
+            // Set default parameter values
+            filter.set(paramHue, 0.0 / 100.0)
+            filter.savePreset(defaultParameters)
+            bslider.value = filter.getDouble(paramHue) * 100.0
+        }
+    }
+
+    GridLayout {
+        columns: 3
+        anchors.fill: parent
+        anchors.margins: 8
+
+        Label {
+            text: qsTr('Preset')
+            Layout.alignment: Qt.AlignRight
+        }
+        Preset {
+            Layout.columnSpan: 2
+            parameters: defaultParameters
+            onPresetSelected: {
+                bslider.value = filter.getDouble(paramHue) * 100.0
+            }
+        }
+
+        Label {
+            text: qsTr('Hue')
+            Layout.alignment: Qt.AlignRight
+        }
+        SliderSpinner {
+            id: bslider
+            minimumValue: 0
+            maximumValue: 100
+            suffix: ' %'
+            value: filter.getDouble(paramHue) * 100.0
+            onValueChanged: filter.set(paramHue, value / 100.0)
+        }
+        UndoButton {
+            onClicked: bslider.value = 0
+        }
+
+        Item {
+            Layout.fillHeight: true
+        }
+    }
+}
--- a/src/qml/filters/hue/meta_frei0r.qml
+++ b/src/qml/filters/hue/meta_frei0r.qml
@@ -1,0 +1,9 @@
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+    type: Metadata.Filter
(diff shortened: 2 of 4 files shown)


src/qml/filters/hue/ui_frei0r.qml
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2014-2015 Meltytech, LLC
+ * Author: Brian Matherly <pez4brian@yahoo.com>
```
bmatherly: I am not the author :)

vhodde:
Let me try with avformat tomorrow. From a user friendliness perspective, it might be better as most softwares group those 3 settings together (I don't use them much so can't say from my own experience). And this way we can remove the existing "saturation" filter.

src/qml/filters/hue/ui_frei0r.qml
vhodde: Haha, sorry, I am new to this. Actually what should I put in Copyright? and do I have to put an email?

ddennedy: Copyright (c) 2018 Meltytech, LLC

And you can leave out the authorship info. We should start doing that, in general, Brian as we have git log and typically multiple authors.

bmatherly: I was thinking the same thing. Copyright is important, author is not.

src/qml/filters/hue/ui_frei0r.qml
```diff
@@ -0,0 +1,76 @@
...
+import QtQuick.Layouts 1.0
+import Shotcut.Controls 1.0
+
+Item {
+    property string paramHue: '0'
+    property var defaultParameters: [paramHue]
+    width: 350
+    height: 50
+    Component.onCompleted: {
+        filter.set('start', 1)
```
ddennedy: There is no 'start' property on this filter. You can remove that line.

ddennedy:
> Okay, I have done the Hue/Lightness/Saturation filter with avformat (screenshot below). 

We strongly shy away from using bare values in parameters. They should be mapped into percentage or have some unit of measure indicated in the UI.
Does this have a preset? I will not put you on the hook for adding keyframes support, but the preset should be included.

> what should we do with the existing "Saturation" filter

Good question. One option is to mark it hidden in the meta.qml. That prevents it from the showing in the filter chooser for new projects, but it will still load for old projects. Another option is to just keep it visible because people are looking for Saturation because they have seen it web pages and video tutorials. 

When the time comes to add search and categories to the filter chooser, I have been thinking about whether to let a filter have aliases and appear in the list more than once. For example, this could show up as 3 filters all linked to the same filter - if you see what I mean. Of course, that is not available today, and I am still undecided. What is your opinion on the two options above?

vhodde:
> We strongly shy away from using bare values in parameters.

Okay I will update the plugin with percentage (this weekend). I will upload changes once done.

> What is your opinion on the two options above?

I would rather choose 2nd option: leave it for the moment. It can be easily removed in the future when the new filter is there or when we find another idea. We better keep users' habits unchanged for the moment.

ddennedy:
One major drawback to using avfilter is that it does not support keyframes, and I am not certain that it can be added (supported by libavilter, in general). I think that may be a strong case for using frei0r.hueshift0r, which does support it. We have an easy change to make to MLT frei0r module, and then all frei0r modules will support keyframes.

bmatherly:
> One major drawback to using avfilter is that it does not support keyframes, and I am not certain that it can be added (supported by libavilter, in general).

I think that keyframes could be supported by the avfilter filter by destroying the filter and recreating it between each frame. That is not ideal, but I don't think the performance impact would be very noticeable. Of course, this would not work for a specific filter if it relied on previous cached frame data - none come to mind that have that requirement.

ddennedy:
> I think that keyframes could be supported by the avfilter filter by destroying the filter and recreating it 

I am pessimistic about that approach.

bmatherly:
>>I think that keyframes could be supported by the avfilter filter by destroying the filter and recreating it

>I am pessimistic about that approach.

That's a fair comment. In order to animate properties, the entire filter graph would have to be recreated every time a property value changes - which could be every frame in the worst case. I emulated this by setting "pdata->reset = 1;" in filter_get_image(). When I tested this, the filtergraph was recreated with every video frame. The performance was actually OK on my machine. But the problem is that the filter graph creation is noisy on the console and there were a lot of messages spamming the log. So, I think we could make it work, but it might be clunky.

For the simple use case of color correction, I don't think keyframes would be important for this filter. But if the filter had keyframes, some users might find a way to use it to add artistic effects. So I guess there could be some value there.

vhodde:
Hello, I changed "raw" values with percentage as suggested (screenshot below): Hue and Lightness goes from 0 to 200 % and Saturation goes from 0 to 500 % (actually can go up to 1000 % but not sure if it's really useful to go that high).

Regarding your discussion with keyframes, I haven't used this feature yet so can't help you much. Just let me know if you are still okay to add this filter as it is or go for frei0r...

image

ddennedy:
I will cherry-pick your commits for the avfilter since this branch.
Other reasons to keep the existing saturation filter is it has a GPU version (these do not and not trivial to add), and I am adding support for keyframes to them. Keyframes for hue is less important.

ddennedy:
partially merged in aa374f536

vhodde:
Thank you. Happy to see that my first (small) contribution to open source software can be added live :)

ddennedy:
@richardpl not in MLT's `avfilter` module. I offered to pay you to work on it but then never heard back (except for your suggestion how **I** can do it). Are you interested to be paid to work on that?

richardpl:
How to animate filter properties via melt?

bmatherly:
> How to animate filter properties via melt?

I see the documentation here is a little dated:
https://www.mltframework.org/docs//propertyanimation/

In general, the text format is <time>=<value> separated by ";"

Example:

`melt.exe color:red -filter brightness level=00:00:00.400=1.71111;00:00:00.983=0.22963;00:00:01.583=1.55556;00:00:02.700=0.316543;00:00:03.267=1.7037;00:00:03.983=0.355556`

The time can  be either a time format string (as in my example above) or a raw frame number like this:
`melt.exe color:red -filter brightness level=10=1.71111;30=0.22963;45=1.55556;90=0.316543;120=1.7037;150=0.355556`

PR #613, opened 2018-07-27
Fixes for changing profile.

bmatherly:
The project and player get messed up when changing between profiles with different frame rates.


ddennedy:
It looks like `Player::onShowFrame` is some remnant of the remove melted/mvcp code.

bmatherly:
> It looks like Player::onShowFrame is some remnant of the remove melted/mvcp code.

In retrospect, I wish that I had made that change in a separate commit.

PR #619, opened 2018-08-19
Timer filter

bmatherly:
This pull request adds a new "timer" filter. It also requires https://github.com/mltframework/mlt/pull/362

I factored out some common code from the "Text" filter. But I didn't modify the text filter ui.qml file to reduce the risk of a regression. The text filter qml could be easily modified to use the common code in a future commit.


bmatherly:
screenshot from 2018-08-19 18-30-53

PR #630, opened 2018-09-09
Fix unenclosed <b> tag in TextProducerWidget

rschiang:
Found during translation. :P


ddennedy:
Since the closing tag comes at the end of the string, and I want to release in a day or two; I will wait to merge this after the next release because I do not want to break any existing translations this close to release.

PR #635, opened 2018-09-24
Add grid overlay to player.

bmatherly:
Here is an implementation of a grid overlay for your consideration.

I have thought about how snapping could be implemented. The RectangleControl could be modified to identify snap locations by accessing the video.grid property and rounding to the nearest snap location when onReleased is called.

Other debatable topics I can think of:
* What color should the grid be? I chose gray, but maybe white or black?
* What are the best grid options? I chose 2, 3 ,4 ,16.
* tried to test all the zoom scenarios that I could. I think it works, but it was tricky and there could be a bug in there.


ddennedy:
I initially chose 4x4, and that is what I see on screen; however, the menu showed 2x2 checked. Toggle it off and back on again, it displays as 2x2.
As for color, I think a white and black line along-side each might work good (white with black shadow?). The gray is fairly difficult to see. I can hear users now complaining that there are faint lines on their video that they cannot get rid off! ;-) Perhaps another option is to add a sub-menu with a color choice of a few primaries in addition to black, white, gray.
I wonder if we should include the outside border of the grid or not. I was not expecting to see it.

ddennedy:
Here it is as white with black for consideration (sans outline):
image
```
diff --git a/src/qml/modules/Shotcut/Controls/VuiBase.qml b/src/qml/modules/Shotcut/Controls/VuiBase.qml
index f1298ee5..5265a345 100644
--- a/src/qml/modules/Shotcut/Controls/VuiBase.qml
+++ b/src/qml/modules/Shotcut/Controls/VuiBase.qml
@@ -20,11 +20,29 @@ DropArea {
                 var rectY = video.rect.y - video.offset.y + zoomOffsetY
                 var gridSizeX = rectW / video.grid
                 var gridSizeY = rectH / video.grid
-                ctx.lineWidth = 2
-                ctx.strokeStyle = "#777777"
+
+                ctx.lineWidth = 1
+                ctx.strokeStyle = "#000000"
+                ctx.beginPath()
+                // vertical grid lines
+                for(var x = 1; x * gridSizeX < rectW; x++)
+                {
+                    context.moveTo(rectX + x * gridSizeX + 1, rectY)
+                    context.lineTo(rectX + x * gridSizeX + 1, rectY + rectH)
+                }
+                // horizontal grid lines
+                for(var y = 1; y * gridSizeY < rectH; y++)
+                {
+                    context.moveTo(rectX, rectY + y * gridSizeY + 1)
+                    context.lineTo(rectX + rectW, rectY + y * gridSizeY + 1)
+                }
+                // draw and close
+                context.stroke()
+                context.closePath()
+
+                ctx.lineWidth = 1
+                ctx.strokeStyle = "#ffffff"
                 ctx.beginPath()
-                // Outline
-                ctx.rect(rectX + 1, rectY + 1 , rectW - 2, rectH - 2)
                 // vertical grid lines
                 for(var x = 1; x * gridSizeX < rectW; x++)
                 {
```

ddennedy:
Here is the same video with your 2-wide gray for comparison:
image

ddennedy:
Where I really notice a difference is this clip. Before:
image

After:
image

PR #638, opened 2018-09-26
Add snap to grid for VUI.

bmatherly:
Here is a implementation of "snap to grid" for your consideration.

The snapMargin can be modified to change how quickly the line snaps to the grid.

### DIFF
--- a/src/qml/modules/Shotcut/Controls/RectangleControl.qml
+++ b/src/qml/modules/Shotcut/Controls/RectangleControl.qml
@@ -51,6 +51,38 @@ Item {
         bottomLeftHandle.y = bottomRightHandle.y
     }
 
+    function snapX(x) {
+        if (video.grid == 0) {
+            return x
+        }
+        var gridSize = parent.width / video.grid
+        var snapMargin = gridSize / 5
+        var delta = x % gridSize
+        if (delta < snapMargin) {
+            return x - delta
+        } else if ((gridSize - delta) < snapMargin) {
+            return x + gridSize - delta
+        } else {
+            return x
+        }
+    }
+
+    function snapY(y) {
+        if (video.grid == 0) {
+            return y
+        }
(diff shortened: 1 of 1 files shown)


ddennedy:
Overall, it is good and handy! I think the snapping is a little too strong even though you can argue it is easy to turn grid off. Some people will want the ability to show the guide but turn off snapping. I think we can add a checkbox to the grid menu for that option, and I can add support for the safe areas I already added to the grid menu.

bmatherly:
Thanks for the comments. I don't have an emotional attachment to this review. Are you interested in taking over it? Otherwise, I am willing to implement your suggestions and submit changes.

ddennedy:
I'm working on it since I added the safe zones, but this now highlights how inaccurate the rectangle control is when zoomed in :-( The snapping and image accurate with respect to the grid but not the rectangle control itself.

bmatherly:
Great!

I'll be happy to stay out of your way. Let me know if you want help with anything :)

ddennedy:
Should the new "Snap to Grid" option also snap to safe area guides? I ask because technically they are not a grid.

ddennedy:
Also, when thinking about that, consider that when positioning (not resizing), the rectangle edges do not snap - only the center does.

bmatherly:
> Should the new "Snap to Grid" option also snap to safe area guides? I ask because technically they are not a grid.

I would say "yes". From a broadcast perspective, the safe areas would be used for logo placement - and snapping to them would be useful. Maybe "Snap to Grid" could just be "Snap"? or "Snap to lines"?

> Also, when thinking about that, consider that when positioning (not resizing), the rectangle edges do not snap - only the center does.

True. I think that would be useful, but less important because it is not hard to grab the outside of the rectangle to snap the edges to the grid. I would use the outside of the rectangle to position the outside and I would use the center positioner to "center" the rectangle. I played with that and thought it was fine.

PR #643, opened 2018-10-01
Add "Remove Spot" filter.

bmatherly:
I tried this filter in various scenarios and I think it will be pretty useful. However, there are some issues that I  don't like:

1) The avfilter filters do not support keyframes. I expect people would like to be able to keyframe this so that they could remove a spot that is moving.

2) avfilter.delogo uses absolute pixel positions. This will cause trouble when the project resolution is changed. I know we have that problem now with some other filters. But those will all be resolved when we implement a general solution. This filter could not be fixed with a general solution and would always have this problem unless we implement some kind of filter-specific solution.

3) avfilter.delogo is very sensitive about the x/y/w/h parameters. If they go outside of the 1 pixel boundary, the filter breaks. This limitation forced me to add more complicated bounds checking to the code. Not terrible, but would rather not have to worry about it.

Feel free to review this pull request. But don't merge it yet. I want to take some time to assess the effort to port the avfilter.delogo code to MLT so that it can use rect properties and keyframing to fix the issues I mentioned above.

### DIFF
--- a/src/qml/filters/delogo/meta.qml
+++ b/src/qml/filters/delogo/meta.qml
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2018 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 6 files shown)


ddennedy:
I found it will crash (using Windows) if the rectangle is nearly the size of image. I see messages like the following:
```
[filt @ 000000005c6c70c0] auto-inserting filter 'auto_scaler_1' between the filter 'in' and the filter 'filt'
[auto_scaler_1 @ 000000005c6c80c0] w:1280 h:720 fmt:yuyv422 sar:1/1001 -> w:1280 h:720 fmt:yuv422p sar:1/1001 flags:0x2
[auto_scaler_0 @ 000000005c6c6bc0] w:1280 h:720 fmt:yuv422p sar:1/1001 -> w:1280 h:720 fmt:yuyv422 sar:1/1001 flags:0x2
```
That suggests like the filter is rather complex and might be difficult to port. If we move forward with this, we should constrain the rectangle to something like 20% of profile resolution.

bmatherly:
Closing in favor of this implementation https://github.com/mltframework/shotcut/pull/650

PR #645, opened 2018-10-05
Grid & snap improvements

bmatherly:
Per discussion here:
https://forum.shotcut.org/t/v18-10-beta-program-and-download/7510/15


ddennedy:
Thank you!

PR #650, opened 2018-10-14
Add Spot Remover filter.

bmatherly:
Alternative implementation to https://github.com/mltframework/shotcut/pull/643

Requires https://github.com/mltframework/mlt/pull/373


ddennedy:
It works beautifully! Thank you

PR #656, opened 2018-11-03
Move installation of system libraries.

bmatherly:
I like to use the build script for my development workflow. But copying the system libraries takes a long time. I think the system libraries should only be copied when creating a package. I'm not sure if the ARCHIVE parameter is the right switch to trigger the copy.


ddennedy:
That is fine and I think ARCHIVE is the correct flag as it corresponds to packaging.

PR #657, opened 2018-11-03
Enable video waveform scope.

bmatherly:
Now that the GPU effects mode is a hidden advanced feature, I would propose that we enable the video waveform scope when GPU effects are disabled. I would like to improve the video waveform scope and add more scopes. I think these scopes are important to improve color grading capabilities in Shotcut.


ddennedy:
I think this is a good call now. Thanks for thinking of it! And the contribution of course.

PR #659, opened 2018-11-09
Image File Compression

Peppernrino:
Ran a series of lossless compression algorithms to make the images smaller. Total reduction of 84,673 bytes.

Example here: https://pastebin.com/MTtcVck2


ddennedy:
I am not very interested this because the saving is insignificant, and I maintain these from other sources where icons are added over time. I would be interested in a pull request that integrates with the build system or github.

Peppernrino:
Ok. :)

ddennedy:
I need a process that can be repeated, and I do not want it to to depend on a crapload of utilities.  Pick one or two and integrate it as a make target in either src/src.pro or new scripts/Makefile. Then, I can periodically run make to repeat the process when icons are changed or added. Thanks

ddennedy:
Superseded by #706

PR #667, opened 2018-11-12
Waveform improvements

bmatherly:
With these changes it is possible to make measured changes to video levels.

### DIFF
--- a/src/widgets/scopes/videowaveformscopewidget.cpp
+++ b/src/widgets/scopes/videowaveformscopewidget.cpp
@@ -18,7 +18,12 @@
 
 #include "videowaveformscopewidget.h"
 #include <Logger.h>
+#include <QMouseEvent>
 #include <QPainter>
+#include <QToolTip>
+
+const qreal IRE0 = 16;
+const qreal IRE100 = 235;
 
 VideoWaveformScopeWidget::VideoWaveformScopeWidget()
   : ScopeWidget("VideoZoom")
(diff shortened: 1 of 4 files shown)


src/widgets/scopes/videowaveformscopewidget.cpp
```diff
@@ -82,17 +87,65 @@ void VideoWaveformScopeWidget::paintEvent(QPaintEvent*)
...
+    int ire = (ire0y - event->pos().y()) / ireStep;
+
+    m_mutex.lock();
+    int frameWidth = m_displayImg.width();
+    m_mutex.unlock();
+
+    if(frameWidth != 0)
+    {
+        int pixel = frameWidth * event->pos().x() / width();
+        text =  QString("Pixel: %1\nIRE: %2").arg(QString::number(pixel), QString::number(ire));
```
ddennedy: Should this string be translatable?

src/widgets/scopes/videowaveformscopewidget.cpp
```diff
@@ -82,17 +87,65 @@ void VideoWaveformScopeWidget::paintEvent(QPaintEvent*)
...
+    }
+    else
+    {
+        text =  QString("IRE: %1").arg(QString::number(ire));
```
ddennedy: Should this string be translatable?

src/widgets/scopes/videowaveformscopewidget.cpp
bmatherly: https://github.com/mltframework/shotcut/pull/669

PR #672, opened 2018-11-23
Misc. typos

luzpaz:
Found via `codespell`


luzpaz:
Reverted `export-edl.js`
Good idea regarding the make target. :+1:

ddennedy:
I added the make target in 6bea957.

PR #676, opened 2018-12-06
Create shotcut.1

alexmyczko:
a very basic manual page to be installed in /usr/share/man/man1 and read with man shotcut
feel free to improve it. i'm using it with the debian/ubuntu package: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=880479

### DIFF
--- a/packaging/linux/shotcut.1
+++ b/packaging/linux/shotcut.1
@@ -1,0 +1,34 @@
+.TH SHOTCUT "1" "December 2018" "Shotcut 18.12.06" "User Commands"
+.SH NAME
+shotcut \- video editor
+.SH SYNOPSIS
+.B shotcut
+[\fI\,options\/\fR] \fI\,resource\/\fR
+.SH OPTIONS
+.TP
+\fB\-h\fR, \fB\-\-help\fR
+Displays this help.
+.TP
+\fB\-v\fR, \fB\-\-version\fR
+Displays version information.
+.TP
+\fB\-\-fullscreen\fR
+Fill the screen with the Shotcut window.
+.TP
+\fB\-\-noupgrade\fR
+Hide upgrade prompt and menu item.
+.TP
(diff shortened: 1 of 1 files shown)


packaging/linux/shotcut.1
```diff
@@ -0,0 +1,34 @@
+.TH SHOTCUT "1" "December 2018" "Shotcut 18.12.06" "User Commands"
+.SH NAME
+shotcut \- video editor
+.SH SYNOPSIS
+.B shotcut
+[\fI\,options\/\fR] \fI\,resource\/\fR
```
ddennedy: "resource" can be multiple files and folders and is optional. [GNU ls uses `[FILE]...`](https://linux.die.net/man/1/ls). I can improve the usage output from the executable for consistency.
```
--- src/main.cpp
+++ src/main.cpp
@@ -144,8 +144,8 @@ public:
             QCoreApplication::translate("main", "The directory for app configuration and data."),
             QCoreApplication::translate("main", "directory"));
         parser.addOption(appDataOption);
-        parser.addPositionalArgument("resource",
-            QCoreApplication::translate("main", "A file to open."));
+        parser.addPositionalArgument("[FILE]...",
+            QCoreApplication::translate("main", "Zero or more files or folders to open"));
         parser.process(arguments());
 #ifdef Q_OS_WIN
         isFullScreen = false;
```

alexmyczko:
Looks good, go ahead! (i'm using debian/manpages file for the manual page to get installed, after installation gzip -9 should be run on it)

ddennedy:
I am not going to add gzip upon install, and I am removing the version number from the man page because I did not want to maintain the version number for every highly build and monthly release. If you need changes for these things, you need to submit a patch. For example, you need to figure out how to turn the page into a template where the version variable gets substituted.

PR #690, opened 2019-01-12
Improve Color Grading filter UI.

bmatherly:
* Spinners go from -100 to + 100
* All spinners are at 0 by default
* Scrolling on the color wheel increments the spinners by 0.1


ddennedy:
Is this backwards compatible with existing projects? Does it still work with Movit?

PR #698, opened 2019-02-24
make it compile with Qt 5.12

eszlari:


ddennedy:
No verification of compatibility with existing non-flatpak builds means this must wait until after the next release.

ddennedy:
I confirmed this is working.

PR #699, opened 2019-02-24
Linux/Wayland: make window title bar icon work

eszlari:
Otherwise, only a generic icon is shown.

This must match the naming scheme of `packaging/linux/org.shotcut.Shotcut.desktop`.

Also `QGuiApplication::setDesktopFileName("org.shotcut.Shotcut");` needs to be added somewhere in `main`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -113,7 +113,7 @@ public:
 #endif
         addLibraryPath(dir.absolutePath());
         setOrganizationName("Meltytech");
-        setOrganizationDomain("meltytech.com");
+        setOrganizationDomain("shotcut.org");
         setApplicationName("Shotcut");
         setApplicationVersion(SHOTCUT_VERSION);
         setAttribute(Qt::AA_UseHighDpiPixmaps);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -113,8 +113,9 @@ public:
 #endif
         addLibraryPath(dir.absolutePath());
         setOrganizationName("Meltytech");
-#ifdef Q_OS_LINUX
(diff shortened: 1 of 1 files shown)


eszlari:
example: https://user-images.githubusercontent.com/15967631/46910388-5ad76e00-cf19-11e8-9c82-7267edc5e04e.png

ddennedy:
This breaks loading existing settings on macOS. Also, technically, the organization responsible for Shotcut is Meltytech, whose domain is meltytech.com and shotcut.org is the product website. You should consider changing the name of the desktop file. Does that mean the package name must change to com.meltytech.Shotcut? That would be confusing for users to change and remember. I am open to a compile define macro to change this for flatpak.

eszlari:
> You should consider changing the name of the desktop file. Does that mean the package name must change to com.meltytech.Shotcut?

Unfortunately, flatpak doesn't support renaming yet: https://github.com/flatpak/flatpak/issues/963

ddennedy:
OK, please add `setDesktopFileName("org.shotcut.Shotcut");` inside the `#ifdef Q_OS_LINUX`. Also, I think this change could apply to BSD in which case it should be `#if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)`

eszlari:
> please add `setDesktopFileName("org.shotcut.Shotcut");`

I'm no C++ expert, but will this work? Notice that, it's called `QGuiApplication::setDesktopFileName` and **not** `QCoreApplication::setDesktopFileName`!

ddennedy:
Yes, because Shotcut's Application inherits from QCoreApplication.

PR #703, opened 2019-02-26
Playlist

bmatherly:
Adds features for sorting by file date and modifying file date.

Depends on https://github.com/mltframework/mlt/pull/418

This change is not urgent and should probably wait until after the next release.

Comments welcome.

### DIFF
--- a/src/commands/playlistcommands.cpp
+++ b/src/commands/playlistcommands.cpp
@@ -164,4 +164,37 @@ void MoveCommand::undo()
     m_model.move(m_to, m_from);
 }
 
+SortCommand::SortCommand(PlaylistModel& model, int column, Qt::SortOrder order, QUndoCommand *parent)
+    : QUndoCommand(parent)
+    , m_model(model)
+    , m_column(column)
+    , m_order(order)
+{
+    m_xml = MLT.XML(m_model.playlist());
+    QString columnName = m_model.headerData(m_column, Qt::Horizontal, Qt::DisplayRole).toString();
+    setText(QObject::tr("Sort playlist by %1").arg(columnName));
+}
+
+void SortCommand::redo()
+{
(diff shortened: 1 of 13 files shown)


ddennedy:
There is a little risk with people not agreeing with the approach:
- modifying file system instead of embedded metadata
- not performed by clicking column heading
- not available in image properties dialog

However, I am going to accept it nevertheless; just sharing my anticipation/concern.

bmatherly:
> modifying file system instead of embedded metadata

I agree. It would be interesting to know if there is a normative/defacto method to associate asset times with media. I am aware of embedded timecodes - which are not supported in a standardized way by consumer devices. There is also the creation_date metadata provided by ffmpeg. I have found this to actually be more reliable than the file creation date when it is present. But it is not supported by all formats. I am open to changing the approach - it would be nice to get this right the first time and not regret it in the future.

My research was what led me to implement this:
https://github.com/mltframework/mlt/commit/53eaa8b3ffe0655f38863c7973a10c0d1db23a5f
We could generalize this logic in a mlt_producer function and make it generally available.
The functions would be:
* mlt_producer_get_date(): returns the creation date as determined by a cascading detection algorithm
* mlt_producer_set_date(): sets the creation date as specified by adding a property to the producer. This property is always the preferred date when mlt_produer_get_date() is called.

> not performed by clicking column heading

I actually implemented clicking column headings. But then I changed my approach. There were two problems:
1) I worry that people might think that clicking the column headings only changes the order in the view and not the order in the playlist itself. Clicking a command in a menu feels more "permanent".
2) I did not find a way in QT to only enable sorting on some columns. The "#", "in" and "start" headings do not make sense to sort by.

> not available in image properties dialog

Good idea. I will add this in a future commit. Let me know if you think of any others.

Issue #696, opened 2019-02-20
Compile error in util.cpp v19.01.27


TemporarilyOffline:
Error Message:

util.cpp: In static member function ‘static uint Util::versionStringToUInt(const QString&)’:
util.cpp:199:5: error: ‘const_reverse_iterator’ is not a member of ‘QList<QString>’
     QList<QString>::const_reverse_iterator i;
     ^
util.cpp:201:10: error: ‘i’ was not declared in this scope
     for (i = list.crbegin(); i != list.crend(); ++i, factor *= 1000)
          ^
util.cpp:201:19: error: ‘class QStringList’ has no member named ‘crbegin’
     for (i = list.crbegin(); i != list.crend(); ++i, factor *= 1000)
                   ^
util.cpp:201:40: error: ‘class QStringList’ has no member named ‘crend’
     for (i = list.crbegin(); i != list.crend(); ++i, factor *= 1000)
                                        ^
Makefile:2527: recipe for target 'util.o' failed
make[1]: *** [util.o] Error 1
make[1]: Leaving directory '/home/steve/Desktop/workspace/shotcut-19.01.27/src'
Makefile:69: recipe for target 'sub-src-make_first' failed
make: *** [sub-src-make_first] Error 2

OS:  Ubuntu 16.04 LTS

Shotcut Version: 19.01.27

$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/powerpc-linux-gnu/5/lto-wrapper
Target: powerpc-linux-gnu
[...]

PR #704, opened 2019-02-27
qmake: check Qt version

eszlari:
copied from:
https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/qtcreator.pri
https://code.qt.io/cgit/qt-creator/qt-creator.git/tree/qtcreator.pro

fixes:
* prevents issues like: https://github.com/mltframework/shotcut/issues/696
* `setDesktopFileName` needs Qt 5.7 (https://github.com/mltframework/shotcut/pull/699)

### DIFF
--- a/shotcut.pri
+++ b/shotcut.pri
@@ -1,0 +1,22 @@
+defineTest(minQtVersion) {
+    maj = $$1
+    min = $$2
+    patch = $$3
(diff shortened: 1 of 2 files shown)


ddennedy:
I want the minimum version to be 5.9.0.

PR #718, opened 2019-03-26
Use new producer datetime functions.

bmatherly:
Requires https://github.com/mltframework/mlt/pull/428

This change will use the datetime property on producers to sort items in the playlist.

### DIFF
--- a/src/dialogs/filedatedialog.cpp
+++ b/src/dialogs/filedatedialog.cpp
@@ -17,25 +17,30 @@
 
 #include "filedatedialog.h"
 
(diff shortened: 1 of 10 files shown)


ddennedy:
I just noticed that sort is not working if you include the same clip in the playlist more than once (same clip name). Please look into that.

bmatherly:
> I just noticed that sort is not working if you include the same clip in the playlist more than once (same clip name). Please look into that.

https://github.com/mltframework/shotcut/pull/726

PR #719, opened 2019-03-28
qmake: check for qmelt

eszlari:
Apparently, Arch Linux offers a Shotcut build without qmelt:

patch to rename qmelt:
https://git.archlinux.org/svntogit/community.git/tree/trunk/melt.patch?h=packages/shotcut

no webvfx in `depends`:
https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/shotcut

### DIFF
--- a/shotcut.pro
+++ b/shotcut.pro
@@ -5,6 +5,13 @@ include(shotcut.pri)
     error("Use at least Qt 5.9.0.")
 }
 
+QMELT_OUTPUT = $$system(qmelt -version, blob)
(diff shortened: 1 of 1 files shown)


ddennedy:
This breaks my builds because qmelt is not in the path on those machines, and I do not want to add a mock. I do not understand why they do not include WebVfx as they are depending on qt5-webkit. Maybe the maintainer does not know about it, but you would think qmelt would be a strong enough hint.

PR #774, opened 2019-09-23
Add gradient to lightshow, spectrum and waveform filters.

bmatherly:
This adds a new gradient control for the filters that support gradients. I would be interested in your opinion of the usability of the control. Maybe there is a different way to represent the gradient control that would be more intuitive.

### DIFF
--- a/src/qml/filters/lightshow/ui_lightshow.qml
+++ b/src/qml/filters/lightshow/ui_lightshow.qml
@@ -23,7 +23,7 @@ import Shotcut.Controls 1.0
 Item {
     property string rectProperty: "rect"
     property rect filterRect: filter.getRect(rectProperty)
-    property var defaultParameters: [rectProperty, 'color.1', 'color.2' ,'osc', 'frequency_low', 'frequency_high', 'threshold']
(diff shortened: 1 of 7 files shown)


ddennedy:
There is a bug when there are multiple filters using the GradientControl. They interfere with one another when switching filters in the UI. This is most obvious with differing number of stops, but it also seems the colors become corrupt. I found something similar recently in the Rectangle VUI control. I had to make the default value in the filter UI invalid and make the control ignore invalid. In the case of rectangle, the invalid state is width or height < 0.

Regarding usability, in general, I think the GradientControl is nice. But I worry about 2 things. First, it is not very friendly as a single color because people will not realize it is a gradient with stops, click the bar, and nothing happens. I think the single color case should not look like a bar with stops but be like today: a simple square. IOW, I think the user should opt into a gradient. Secondly, because the gradient is more advanced, perhaps the default should be a single color.

bmatherly:
Thanks your for taking the time to look at this. I believe I have resolved your comments. More comments are welcome. If there are no further comments, feel free to squash the commits.

PR #799, opened 2019-12-02
Zoom scope

bmatherly:
Proposal to add a new "Zoom" scope.

I also liked your name suggestion "Inspection" scope. Do you have a preference?

It is informative to open the color bar producer and zoom in with the widget. The chroma bleeding is very apparent.

I would also be interested in general feedback about usability.


src/glwidget.h
```diff
@@ -113,6 +113,7 @@ public slots:
     void offsetChanged();
     void imageReady();
     void snapToGridChanged();
+    void toggleZoom(bool);
```
bmatherly: I added this so that the scope can toggle the player zoom to "best fit". Maybe you can think of a better way?

src/mainwindow.cpp
```diff
@@ -224,6 +224,7 @@ MainWindow::MainWindow()
     connect(MLT.videoWidget(), SIGNAL(started()), SLOT(processMultipleFiles()));
     connect(MLT.videoWidget(), SIGNAL(paused()), m_player, SLOT(showPaused()));
     connect(MLT.videoWidget(), SIGNAL(playing()), m_player, SLOT(showPlaying()));
+    connect(MLT.videoWidget(), SIGNAL(toggleZoom(bool)), m_player, SLOT(toggleZoom(bool)));
```
bmatherly: Added to be able to toggle the player zoom from the scope.

ddennedy:
> The chroma bleeding is very apparent.

When not using GPU processing, MLT is instructed to render to yuv422. Then, that is converted to yuv420p into the shared frame to make the GL operations faster (less data transfer) and simpler. That probably contributes. Maybe it is not such a good idea to release this scope yet in light of that. What is your thought? Are we opening ourselves up to criticism since this operates at preview quality? Yes, I know most people will encode to yuv420p, but not always.

Other issues I noticed in quick test:

1. When I restart the waveform video scope opened instead of the zoom scope.
2. The color picker behaves different for the filters now. See White Balance, for example, it picks up strange colors. You can decide whether to try to fix it on top of your changes, or revert the color picker changes.

ddennedy:
Regarding the apparent chroma bleeding, I found that our existing zoom 1000% exposes this just as well only blurrier due to scaling interpolation. No one has complained or used it against us.

Hmm... I see I can reduce zoom to 2x. Next request will be 1x and scaled to fit. Then, add scrub bar and player controls and make this a dockable player. Lastly, make Source and Project players each their own panel. Give an inch... ;-)

bmatherly:
Thanks for testing.

> Maybe it is not such a good idea to release this scope yet in light of that. What is your thought?

I would like to see if it can easily be improved before I make an opinion. If rendering in RGB would help, I think that would be a minimal change since almost all filters use RGB anyway. I will do some experiments.

Regardless, I expect this scope will encourage extra scrutiny. We should welcome that (to some degree) while we also work to manage expectations.

> When I restart the waveform video scope opened instead of the zoom scope.

I have only experienced this after I switch to a different git branch and the previously selected scope is no longer available - then the system picks a different scope to show. Can you reproduce this when using the same code branch?

> The color picker behaves different for the filters now.

Fixed in https://github.com/mltframework/shotcut/pull/799/commits/271dae61c0387031822b636ef248d3a11f262eb9

>  I see I can reduce zoom to 2x

The zoom range is currently 2x-20x. I would be open to suggestions to that. I think that 5x could be the minimum. I also thought about making zoom steps like a microscope/telescope: 5x, 10x, 20x.

bmatherly:
> I would like to see if it can easily be improved before I make an opinion. If rendering in RGB would help, I think that would be a minimal change since almost all filters use RGB anyway. I will do some experiments.

https://github.com/mltframework/shotcut/pull/799/commits/bdae20f4e27d9f4404327137a04e2d6c4adcb083 is a proposed change that completely eliminates the chroma bleeding. It renders in RGBA24 instead of yuv422.

The new image caching feature in shared frame is going to be very useful. If the image is ever converted for any reason, the converted image is available to anything else that wants to use it. In this case, the frame is rendered to RGB and converted to YUV. This sounds wasteful, but some of the scopes need both, and they would not have to make the conversion in the scope thread because of the cache.

ddennedy:
> bdae20f is a proposed change that completely eliminates the chroma bleeding. It renders in RGBA24 instead of yuv422.

Hey, I am glad you figured this out! I am OK to render to RGBA for the most. The only drawback I see is that previewing clips or using Shotcut as a simple player are not as fast due to RGB conversion and back. However, those are no longer primary use cases by far.

> When I restart the waveform video scope opened instead of the zoom scope.

This mysteriously went away. Must have been something in saved Qt window state. Also, the color picker works as expected. Merging

bmatherly:
Thanks Dan!

> The only drawback I see is that previewing clips or using Shotcut as a simple player are not as fast due to RGB conversion and back.

I had also thought about the idea of letting advanced users choose the preview render format. They could choose between:
* 4:2:0 (yuv420p)
* 4:2:2 (yuv422)
* 4:4:4 (rgba24)

I would be willing to implement something like this.

Advantages:
* The user would be able to preview in the format they plan to export in

Disadvantages:
* It could be confusing for novice users
* I don't like to clutter the UI with many options
* The result would really only be noticeable in the zoom scope and maybe the player when zoomed way in

ddennedy:
yuv420p is not a valid direct render target in MLT. I do not want to offer that now as things need to change and expand with respect to processing and pixel format.

PR #811, opened 2019-12-16
Scope optimization

bmatherly:
Adds some optimizations to this PR: https://github.com/mltframework/shotcut/pull/810


ddennedy:
Thanks but this is going in after the 19.12 release.

PR #824, opened 2020-01-25
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds](https://reproducible-builds.org/) effort we noticed that `shotcut` could not be built reproducibly.

This is because it uses the current build date for the `SHOTCUT_VERSION` variable. This PR bases this value on [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) instead if available.

This was originally filed in Debian as [#949817](https://bugs.debian.org/949817).

### DIFF
--- a/src/src.pro
+++ b/src/src.pro
@@ -325,7 +325,7 @@ debug_and_release {
 LIBS += -lCuteLogger
 
 isEmpty(SHOTCUT_VERSION) {
-    !win32:SHOTCUT_VERSION = $$system(date "+%y.%m.%d")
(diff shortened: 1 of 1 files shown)


lamby:
@ddennedy The major disadvantage of using `SHOTCUT_VERSION` like that is that it requires that all distributions (etc.) have to independently learn and discover that this is the way to ensure a reproducible build and to modify their buildsystems accordingly. Sure, this is not too arduous but if we can do centrally using a well-known pattern (ie. `SOURCE_DATE_EPOCH` then that is surely superior).

Would you thus be amenable to a patch that worked on MacOS (and BSD…) `date` too?

ddennedy:
Yes please amend to support BSD date as well.

lamby:
@ddennedy Updated. :)

PR #825, opened 2020-01-26
Add new pitch filter

bmatherly:


scripts/build-shotcut.sh
```diff
@@ -860,6 +875,12 @@ function set_globals {
   fi
   CFLAGS_[17]="$CFLAGS"
   LDFLAGS_[17]=$LDFLAGS
+
+  #####
+  # rubberband
+  CONFIG[18]="./configure --prefix=$FINAL_INSTALL_DIR --disable-programs --disable-vamp --disable-ladspa"
```
ddennedy: To fix Windows build I added
```
  if test "$TARGET_OS" = "Win32" -o "$TARGET_OS" = "Win64" ; then
    CONFIG[18]="${CONFIG[18]} --host=$HOST"
  fi
```

src/qml/filters/audio_pitch/meta.qml
```diff
@@ -0,0 +1,26 @@
...
+        allowAnimateOut: false
+        simpleProperties: ['octaveshift']
+        parameters: [
+            Parameter {
+                name: qsTr('Pitch')
+                property: 'octaveshift'
+                isSimple: true
+                isCurve: true
+                minimum: -1.0
+                maximum: 1.0
```
ddennedy: This range is much smaller than `filter_rbpitch.yml`. What do you think about expanding it? -2 - 2 works with 4x - 0.25x speed without sounding too horrible IMO. 2x - 0.5x seems quite limited.

src/qml/filters/audio_pitch/ui.qml
```diff
@@ -0,0 +1,201 @@
...
+                    filter.resetProperty('octaveshift')
+                    filter.set('octaveshift', octaveSlider.value)
+                }
+            }
+        }
+
+        Label {
+            text: qsTr('Frequency Scale')
+            Layout.alignment: Qt.AlignRight
+            ToolTip { text: qsTr('Specify the pitch shift as a frequency scale. 0.5 halves the frequency (down one octave). 2 doubles the frequency (up one octave). 1 is unchanged.') }
```
ddennedy: Is there a way to express this as the inverse so one can use the same value from Speed here?

scripts/build-shotcut.sh
bmatherly: For the link errors, do we link statically on Windows? If so, does it work to add this to rubberband.pc:

```
Libs.private: -lfftw3 -lsamplerate
```

If that works, I can submit a patch.

src/qml/filters/audio_pitch/meta.qml
bmatherly: Yes. I am open to that. I admit that I was conservative. It is easier to give more feature than to take away. Will make an updated patch shortly.

src/qml/filters/audio_pitch/ui.qml
bmatherly: That is an interesting idea that I did not think of. My plan was to add a "pitch compensation" button to the avformat producer widget to handle the case of pitch correction for speed.

I am very open to discussion about the representation of the ratio. Options:
1) Octaves (currently implemented)
2) Frequency scale (currently implemented)
3) Semitones
4) Speed compensation ratio
5) maybe something else in musical notation?

scripts/build-shotcut.sh
ddennedy: I have fixed it locally. I will patch it when merging. On Windows, we have to use `-lfftw3-3` which I think is unusual and should not be sent upstream. My patch is like `sed 's/-lrubberband/-lrubberband -lfftw3-3 -lsamplerate/' -i rubberband.pc.in`. I would have to run another test using Libs.private.

bmatherly: Ok. Understood. If you do come up with any patches that can be sent upstream, let me know and I would be happy to submit them.

src/qml/filters/audio_pitch/ui.qml
ddennedy: I figured you would do that, but I plan to make beta release this Sunday. I expect to include this PR, but I do not know if you can get Speed integration done. If not, a simple way to simply copy the value from Speed to this filter would be ideal since I think 90% of the uses will be for that. Otherwise, I do not fell more representations are needed at this time.

scripts/build-shotcut.sh
bmatherly: I noticed this:
https://github.com/breakfastquay/rubberband/blob/master/configure#L1344
LIBS        libraries to pass to the linker, e.g. -l<library>
FFTW_LIBS   linker flags for FFTW, overriding pkg-config

So maybe you can add: 
FFTW_LIBS=-lfftw3-3
or
LIBS="-lfftw3-3 -lsamplerate"

Or, if you are happy with the sed work-around, maybe it is fine.

src/qml/filters/audio_pitch/ui.qml
bmatherly: Speed compensation by Sunday would be tight...

Following on your suggestion... How about if I replace "Frequency Scale" with "Speed Compensation" and it will have a range 0.25 - 4.0. But it will apply the inverse of that value to match the corresponding speed. That would be easy to implement.

scripts/build-shotcut.sh
ddennedy: To clarify, rubberband builds, but it is the MLT module failing to link. `Libs.private: -lfftw3-3 -lsamplerate` does not work because `pkg-config --libs rubberband` does not emit them, and we are linking statically to rubberband because for some reason it is build .so files under under cross-compile in addition to the static link library. So, the MLT builds only accepts the static lib. I am not particularly to keen to debug that build issue in their project.

` FFTW_LIBS` is used to override what is in fftw3.pc, but that is correct in our win32 prebuild bundle. I doubt that will help the MLT build. I will stick to this sed hack for now.

src/qml/filters/audio_pitch/ui.qml
ddennedy: Yes, please

bmatherly: Done. See the latest. Also doubled the range.

The idea works well. Type in a speed change in the properties panel, add a pitch filter, type the same number into the pitch compensation field and the audio sounds normal.

PR #832, opened 2020-03-04
Update shotcut_de.ts

Joshi234:
Some small changes to german translation


ddennedy:
Thank you for your willingness to help, but I only accept translation updates through transifex.com where there are multiple collaborators in the German team.

https://www.transifex.com/ddennedy/shotcut/

PR #838, opened 2020-03-16
Fix typo in French translation

wjt:
A space was missing between two words.


ddennedy:
We only accept translation updates from transifex.com

PR #841, opened 2020-03-25
Fix behavior of previous/next keyframe button

yuyichao:
Currently when seeking the previous/next keyframe button, no matter which button
is pressed, it is always the previous/next keyframe corresponding to
the currently selected parameter that is selected and the current track is changed
to the track corresponding to the button only after that, i.e. for two tracks

--o1----o2----|----o3-----
--o4---------------o5-----

where `o*` is a keyframe and `|` is the cursor and the current track, pressing previous keyframe
button on the second parameter will jump to `o2` instead of `o4`.
I don't think this is the correct/expected behavior especially since `o2` isn't even a
keyframe for the second parameter at all.

This commit changed the behavior to change the current track first before selecting the keyframe.

This behavior was first introduced in 417a2f8c06907d658d3fc12a115eb097300eeb27.


ddennedy:
Thank you, you are correct about expected behavior.

PR #855, opened 2020-04-14
Linux: add 128x128 icon

eszlari:
Flathub now expects a 128x128 icon:
https://flathub.org/builds/#/builders/19/builds/2157


eszlari:
The 128x128 is made from `packaging/macos/shotcut.icns`.

PR #857, opened 2020-04-18
add External Monitor "NDI Output" feature

walterav1984:
This feature will make the Shotcut "External Monitor" output (pause/playback/scrubing) available as a NewTek NDI video&audio stream using the mlt/melt consumer. Other NDI capable programs like "OBS Studio" or "CasparCG server" may than be able to use "hostname (Shotcut Output)" as a valid NDI source.

This requires building mlt/melt with the NewTek NDI sdk "include" files and NDI "lib" files before building Shotcut. Tested using Shotcut 20.04.18 on ubuntu-mate 18.4.4 LTS amd64 with libndi 4.5 in obs-studio(ppa) 25.04 with obs-ndi from Palakis see https://github.com/Palakis/obs-ndi/releases.

Shotcut forum topic:
https://forum.shotcut.org/t/please-could-you-add-newtek-ndi-support/5209/5

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -601,6 +601,14 @@ void MainWindow::setupSettingsMenu()
         action->setData(QString("sdi"));
         m_externalGroup->addAction(action);
     }
+    
+    Mlt::Consumer ndi(MLT.profile(), "ndi");
+    if (ndi.is_valid()) {
+        QAction* action = new QAction("NDI Output", this);
+        action->setCheckable(true);
+        action->setData(QString("ndi:Shotcut Output"));
+        m_externalGroup->addAction(action);
+    }
 #endif
 
     Mlt::Profile profile;


ddennedy:
Thanks, but the hard part is to add build support for all platforms and test and debug it across all 3 supported OS. This will have to wait until I decide to get around to doing that.

walterav1984:
This PR was indeed very simple and only tested on Ubuntu Linux but if building, testing and debuging would be finished is shipping NDI support with Shotcut packages allowed license wise or with a seperate installer for the libs like Palakis does for OBS Studio for 3 platforms?

ddennedy:
I am worried a little about the license of linking a GPL Shotcut against the NDI SDK (indirectly, still a MLT module, but delivered as a GPLv3 bundle). However, the MLT module is LGPL and can be made a separate add-on. I am not worried about adding the UI  support in Shotcut that only makes itself appear when the MLT module is available.

walterav1984:
> worried a little about the license of linking
A license issue/violation has happened the otherway around (NewTek > FFmpeg), maybe this thread from the past might be insightfull of what is acceptable:
https://trac.ffmpeg.org/ticket/7589  

> not worried about adding the UI support in Shotcut that only makes itself appear when the MLT module is available

It's a good idea to make the module only a visible option if installed and or when NDI libraries get detected just like the Decklink Output only popups when the Desktop Video Driver/hardware is installed.

However I chose the "DVEO VidPort" as example code so it always shows up for this NDI example, but should the "DVEO VidPort" only get listed if the device/drivers are installed on the system?

ddennedy:
> However I chose the "DVEO VidPort" as example code so it always shows up for this NDI example, but should the "DVEO VidPort" only get listed if the device/drivers are installed on the system?

This "always shows" prevents me from being able to merge this. Otherwise, I would merge even if you would need to build the MLT NDI module manually. As it stands, I am not in a position to respond to thousands of people asking for it (Shotcut has a big audience). The DVEO VidPort is Linux only - a much smaller audience. A MLT service's constructor should try to fail if a runtime dependency is not available. In the case of NDI, it seems this is possible but not done in `consumer_ndi_init()`. In the case of `sdi`, it can depend on `dev_video` and `dev_audio` properties, which are not available at the time of construction in `consumer_SDIstream_init()`. Typically, we would expect additional information needed for construction in the `arg` argument. However, multiple arguments makes this more complicated. Also, [DVEO's drivers](https://dveo.com/drivers.shtml) are no longer maintained and only compatible with Linux kernel up to version 3.x. So, I am going to drop support for DVEO in Shotcut and deprecate it in MLT.

walterav1984:
> This "always shows" prevents me from being able to merge this.

It was a nice oppertunity for me or others to use it under linux, but I'm far from a skilled coder to make this work similar as in available/hiding in the UI like the Decklink external monitor option.

> As it stands, I am not in a position to respond to thousands of people asking for it (Shotcut has a big audience).

I fully understand since due to popularity Palakis his GitHub obs-ndi "issues" page also starts to rise. 

The DVEOVidPort already got removed from master yesterday, should I close this PR and maybe continue in the forum with ubuntu focussed build instructions?

ddennedy:
Leave this open for now.

gnafuthegreat:
I just happened by this while looking at Shotcut, and I thought it might be worth mentioning libNDI:

https://code.videolan.org/jbk/libndi

I'm not sure if that would help in this situation, but it at least should answer the license concerns in the situations it does cover.

walterav1984:
Hi @gnafuthegreat libndi is indeed a interesting project but its focus currently is on compatibility for decode/playback. However for this shotcut PR encode of NDI streams is required.

gnafuthegreat:
> Hi @gnafuthegreat libndi is indeed a interesting project but its focus currently is on compatibility for decode/playback. However for this shotcut PR encode of NDI streams is required.

Aah, good point.

PR #866, opened 2020-05-12
Add QML interface for avfilter.vaguedenoiser

austincbrooks:


src/qml/filters/vaguedenoiser/meta.qml
```diff
@@ -0,0 +1,34 @@
...
+
+Metadata {
+    type: Metadata.Filter
+    name: qsTr('Reduce Noise: Wavelet')
+    objectName: 'vaguedenoiser'
+    mlt_service: 'avfilter.vaguedenoiser'
+    qml: 'ui.qml'
+
+    keyframes {
+        allowTrim: true
```
ddennedy, line 32: This is the default value, so you can remove `keyframes` block if you want.

PR #867, opened 2020-05-18
Slideshow

bmatherly:
Initial beta implementation of a slideshow generator.

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -15,14 +15,22 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "dialogs/longuitask.h"
 #include "timelinecommands.h"
 #include "mltcontroller.h"
 #include "shotcut_mlt_properties.h"
 #include "settings.h"
 #include <Logger.h>
+
+
 #include <QMetaObject>
 
 namespace Timeline {
+
+Mlt::Producer* deserializeProducer(QString& xml)
+{
+    return new Mlt::Producer(MLT.profile(), "xml-string", xml.toUtf8().constData());
+}
 
 AppendCommand::AppendCommand(MultitrackModel &model, int trackIndex, const QString &xml, QUndoCommand *parent)
     : QUndoCommand(parent)
@@ -34,24 +42,27 @@ AppendCommand::AppendCommand(MultitrackModel &model, int trackIndex, const QStri
     setText(QObject::tr("Append to track"));
 }
 
-
 void AppendCommand::redo()
 {
     LOG_DEBUG() << "trackIndex" << m_trackIndex;
+    LongUiTask longTask(QObject::tr("Append to Timeline"));
     m_undoHelper.recordBeforeState();
-    Mlt::Producer clip(MLT.profile(), "xml-string", m_xml.toUtf8().constData());
-    if (clip.type() == playlist_type) {
-        Mlt::Playlist playlist(clip);
+    Mlt::Producer* producer = longTask.runAsync<Mlt::Producer*>(QObject::tr("Preparing"), deserializeProducer, m_xml);
+    if (producer->type() == playlist_type) {
+        Mlt::Playlist playlist(*producer);
         int count = playlist.count();
         for (int i = 0; i < count; i++) {
+            longTask.reportProgress(QObject::tr("Appending"), i, count);
             QScopedPointer<Mlt::ClipInfo> info(playlist.clip_info(i));
-            clip = Mlt::Producer(info->producer);
+            Mlt::Producer clip = Mlt::Producer(info->producer);
             clip.set_in_and_out(info->frame_in, info->frame_out);
             m_model.appendClip(m_trackIndex, clip);
         }
     } else {
-        m_model.appendClip(m_trackIndex, clip);
+        m_model.appendClip(m_trackIndex, *producer);
     }
+    longTask.reportProgress(QObject::tr("Finishing"), 0, 0);
+    delete producer;
     m_undoHelper.recordAfterState();
 }
 
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -39,8 +39,19 @@ void AppendCommand::redo()
 {
     LOG_DEBUG() << "trackIndex" << m_trackIndex;
     m_undoHelper.recordBeforeState();
-    Mlt::Producer producer(MLT.profile(), "xml-string", m_xml.toUtf8().constData());
-    m_model.appendClip(m_trackIndex, producer);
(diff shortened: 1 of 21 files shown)


ddennedy:
It passes a quick functional test. A code review is pending. A few things I noticed:

1. I think the only big problem I see there is that there is no undo item for the timeline operation.
2. I am curious: Did you utilize preview scaling for the preview?
3. The [UI convention](https://shotcut.com/notes/ui-conventions/) is not to capitalize every word in field labels (different than menu items). I notice you have not been following that in your filters, but I have not yet decided to break translations for it. Please change these labels; for example, "Clip duration" instead of "Clip Duration".
4. I think the dialog Apply button should be changed to OK. Apply is typically used when you can make some things take affect without closing. And change Close to Cancel. This makes it more consistent with our dialog usage. I almost always use simply OK and Cancel.
5. Should we add Add Selected to Slideshow to the context menu?

bmatherly:
> 1. I think the only big problem I see there is that there is no undo item for the timeline operation.

Done. Also added for "Add Selected to Timeline".

> 2. I am curious: Did you utilize preview scaling for the preview?

I think that happens automatically if I request a smaller size image from the producer?
https://github.com/mltframework/shotcut/pull/867/files#diff-2ea532fe60e8a146c4ae9fe60c613575R162
Or do I have to do something else to trigger scaling?

> 3. The [UI convention](https://shotcut.com/notes/ui-conventions/) is not to capitalize every word in field labels (different than menu items). I notice you have not been following that in your filters, but I have not yet decided to break translations for it. Please change these labels; for example, "Clip duration" instead of "Clip Duration".

Done

> 4. I think the dialog Apply button should be changed to OK. Apply is typically used when you can make some things take affect without closing. And change Close to Cancel. This makes it more consistent with our dialog usage. I almost always use simply OK and Cancel.

Done

> 5. Should we add Add Selected to Slideshow to the context menu?

I would be inclined to exclude it (but I do not feel strongly). The context menu seems to be reserved for actions that apply to a single clip. The Hamburger menu seems to have actions that apply to the whole playlist. If we do add it to the context menu, we should also add "Add selected to Timeline".

ddennedy:
> Also added for "Add Selected to Timeline".

THANK YOU! I noticed that it was missing after I wrote this. :-\

> I think that happens automatically if I request a smaller size image from the producer?

Yes, I just had not looked at the code at that time. The code looks good to me without getting in too deep. I am ready to merge unless you have more to do.

> The context menu seems to be reserved for actions that apply to a single clip.

I noticed this as well when reviewing after my reply. This is inconsistent with Timeline clip's context menu. I think this should also include any actions that can operate with whatever is selected but it also needs some work perhaps to differentiate actions for single item versus multiple. As it stands, many of context menu actions are ambiguous with multiple selected. We will leave that as a separate change.

bmatherly:
> I am ready to merge unless you have more to do.

I have more work to do:
* Display a progress bar for long operations (I have a torture test with 400 clips that takes a minute to deserialize)
* Implement the blurred padding for aspect ratio conversion

But I think it has reached a useful point and it would be a good time to merge the progress thus far.

bmatherly:
I am pretty happy with the progress bar. Have a look and see what you think.

I notice that when I add a lot of files to the slideshow, it consumes a lot of memory. The memory consumption occurs while the slideshow playlist is being appended to the timeline. This process uses the same function as the "Add Selected to Timeline". But adding the playlist directly to the timeline does not seem to consume as much memory as adding the slideshow to the timeline. Maybe the filters and transitions cause much higher memory consumption.

If you have time, it would be good to try a test with adding 100 images to a slideshow on your system to see how it holds up.

I think that what is proposed here is suitable for code review or beta release.

ddennedy:
I made a test with 87 images, and the memory usage was about what I expected (much less with previewing scaling on). The affine filter/transition requests images at source resolution when preview scaling is not used. Then, the qimage producer caches that image, but uses mlt_cache LRU to limit it. I did notice that when I closed this project, the memory was not released. This is probably due to the memory pool. I noticed that when I repeatedly opened a small project with 7 clips on the timeline, and played it through each time, it does not use much more memory (+0.1GB after 20 times).

The progress bar is nice, but it can take a little while to appear - about the same amount of time as it takes to generate the preview. This can be a nice thing to apply to some other tasks in Shotcut at a later time.

One thing I noticed is that it is difficult to adjust the center crop when also using transitions because the start and keyframes are within the transition. So, when you try to adjust them you cannot see that part of the video. If it instead had used the Crop: Source filter with the Center option, I could simply go in there and adjust the bias. However, I realize this is a big change at this point and not yet asking for that.

ddennedy:
I have a commit to add to master that I tested in your branch. It calls `mlt_pool_purge()` whenever the project is closed either explicitly or implicitly by opening a new project. That helps considerably to reclaim memory. Although, in my test case:

0. start Shotcut
0. open just one image to initialize many things
0. note baseline memory level (330.8 MiB)
0. add many photos to playlist
0. play it through in project player
0. note memory level (558.7 MiB)
1. make a slideshow of all images
1. play it through
0. note memory level (1.8 GiB)
5. close
6. compare memory usage (828 MiB)

It reclaimed a lot of memory - much better than before - but not close to baseline. There are plenty of other scenarios to show where memory seems to be unreleased. I am confident purging the pool between projects is sensible, but I wonder if it should also be done at the time of Export File.

bmatherly:
> (+0.1GB after 20 times).

This is consistent with my testing.

> The progress bar is nice, but it can take a little while to appear - about the same amount of time as it takes to generate the preview.

Good point. That is the time it takes to generate the slideshow playlist. I added a commit to show a progress bar during that step of the process.

> This can be a nice thing to apply to some other tasks in Shotcut at a later time.

I think so. There are two kinds of long operations:
* Background operations - can process in a thread in the background and the user can continue to work while the operation occurs (e.g. waveform generation)
* Foreground operations - operations that need to make modifications to the UI or the underlying models. The user should not interact with the application during a foreground operation.

I think we can reapply the progress bar for any kind of foreground operation

> One thing I noticed is that it is difficult to adjust the center crop when also using transitions because the start and keyframes are within the transition. So, when you try to adjust them you cannot see that part of the video. If it instead had used the Crop: Source filter with the Center option, I could simply go in there and adjust the bias. However,

I noticed the same problem. To work around this, I temporarily changed the transition type to "cut" so that I can see the entire clip in preview. Then, after adjusting the affine filter, I set the transition back. This is not clean. But maybe there could be some way to "disable" the transition momentarily during preview and then re-enable it? I would be interested in discussion on this.

>  I am confident purging the pool between projects is sensible, but I wonder if it should also be done at the time of Export File.

This sounds promising.

I would imagine that there are probably also some memory leaks lurking in the code that I have written over the years. I have never been very proactive in analyzing for memory leaks in my code. I just "try to remember" to delete what I allocate.

PR #877, opened 2020-06-15
Add position tooltip to scrubbar.

bmatherly:
Seems useful. But maybe clutters up the screen too much?

### DIFF
--- a/src/scrubbar.cpp
+++ b/src/scrubbar.cpp
@@ -17,6 +17,8 @@
 
 #include "scrubbar.h"
 #include "mltcontroller.h"
+
+#include <QToolTip>
 #include <QtWidgets>
 
 static const int selectionSize = 14; /// the height of the top bar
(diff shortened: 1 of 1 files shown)


ddennedy:
I like it with only the position time and not a duration as that info is already displayed.
```
--- src/scrubbar.cpp
+++ src/scrubbar.cpp
@@ -164,5 +164,5 @@ void ScrubBar::mouseMoveEvent(QMouseEvent * event)
     }
     else if (event->buttons() == Qt::NoButton && MLT.producer()) {
-        QString text = QString(MLT.producer()->frames_to_time(pos)) + QString(" / ") + QString(MLT.producer()->frames_to_time(m_max));
+        QString text = QString::fromLatin1(MLT.producer()->frames_to_time(pos));
         QToolTip::showText(event->globalPos(), text);
     }
```

bmatherly:
Agreed. Done.

ddennedy:
This was merged in c848fbf7fed1cd9a626d545109e46438ad897d6a. I also added similar to timeline and keyframes.

PR #887, opened 2020-07-11
Update org.shotcut.Shotcut.desktop

henrikp:
XDG standards tell that comment fields in the desktop files shouldn't be redundant. My Russian and French aren't good enough to replace those fields with something else.

Signed-off-by: Henrik Pihl <ahvenas@gmail.com>


juippis:
Here's the output from `desktop-file-validate` that's part of https://freedesktop.org/wiki/Software/desktop-file-utils
```
$ desktop-file-validate org.shotcut.Shotcut.desktop 
org.shotcut.Shotcut.desktop: warning: value "Video Editor" for key "Comment" in group "Desktop Entry" looks the same as that of key "GenericName"
org.shotcut.Shotcut.desktop: warning: value "Logiciel de montage vidéo" for key "Comment[fr]" in group "Desktop Entry" looks the same as that of key "GenericName[fr]"
org.shotcut.Shotcut.desktop: warning: value "Видеоредактор" for key "Comment[ru]" in group "Desktop Entry" looks the same as that of key "GenericName[ru]"
```

PR #894, opened 2020-07-28
Pillar echo

bmatherly:
Requires https://github.com/mltframework/mlt/pull/598

### DIFF
--- a/src/widgets/slideshowgeneratorwidget.cpp
+++ b/src/widgets/slideshowgeneratorwidget.cpp
@@ -38,9 +38,10 @@
 
 enum
 {
-    ASPECT_CONVERSION_PAD_BLACK = 0,
-    ASPECT_CONVERSION_CROP_CENTER = 1,
-    ASPECT_CONVERSION_CROP_PAN = 2,
+    ASPECT_CONVERSION_PAD_BLACK,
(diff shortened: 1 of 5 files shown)


src/widgets/slideshowgeneratorwidget.cpp
```diff
@@ -66,6 +67,7 @@ SlideshowGeneratorWidget::SlideshowGeneratorWidget(Mlt::Playlist* clips, QWidget
     grid->addWidget(new QLabel(tr("Aspect ratio conversion")), 1, 0, Qt::AlignRight);
     m_aspectConversionCombo = new QComboBox();
     m_aspectConversionCombo->addItem(tr("Pad Black"));
+    m_aspectConversionCombo->addItem(tr("Pad Blur"));
```
ddennedy, line 70: I want to be a little respectful to Linux distro packagers that will have not yet upgraded MLT (as it may not be released again soon) and hide this option if the `pillar_echo` filter is not available.

bmatherly: > I want to be a little respectful to Linux distro packagers that will have not yet upgraded MLT (as it may not be released again soon) and hide this option if the `pillar_echo` filter is not available.

Done

PR #937, opened 2020-11-02
Update shotcut_de.ts

DJE98:
Changed Synchronizität (kind of dream) to Synchronität (at the same time) in german translation


ddennedy:
I can only accept translation updates through Transifex as mentioned on our [How To  Contribute](https://www.shotcut.org/howtos/contribute/) page. Otherwise, your updates will be overwritten by the updates from transifex.

DJE98:
Screenshot_2020-11-03 Shotcut Transifex Editor

Thank you for your help. I added the correction as suggestion.

PR #967, opened 2020-11-29
WIP: add cmake build system

eszlari:
only tested on Linux, no install target, build works from the tree

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,41 @@
+cmake_minimum_required(VERSION 3.13)
+
+string(TIMESTAMP SHOTCUT_VERSION "%y.%m.%d")
+
+project(shotcut
+  VERSION ${SHOTCUT_VERSION}
+  DESCRIPTION "cross-platform (Qt), open-source (GPLv3) video editor"
+  HOMEPAGE_URL "https://www.shotcut.org"
+  LANGUAGES CXX
+)
+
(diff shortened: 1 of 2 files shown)


ddennedy:
Why?

eszlari:
> Why?

* dependency discovery
* same build system for both mlt & shotcut
* faster (https://ninja-build.org)
* better IDE/editor support
* can generate `compile_commands.json` (https://clangd.llvm.org)
* pre-compiled headers

ddennedy:
Thanks, and I agree these are nice to have, but I have a ton to deal with. I do not want to personally deal with a new build system in the near term. Qt 6 will be the main driver, but there is a lot of work to even get close to trying it. In order to merge this, it will need:
- updated build scripts
- work on all supported OS
- When I open the project in Qt Creator I also want to be able to see the OTHER_FILES in the Projects tree.

eszlari:
>     * work on all supported OS

Linux & Windows: OK
macOS: I don't have any Apple devices.

eszlari:
btw: What would be the TODO list to switch MLT to completely to cmake (remove the old gmake system)?

brunetton:
Hi @eszlari sorry to ask but: how to compile using cmake ? (after cloning your fork repo and checkouting `cmake` branch, of course). Thanks

brunetton:
🎉🎉🎉

PR #978, opened 2020-12-13
Hdr convert

bmatherly:
Tested this with the life of pie clip and it worked very well

image

### DIFF
--- a/src/dialogs/transcodedialog.ui
+++ b/src/dialogs/transcodedialog.ui
@@ -9,8 +9,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>367</width>
-    <height>311</height>
+    <width>418</width>
+    <height>340</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -142,20 +142,10 @@
      <layout class="QVBoxLayout" name="verticalLayout_2">
       <item>
        <layout class="QGridLayout" name="gridLayout" columnstretch="0,1">
-        <item row="1" column="1">
-         <widget class="QCheckBox" name="fpsCheckBox">
-          <property name="toolTip">
-           <string>Enable frame rate override.</string>
-          </property>
+        <item row="3" column="0">
+         <widget class="QLabel" name="frcLabel">
           <property name="text">
-           <string>Override frame rate</string>
-          </property>
-         </widget>
-        </item>
-        <item row="0" column="0" colspan="2">
-         <widget class="Line" name="line">
-          <property name="orientation">
-           <enum>Qt::Horizontal</enum>
+           <string>Frame rate conversion</string>
           </property>
          </widget>
         </item>
@@ -166,6 +156,34 @@
           </property>
          </widget>
         </item>
+        <item row="0" column="0" colspan="2">
+         <widget class="Line" name="line">
+          <property name="orientation">
+           <enum>Qt::Horizontal</enum>
+          </property>
+         </widget>
+        </item>
+        <item row="3" column="1">
+         <widget class="QComboBox" name="frcComboBox">
+          <property name="toolTip">
+           <string>Frame rate conversion method
+
+Duplicate: Duplicate frames.
+Blend: Blend frames.
+Motion Compensation: Interpolate new frames using motion compensation. This method is very slow and may result in artifacts.</string>
+          </property>
+         </widget>
+        </item>
+        <item row="1" column="1">
+         <widget class="QCheckBox" name="fpsCheckBox">
+          <property name="toolTip">
+           <string>Enable frame rate override.</string>
+          </property>
+          <property name="text">
+           <string>Override frame rate</string>
+          </property>
+         </widget>
+        </item>
         <item row="2" column="0">
          <widget class="QLabel" name="fpsLabel">
           <property name="text">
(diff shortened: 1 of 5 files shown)


src/widgets/avformatproducerwidget.ui
```diff
@@ -841,7 +867,7 @@
        </property>
        <property name="icon">
         <iconset theme="show-menu" resource="../../icons/resources.qrc">
-         <normaloff>:/icons/oxygen/32x32/actions/show-menu.png</normaloff>:/icons/oxygen/32x32/actions/show-menufill.png</iconset>
+         <normaloff>:/icons/oxygen/32x32/actions/show-menu.png</normaloff>:/icons/oxygen/32x32/actions/show-menufull.png</iconset>
```
ddennedy, line 870: This is a weird change, but as far as I can tell from looking at other UI XML files, there should be neither a "fill" or a "full" here.

src/dialogs/transcodedialog.ui
```diff
@@ -159,31 +163,34 @@
           </property>
          </widget>
         </item>
-        <item row="2" column="1">
-         <widget class="FrameRateWidget" name="fpsWidget" native="true">
+        <item row="3" column="1">
+         <widget class="QComboBox" name="frcComboBox">
           <property name="toolTip">
-           <string>Override the frame rate to a specific value.</string>
+           <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Frame Rate Conversion Method&lt;/span&gt;&lt;/p&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Duplicate: &lt;/span&gt;Duplicate frames&lt;/p&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Blend: &lt;/span&gt;Blend frames&lt;/p&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Motion Compensation: &lt;/span&gt;Interpolate new frames using motion compensation. This method is very slow and may result in artifacts.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
ddennedy, line 169: I do not like this rich text tooltip because it makes translation difficult or easy to break. Translation UIs often do not provide a rich text editor, and they can easily break the HTML or simply skip over it as "too much work; maybe sometime later." Maybe it is better to keep the tooltip simple and have more documentation in the forum docs category.

src/widgets/avformatproducerwidget.ui
bmatherly: Yes. Odd. Qt Creator does strange things sometimes. I do not like that it doesn't keep gridview items in order in the XML.

Fixed.

src/dialogs/transcodedialog.ui
bmatherly: Done

src/widgets/avformatproducerwidget.cpp
```diff
@@ -369,6 +369,33 @@ void AvformatProducerWidget::onFrameDecoded()
...
                 }
                 ui->videoTableWidget->setItem(3, 1, new QTableWidgetItem(pix_fmt));
+                key = QString("meta.media.%1.codec.colorspace").arg(i);
+                int colorspace = m_producer->get_int(key.toLatin1().constData());
+                QString csString = tr("unknown");
+                switch (colorspace) {
+                    case 240: csString = "SMPTE ST240"; break;
+                    case 601: csString = "ITU-R BT.601"; break;
+                    case 709: csString = "ITU-R BT.709"; break;
+                    case 2020: csString = "ITU-R BT.2020"; break;
```
ddennedy, line 379: `case 2020` will have to change per the latest MLT branch.

bmatherly: Done.

ddennedy (approved):

ddennedy:
I noticed that if I click Cancel and then Convert in the Properties panel, it does not automatically convert to BT.709. Was that intentional? What is your thought on that?

bmatherly:
> I noticed that if I click Cancel and then Convert in the Properties panel, it does not automatically convert to BT.709. Was that intentional? What is your thought on that?

Yes. That was intentional. 

I call  dialog.set709Convert(true) when prompting based on color transfer characteristics. But I did not change it to convert by default. If the user cancels and then clicks "convert" they would have to check the 709 box. My thinking is that I do not know what happens if we run an existing 709 clip through that processing. Might the pixels round to different numbers? So the checkbox should only be checked when prompting the user for this specifically, or if the user checks it.

I am open to suggestions on that.

ddennedy:
But when the user clicks Convert we also know what the colorspace and transfer are. Should we turn it on based on transfer?

bmatherly:
That is a good point. it would not be hard to add the check for the button click.

Maybe the user will be confused that sometimes it is checked automatically and some time it is not.

Do we even need the checkbox?
Or should we give a message when we check it on their behalf?
Or we could display the advanced section when we check it so they might noticed it is checked?

I propose:
* Keep the check box
* Perform the HDR test any time the dialog is opened
* If we check the box for them, also show the advanced section automatically

ddennedy:
The first two bullet items are done plus a couple other small things I noticed while testing tonight. I do not feel the need for item 3, but go ahead if you want.

PR #998, opened 2021-01-01
remove private qt headers

xdch47:
Not neccesary as no `#include <private/*>` in the source files.

Remove qmake warning:
```
Project MESSAGE: This project is using private headers and will therefore be tied to this specific Qt module build version.
Project MESSAGE: Running this project against other versions of the Qt modules may crash at any arbitrary point.
Project MESSAGE: This is not a bug, but a result of using Qt internals. You have been warned!
```


ddennedy:
Thanks. Private usage was removed in e494b76 but this was overlooked.

PR #1005, opened 2021-01-14
Check for missing files before export

### LINKED ISSUE #990: When deleting a source file, nothing alterts me when exporting that a source file is missing
Steps to reproduce

- add source files, make a video montage (not sure is this is the correct word)
- delete a source file 
- you'll be able to edit, see preview as if everything is doing well
- when you're happy, try to export the film
- at this step, nothing warn you that a source file is missing, ie the world "INVALID" in the exported video (I was sure it was a codec problem, or a codec / container format incompatibility) 

Proposal: before exporting, check that all source files are present and warn the user if one ore more source files are missing (present a list of those file could be perfect).

I'm sure I'm not the only one that had this problem ([googling "shotcut export invalid"](https://www.google.com/search?q=shotcut+export+%22invalid%22))

Shotcut v20.11.28

bmatherly:
Fix #990: When deleting a source file, nothing alterts me when
exporting that a source file is missing

### DIFF
--- a/src/docks/encodedock.cpp
+++ b/src/docks/encodedock.cpp
@@ -1389,6 +1389,11 @@ void EncodeDock::on_encodeButton_clicked()
         return;
     }
     bool seekable = MLT.isSeekable(fromProducer());
+
+    if (seekable && checkForMissingFiles()) {
+        return;
+    }
+
     QString directory = Settings.encodePath();
     if (!m_extension.isEmpty()) {
         if (!MAIN.fileName().isEmpty()) {
@@ -2041,4 +2046,40 @@ bool EncodeDock::detectHardwareEncoders()
         Settings.setEncodeHardware(hwlist);
     }
     return !hwlist.isEmpty();
+}
+
+bool EncodeDock::checkForMissingFiles()
+{
+    Mlt::Producer* service = fromProducer();
+    if (!service) {
+        service = MAIN.playlist();
+    }
+    if (!service) {
+        LOG_ERROR() << "Encode: No service to encode";
+        return true;
+    }
+    QScopedPointer<QTemporaryFile> tmp{Util::writableTemporaryFile(MAIN.fileName())};
+    tmp->open();
+    tmp->close();
+    QString fileName = tmp->fileName();
+    tmp->remove();
+    auto isProxy = ui->previewScaleCheckBox->isChecked() && Settings.proxyEnabled();
+    MLT.saveXML(fileName, service, false /* without relative paths */, nullptr, isProxy);
+    MltXmlChecker checker;
+    if (!checker.check(fileName)) {
+        LOG_ERROR() << "Encode: Unable to check XML - skipping check";
+    } else if (checker.unlinkedFilesModel().rowCount() > 0) {
+        QMessageBox dialog(QMessageBox::Critical,
+           qApp->applicationName(),
+           tr("Your project is missing some files.\n\n"
+              "Save your project, close it, and reopen it.\n"
+              "Shotcut will attempt to repair your project."),
+           QMessageBox::Ok,
+           this);
+        dialog.setWindowModality(QmlApplication::dialogModality());
+        dialog.setDefaultButton(QMessageBox::Ok);
+        dialog.exec();
+        return true;
+    }
+    return false;
 }
--- a/src/docks/encodedock.h
+++ b/src/docks/encodedock.h
@@ -159,6 +159,7 @@ private:
     Mlt::Producer* fromProducer() const;
     static void filterX265Params(QStringList& other);
     void onVideoCodecComboChanged(int index, bool ignorePreset = false);
+    bool checkForMissingFiles();
 };
 
 #endif // ENCODEDOCK_H


src/docks/encodedock.cpp
```diff
@@ -2042,3 +2047,38 @@ bool EncodeDock::detectHardwareEncoders()
...
+    if (!service) {
+        LOG_ERROR() << "Encode: No service to encode";
+        return true;
+    }
+    QTemporaryFile tmp;
+    tmp.open();
+    tmp.close();
+    QString fileName = tmp.fileName();
+    tmp.remove();
+    MLT.saveXML(fileName, service, false /* without relative paths */, nullptr, Settings.proxyEnabled());
```
ddennedy, line 2066: This is a different logic than lines 926 - 927:
```
    auto isProxy = ui->previewScaleCheckBox->isChecked() && Settings.proxyEnabled();
    MLT.saveXML(fileName, service, false /* without relative paths */, nullptr, isProxy);
```
Also, I do not see where this temporary file is removed after `MLT.saveXML()`

src/docks/encodedock.cpp
```diff
@@ -2042,3 +2047,38 @@ bool EncodeDock::detectHardwareEncoders()
...
+{
+    Mlt::Producer* service = fromProducer();
+    if (!service) {
+        service = MAIN.playlist();
+    }
+    if (!service) {
+        LOG_ERROR() << "Encode: No service to encode";
+        return true;
+    }
+    QTemporaryFile tmp;
```
ddennedy, line 2061: Use `QScopedPointer<QTemporaryFile> tmp(Util::writableTemporaryFile(filename));` where possible. 
Since you do not yet have a filename you could move this to after the file save dialog, or maybe use MAIN.filename() if not empty.

src/docks/encodedock.cpp
```diff
@@ -1389,6 +1389,11 @@ void EncodeDock::on_encodeButton_clicked()
         return;
     }
     bool seekable = MLT.isSeekable(fromProducer());
+
+    if (seekable && checkForMissingFiles()) {
```
ddennedy, line 1393: Possibly instead of creating a new temporary XML file early here you can integrate with the XML file generated in `createMeltJob()`. I think a drawback is repetition when using From = Each Playlist Item. You can change that logic to break out of loop if `createMeltJob` returns null.

ddennedy (changes requested):

src/docks/encodedock.cpp
bmatherly: >This is a different logic than lines 926 - 927
OK. I will change the logic to match.

> Also, I do not see where this temporary file is removed
"the file will subsequently be removed upon destruction of the QTemporaryFile object"
https://doc.qt.io/qt-5/qtemporaryfile.html

src/docks/encodedock.cpp
bmatherly: Can you explain why a QScopedPointer is helpful here? The QTemporaryFile will be destroyed when the function exits.

Can you also explain the purpose of Util::writableTemporaryFile(filename)?

> Since you do not yet have a filename you could move this to after the file save dialog

My goal was to not even bother to show the file save dialog if there are missing files. It would be an unnecessary step.

src/docks/encodedock.cpp
bmatherly: Yes, I considered that. I did not do that because:
1) I do not even want to prompt the user for a filename if there are missing files
I am open minded about this, but it seems more convenient to me
2) "From=each playlist item" as you mentioned. I could detect the missing file in the loop and bail out. But my message to the user is to close and reopen. They can't do that if 20 encode jobs have just been queued up but the 21st had an error. But maybe that is still a better solution.

I will defer to your preference on this.

bmatherly (commented):
I also considered moving the entire check up into the main application and adding a menu option so that the user can run the check whenever they want. But then I thought that we already do the check every time the file is opened. So that should be good enough. But, unfortunately, that means that the only way a user can trigger a repair is to close and reopen a project.

What are your thoughts about moving it into the main application?

src/docks/encodedock.cpp
ddennedy: We will keep it as you have it.

src/docks/encodedock.cpp
ddennedy: I saw the `tmp.remove()` and thought it might invalidate the auto-removal, but I think you are correct.

ddennedy: The purpose of `Util::writableTemporaryFile(filename)` is to fix a bug where some user's system temp dir was not writable. Maybe someone does that to prevent saving personal info on a shared or public computer. I do not know. Or perhaps it is faulty setup. Nevertheless, it was found troubleshooting a bug report. In that case, the method creates the temp file in the same folder as `filename` where and usually filename was already checked with `Util::warnIfNotWritable()`.

The scoped pointer is used because the method returns a pointer to a new QTemporaryFile and does not destroy it.

src/docks/encodedock.cpp
ddennedy: I saw the `tmp.remove()` and thought it might invalidate the auto-removal, but I think you are correct.

bmatherly:
@ddennedy Another thought occurred to me. Would it work to scan the auto-save file before exporting? Upon clicking the export button, Shotcut could create a new auto-save file and scan that. But I do not know if an auto-save file always exists. Thoughts?

ddennedy:
The autosave file approach would not be sensitive to the Export > From field as you have it now.

bmatherly:
> The autosave file approach would not be sensitive to the Export > From field as you have it now.

That is true. It would be checking the entire project as it does when a project is opened. I guess there are pros and cons to that.

src/docks/encodedock.cpp
bmatherly: Thanks for the explanation. This is done.

ddennedy: I did some testing, and when MAIN.fileName() is empty because it is an unsaved project, then it uses "." for the directory of the temp file when $TMPDIR (%TEMP% on Windows) is not writable. This ends up trying to use the location of the shotcut binary executable, which is usually an unwritable location.  Also, the temp file was not auto-removed. I will fix it in a commit after merge with
```
--- src/docks/encodedock.cpp
+++ src/docks/encodedock.cpp
@@ -2059,5 +2059,10 @@ bool EncodeDock::checkForMissingFiles()
         return true;
     }
-    QScopedPointer<QTemporaryFile> tmp{Util::writableTemporaryFile(MAIN.fileName())};
+    QScopedPointer<QTemporaryFile> tmp;
+    if (MAIN.fileName().isEmpty()) {
+        tmp.reset(Util::writableTemporaryFile(QStandardPaths::writableLocation(QStandardPaths::AppDataLocation) + "/"));
+    } else {
+        tmp.reset(Util::writableTemporaryFile(MAIN.fileName()));
+    }
     tmp->open();
     tmp->close();
@@ -2080,6 +2085,8 @@ bool EncodeDock::checkForMissingFiles()
         dialog.setDefaultButton(QMessageBox::Ok);
         dialog.exec();
+        QFile::remove(fileName);
         return true;
     }
+    QFile::remove(fileName);
     return false;
 }
```

PR #1022, opened 2021-01-24
fix #1004 Toolbar buttons are different sizes

### LINKED ISSUE #1004: Toolbar buttons are different sizes.
using windows 10, version 21.01.12 (nightly)

(21.01.12)
image

(20.11.28)
image

bmatherly:
20.11.28 using Controls 1:
image

Master using Controls 2:
image

This proposal using Controls 2 and explicit sizing:
image

The relative icon size can be easily changed by increasing or decreasing the button padding if you want to try larger or smaller. Lower padding results in larger icons.
https://github.com/mltframework/shotcut/compare/master...bmatherly:toolbar_icon?expand=1#diff-be25bb78c0bd5398a778f928cf7ffbec62fff56a373d6720a39fe86e4fff3ccdR27

### DIFF
--- a/src/qml/modules/Shotcut/Controls/ToolButton.qml
+++ b/src/qml/modules/Shotcut/Controls/ToolButton.qml
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Meltytech, LLC
+ * Copyright (c) 2020-2021 Meltytech, LLC
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -21,21 +21,11 @@ import Shotcut.Controls 1.0 as Shotcut
 
 ToolButton {
     id: control
-    property string iconName
-    property url iconSource
-    property alias tooltip: tooltip.text
 
-    background: Rectangle {
-        id: rect
-        radius: 3
-        width: control.width
-        height: control.height
-        SystemPalette { id: activePalette }
-        color: control.checked? activePalette.highlight : activePalette.button
-    }
-    action: Action {
-        icon.name: iconName
-        icon.source: iconSource
-    }
-    Shotcut.HoverTip { id: tooltip }
+    icon.height: control.height - (verticalPadding * 2)
+    icon.width: control.height -  (horizontalPadding * 2)
+    padding: 3
+
+    SystemPalette { id: activePalette }
+    palette.button: checked ? activePalette.highlight : activePalette.button
 }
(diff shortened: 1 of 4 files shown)


ddennedy:
If we are going to manage the size, then I want it to respond to `settings.smallIcons`

bmatherly:
I think I am ignorant about how the size is currently managed. We are not currently using settings.smallIcons to determine the size. So what does manage the size?

ddennedy:
Your introduction of implicitWidth and implicitHeight is managing the size again. Look at 20.11.28 code to see that we used to manage the size and used settings.smallIcons. Run the older version, toggle that option in the View menu to see the size change.

bmatherly:
I see that 20.11.28 uses smallIcons or an inferred height from a button. But current git master has a fixed toolbar height:
https://github.com/mltframework/shotcut/blob/master/src/qml/views/timeline/TimelineToolbar.qml#L32
Are you saying that it is an oversight that current git master has a fixed height?

ddennedy:
No, I am simply saying make this use smallIcons similar to v20.11.28. Did you run it and see the behavior? I don't care what the current toolbar height is; if I remove or change it in the version in git master, the icon size does not change.

ddennedy:
Toolbar height is set because it looks like crap without it.

image

bmatherly:
Thanks for the clarification. For this latest commit:

Without "Show Small Icons":
image

With "Show Small Icons":
image

Without "Show Small Icons", the height is inferred from an invisible button. I had to add an icon to the hidden button because the default size of a button with no text or icon is very small.

ddennedy (approved):

ddennedy:
Testing this out now on macOS, and the icon sizes are not changing. I do not know if that is a macOS or Qt 5.12 thing. I think that is why I had trouble before. The icons do resize on Windows and Linux. When I test this side-by-side with version 20.11.28 on Windows and Linux, there is still quite a difference. With small icons, the new version is larger, and without small icons it is smaller. I was able to tweak this by not basing it on a default implicitHeight. `settings.smallIcons? 25 : 34` gives me the closest (while disregarding Qt default).

bmatherly:
FWIW: I tested this with Qt 5.12 on Linux. So it may not be related to the Qt version.

I think that fundamentally Controls 2 chooses default sizes differently than Controls 1 did. I am not sure that using a height inferred from an invisible control will be reliable. I think that specifying two fixed sizes (as you suggest) will give the most consistent results.

It is also OK with me if you just revert this change and move on. Maybe there are diminishing returns.

PR #1023, opened 2021-01-30
Add multi-file name dialog for export each playlist item

bmatherly:
This allows the user to create custom names for the exported files
substituting file properties into fields in the file name.

Inspired by this discussion: https://forum.shotcut.org/t/50-mov-videos-to-mp4/24152

image

Includes error checking for duplicate file names and existing files (to avoid overwrite)
image

image

The OK button is disabled until errors are resolved.

### DIFF
--- a/src/dialogs/multifileexportdialog.cpp
+++ b/src/dialogs/multifileexportdialog.cpp
@@ -1,0 +1,265 @@
+/*
+ * Copyright (c) 2019-2020 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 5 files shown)


ddennedy:
On Windows
```
[Warning] <> QObject::connect: signal not found in QPushButton
[Warning] <> QObject::connect: signal not found in QLineEdit
[Warning] <> QObject::connect: signal not found in QComboBox
[Warning] <> QObject::connect: signal not found in QComboBox
[Warning] <> QObject::connect: signal not found in QComboBox
[Warning] <> QObject::connect: signal not found in QLineEdit
```

ddennedy:
I worked around the above because I could not resolve it. I have seen this problem before on Windows, but it is not consistent: There are other new-style connections that work fine.

bmatherly:
Thanks. I had some trouble with the old-style connections not working on Linux. This is a clever workaround.

PR #1027, opened 2021-02-04
Decrease speed muliplication on fastForward

gmartin42:
In addition to your change "Improved the playback speeds of fast forward and rewind to not be so fast and more usable."
I think you should reduce the speed increase of fastforward button since it can be accelerated by pressing on the button again.
This will permit for instance to still understand people talking in the video. Since it is quite imposible at the double speed.


ddennedy:
This does not work. The engine only supports integer multiplied despite the parameter type.

gmartin42:
> 
> 
> This does not work. The engine only supports integer multiplied despite the parameter type.

too bad for me thanks for your answer !

PR #1042, opened 2021-02-17
Negative keyframes 1

bmatherly:
Depends on https://github.com/mltframework/mlt/pull/659

Here is a change for your consideration:
* Do not allow negative keyframes.
* If the user changes the in point to cause a negative keyframe, create a keyframe at position 0 and delete negative keyframes
* Allow keyframes greater than the out point
* Keyframes after the out point are "dimmed" visually to indicate that they are inactive
* If the out point is changed back so that keyframes are in range again, they will still be at the original position
* Remove some redundant interactions. For example, sometimes multiple signals trigger the same function to be called multiple times unnecessarily

There is still some interaction between the timeline and the keyframe in/out points that I have not figured out. Trimming filter in/out in the keyframe dock seems reliable. But trimming clip in/out in the timeline still causes keyframe corruption. I might need help with that.

### DIFF
--- a/src/docks/keyframesdock.cpp
+++ b/src/docks/keyframesdock.cpp
@@ -103,7 +103,6 @@ void KeyframesDock::setCurrentFilter(QmlFilter* filter, QmlMetadata* meta)
     connect(filter, SIGNAL(animateInChanged()), &m_model, SLOT(reload()));
     connect(filter, SIGNAL(animateOutChanged()), &m_model, SLOT(reload()));
     connect(filter, SIGNAL(inChanged(int)), &m_model, SLOT(onFilterInChanged(int)));
-    connect(filter, SIGNAL(outChanged(int)), &m_model, SLOT(onFilterOutChanged(int)));
 }
 
 bool KeyframesDock::event(QEvent *event)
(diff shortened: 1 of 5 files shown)


bmatherly:
@ddennedy friendly reminder about this one. I think this is a good change to improve user experience. And it would be nice to include it before we build another alpha/beta.

ddennedy:
I had the impression I needed to help, and considering nature of the change, I wanted to make a thorough review. I have not gotten around to it yet. I can probably Monday night.

bmatherly:
This change improves keyframe interaction for filter trimming without making clip trimming worse. I think we should try to improve the keyframe interaction when trimming clips, but that could be done in a separate commit/effort.

For this review, I mostly need you to experience the new behavior to make sure it is an improved user experience. The new behavior is deleting keyframes that become negative and adding new keyframes at the interpolation point. Also, keyframes after the filter out point are displayed but greyed out.

If you have time to try it on Monday, here are some test steps:
1) Open a clip
2) Add a gain filter
3) Add keyframes to the gain filter
4) Trim the filter in point so that it goes past a keyframe and observe the behavior
5) Trim the filter out point so that it goes past a keyframe and observe the behavior
6) Apply this patch and repeat steps 1-5 to see the changed behavior

PR #1043, opened 2021-02-17
Use cmake to build MLT

bmatherly:
Putting this out there in case you are interested. It worked for me in Linux.

### DIFF
--- a/scripts/build-shotcut.sh
+++ b/scripts/build-shotcut.sh
@@ -706,24 +706,7 @@ function set_globals {
 
   #####
   # mlt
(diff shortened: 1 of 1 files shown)


ddennedy:
It does not work on macOS and Windows yet.

bmatherly:
Correct. And I do not expect you to drop everything to make it work for Win/Mac. Do you think this is something we want to do soon?

eszlari:
With the last commit in this PR: https://github.com/mltframework/mlt/pull/660, I got it working under Windows.

ddennedy:
I am not eager to do anything soon on MLT or Shotcut except to release version 21.02. ;-) But this is something I expect to be done for the 21.04 release candidate due in early April.

bmatherly:
That is good to know. I do not plan to do any more with this pull request. Feel free to leave it open, or close it if you think you will take another approach.

ddennedy:
I forgot about this PR when I made my branch `mlt_cmake` and got it working. Main difference is that I am not using ninja yet. That could be added.

PR #1049, opened 2021-03-01
integrate Properties > Convert with Time Remap

ddennedy:
This replaces the warning dialog when choosing Time Remap with the transcode dialog used by Properties > Convert.


bmatherly (approved):
Looks good to me.

Any idea what to do about proxy files and how they would mask a source file with b frames?

PR #1050, opened 2021-03-08
Fix FreeBSD segfault.

fernape:
Implement a way to get an idea of the free memory available on FreeBSD.

It can be argued how accurate this is, but we apply this patch in out ports infrastructure. Without it, shotcut segfaults.

fernape@FreeBSD.org


ddennedy:
Thank you

PR #1062, opened 2021-03-29
Add chocolatey badges

adriens:
Add chocolatey badges so people know very quick they know : 

- how popular is the software
- how easy it is to install it with chocolatey


ddennedy:
Sorry, but I do not support or endorse using Chocolatey to install Shotcut. Not only is it not managed by us, but I can see from the version history that it is publishing pre-release versions. The primary reason, however, is that our download page and the Microsoft Store are sources of funding for the project, and Chocolatey bypasses them. Obviously, it is permissible for Chocolatey to distribute Shotcut, but I decline to promote it.

PR #1071, opened 2021-04-11
Fix Shotcut hangs when changing speed on clip

bmatherly:
Steps to reproduce:
1) Set profile to a manual profile (not automatic)
2) Open a clip
3) Change the clip speed parameter
Shotcut will hang

The problem does not occur for clips on the timeline

### DIFF
--- a/src/mltcontroller.cpp
+++ b/src/mltcontroller.cpp
@@ -200,9 +200,7 @@ bool Controller::openXML(const QString &filename)
 
 void Controller::close()
 {
-    if (m_profile.is_explicit()) {
-        pause();
-    } else if (m_consumer && !m_consumer->is_stopped()) {
+    if (m_consumer && !m_consumer->is_stopped()) {
         m_consumer->stop();
     }
     if (isSeekableClip()) {


bmatherly:
Dan, would like you to inspect this because I do not know why the check for explicit profile is in the close function.

ddennedy:
Yes I want to review it later today

ddennedy:
By using git blame and following parent revisions (Qt Creator makes this easy), we get to b82b3b1b as the reason: possibly GPU Effects. I do not recall exactly why except that GPU Effects is less stable when needing to stop and restart the consumer. It might have also been a general optimization added at the same time while making it work with GPU Effects. Stopping and restarting a consumer is generally heavier than pausing and resuming especially due to destroying and creating threads, which gets into `RenderThread` in `glwidget.cpp` and sharing OpenGL contexts. In theory, you only need to fully restart the consumer when in Automatic mode because the MLT profile provides the consumer properties with which the consumer is initialized when starting. And combine all that with the fact that `Mlt::Controller::close()` is called whenever `Mlt::Controller::open()` is called. You see, close is not literally intended to be a full close operation. It is more like this is what needs to be done before changing the controller's producer.

I did not reproduce this when testing on macOS with either 21.03.12 or yesterday's build - even while it was playing and trying several times. What is your reproducibility?

bmatherly:
I recreate the hang using this build on Ubuntu 20.04:
https://github.com/mltframework/shotcut/actions/runs/737762912
I do not recreate it in Windows with the build from the same date.

The circumstances to the hang seems to be very specific:
1) Ubuntu 20.04
2) Profile must be manual
3) Clip must be in the source, not timeline or playlist
4) The player must be paused. I do not experience a crash if I change the speed while the player is playing - it just plays right through and the speed changes without any problems.
5) Realtime (Frame dropping) must be enabled. I do not reproduce the hang when realtime is disabled

Maybe my environment creates the perfect scenario to create this otherwise rare bug.

> Stopping and restarting a consumer is generally heavier than pausing and resuming especially due to destroying and creating threads

That may be true. However, I would say that I do not notice/experience any slow down with the change. 

I can keep digging to try to get to the root of the problem. What I do observe is that without calling close there seems to be multiple threads operating on the producer with the consumer pulling frames/image/audio simultaneous to the avformatproducerwidget pulling frame/image to refresh the widget UI. Then, the consumer stops notifying that new frames are available even though my debug trace indicates that the frame queue is full.

ddennedy:
OK, I reproduced it now including on macOS. A key factor for me in addition to being paused is to use a video mode with a high frame rate like 60.

PR #1076, opened 2021-04-24
Export image improvements

bmatherly:
In response to this suggestion:
https://forum.shotcut.org/t/suggestion-a-little-enhancement-for-export-frame-option/26920
* Suggest a file name in the file dialog
* Remember the previously used extension and default to it for every
  export request
* Automatically set the file name extension based on the selected
  file filter

With these changes, an image can be exported by typing "Ctl-Shift-E" + "Enter"
without the need to interact with the dialog if the restored defaults are
suitable.

### DIFF
--- a/src/dialogs/saveimagedialog.cpp
+++ b/src/dialogs/saveimagedialog.cpp
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2021 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "saveimagedialog.h"
+
+#include "mltcontroller.h"
+#include "settings.h"
+#include "util.h"
+
+#include <QDebug>
+#include <QtMath>
+
+static QString suffixFromFilter(const QString& filterText)
+{
+    QString suffix = filterText.section("*", 1, 1).section(")", 0, 0).section(" ", 0, 0);
+    if ( !suffix.startsWith(".") )
+    {
+        suffix.clear();
+    }
+    return suffix;
+}
+
+SaveImageDialog::SaveImageDialog(QWidget *parent, const QString& caption, QImage& image)
+    : QFileDialog(parent, caption)
+    , m_image(image)
+{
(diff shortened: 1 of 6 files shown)


src/dialogs/saveimagedialog.cpp
```diff
@@ -0,0 +1,130 @@
...
+    QString nameFilter = tr("PNG (*.png);;BMP (*.bmp);;JPEG (*.jpg *.jpeg);;PPM (*.ppm);;TIFF (*.tif *.tiff);;WebP (*.webp);;All Files (*)");
+    setNameFilter(nameFilter);
+
+    QStringList nameFilters = nameFilter.split(";;");
+    QString suffix = Settings.snapSuffix();
+    QString selectedNameFilter = nameFilters[0];
+    for ( const auto& f : nameFilters )
+    {
+        if ( f.contains(suffix) )
+        {
```
ddennedy, line 57: This is using MLT old code style, not Shotcut's.

src/dialogs/saveimagedialog.cpp
```diff
@@ -0,0 +1,130 @@
...
+            selectedNameFilter = f;
+            break;
+        }
+    }
+    selectNameFilter(selectedNameFilter);
+
+    // Use the current player time as a suggested file name
+    QString nameSuggestion = tr("shotcut_%1").arg(MLT.producer()->frame_time( mlt_time_clock ));
```
ddennedy, line 65: "shotcut" should be capitalized and not translatable.

src/settings.cpp
```diff
@@ -225,6 +225,16 @@ void ShotcutSettings::setViewMode(const QString& viewMode)
     emit viewModeChanged();
 }
 
+QString ShotcutSettings::snapSuffix() const
+{
+    return settings.value("snapSuffix", ".png").toString();
```
ddennedy, line 230: I do not like the key name "snap" I had to stop and think for many seconds about what this is until I realize "snapshot." "Snapshot" is kinda OK except we do not have anything called snapshot. `exportFrameSuffix`

ddennedy (changes requested):

src/dialogs/saveimagedialog.cpp
bmatherly: Done

src/dialogs/saveimagedialog.cpp
bmatherly: Done

I think I took some inspiration from VLC which uses the name: "vlcsnap-<timecode>.png". I debated adding a word like "export" or "frame". I do not have a strong opinion about this.  But perhaps some users will since it is not configurable and I would prefer to not add a user configuration for the name format.

src/settings.cpp
bmatherly: Done

bmatherly:
I would also be interested in your opinion about changing the suffix on the file name in the dialog. This is not a built-in feature, so I suppose it is non-standard. I do observe that in the Save-As dialog in MS Paint, it changes the extension as you choose different formats. So this follows that, somewhat.

ddennedy:
> your opinion about changing the suffix on the file name in the dialog.

That is fine since this function depends heavily on suffix.

PR #1077, opened 2021-04-25
Add a button to zoom in on keyframe values

bmatherly:
Attempt to resolve some comments from this discussion:
https://forum.shotcut.org/t/test-the-version-21-04-pre-release/26818/23

### DIFF
--- a/src/qml/views/keyframes/ParameterHead.qml
+++ b/src/qml/views/keyframes/ParameterHead.qml
@@ -202,6 +202,19 @@ Rectangle {
 //                onClicked: timeline.setTrackLock(index, !isLocked)
                 Shotcut.HoverTip { text: isLocked? qsTr('Unlock track') : qsTr('Lock track') }
             }
+
+            Shotcut.ToolButton {
+                Shotcut.HoverTip { text: qsTr('Zoom keyframe values') }
+                focusPolicy: Qt.NoFocus
+                visible: delegateIndex >= 0 && paramHeadRoot.isCurve
+                checkable: true
+                action: Action {
+                    id: zoomFitKeyframeAction
+                    icon.name: 'zoom-fit-best'
+                    icon.source: 'qrc:///icons/oxygen/32x32/actions/zoom-fit-best.png'
+                    onTriggered: root.paramRepeater.itemAt(delegateIndex).zoomHeight = checked
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/qml/views/keyframes/ParameterHead.qml
```diff
@@ -202,6 +202,19 @@ Rectangle {
 //                onClicked: timeline.setTrackLock(index, !isLocked)
                 Shotcut.HoverTip { text: isLocked? qsTr('Unlock track') : qsTr('Lock track') }
             }
+
+            Shotcut.ToolButton {
+                Shotcut.HoverTip { text: qsTr('Zoom keyframe values') }
+                focusPolicy: Qt.NoFocus
+                visible: delegateIndex >= 0 && paramHeadRoot.isCurve
```
ddennedy, line 209: I am a little worried about exposing this on every parameter with curves especially considering there is no way to reset or zoom in or out other than to re-select the filter. What do you think about limiting this to Time Remap for now? I am also worried that this whole change will disappoint when people realize the scale is not saved.

bmatherly: > there is no way to reset or zoom in or out other than to re-select the filter.

Can you explain what you mean by that? The button is a toggle so it is easy to switch back-and-forth between zoomed in and zoomed out. When do you need to re-select the filter?

I agree that people will ask to have the scaling saved. But those requests will be better than the support requests from people who become confused because they selected the zoom long ago and have forgotten.

I actually find this very useful for some other filters like the gain filter where it is difficult to use the keyframes for precision without the zoom. But I also agree that the intuitiveness is not perfect yet. I am thinking about adding value labels to the y axis to show the max, middle and min values. I think this would help with the intuitiveness. What do you think about that idea?

I also wonder your opinion about the highlighted toggle button. I like the highlighting because it brings attention to the fact that it is zoomed. But maybe the aesthetic is not great?

image

ddennedy: > The button is a toggle

I had overlooked that because I did not actually test it. That reduces my concern.

> your opinion about the highlighted toggle button

That is fine. We use this approach on the timeline toolbar. It is definitely needed since we do not change the image.

bmatherly: I was also considering the "magnifying glass" icon for the button - which might further emphasize the "zoom" nature of it.

PR #1097, opened 2021-05-16
Shift keyframes for all filters upon split or trim

bmatherly:
When the in point of a producer is changed due to split or trim,
adjust the keyframes for all filters attached to that producer.

For simple keyframes, keyframes are always relative to the start and
end of the filter. So the 3rd and 4th keyframe need to be moved if
the out point changes.

For advanced keyframes, all keyframes are relative to the start of
the producer. So all keyframes need to be moved if the in point
changes.

Issue reported here:
https://forum.shotcut.org/t/inconsistent-handling-of-keyframes-when-splitting-a-clip/27404

This passed my inspection and testing, but it changes many things and will need more inspection and testing.

### DIFF
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -109,6 +109,8 @@ signals:
     void openFailed(QString);
     void aboutToShutDown();
     void renameRequested();
+    void serviceInChanged(int delta, Mlt::Service*);
+    void serviceOutChanged(int delta, Mlt::Service*);
 
 protected:
     MainWindow();
--- a/src/mltcontroller.cpp
+++ b/src/mltcontroller.cpp
@@ -736,94 +736,26 @@ void Controller::next(int currentPosition)
 void Controller::setIn(int in)
 {
     if (m_producer && m_producer->is_valid()) {
-        // Adjust filters.
-        bool changed = false;
-        int n = m_producer->filter_count();
-        for (int i = 0; i < n; i++) {
-            QScopedPointer<Filter> filter(m_producer->filter(i));
-            if (filter && filter->is_valid()) {
-                if (QString(filter->get(kShotcutFilterProperty)).startsWith("fadeIn")) {
-                    if (!filter->get(kShotcutAnimInProperty)) {
-                        // Convert legacy fadeIn filters.
-                        filter->set(kShotcutAnimInProperty, filter->get_length());
-                    }
-                    filter->set_in_and_out(in, filter->get_out());
-                    changed = true;
-                } else if (!filter->get_int("_loader") && filter->get_in() == m_producer->get_in()) {
-                    filter->set_in_and_out(in, filter->get_out());
-                    changed = true;
-                }
-            }
+        int delta = in - m_producer->get_in();
+        if (!delta) {
+            return;
         }
-        if (changed)
-            Controller::refreshConsumer();
+        adjustClipFilters(*m_producer, m_producer->get_in(), m_producer->get_out(), delta, 0);
         m_producer->set("in", in);
+        Controller::refreshConsumer();
     }
 }
 
 void Controller::setOut(int out)
 {
     if (m_producer && m_producer->is_valid()) {
-        // Adjust all filters that have an explicit duration.
-        bool changed = false;
-        int n = m_producer->filter_count();
-        for (int i = 0; i < n; i++) {
-            QScopedPointer<Filter> filter(m_producer->filter(i));
-            if (filter && filter->is_valid()) {
-                QString filterName = filter->get(kShotcutFilterProperty);
-                if (filterName.startsWith("fadeOut")) {
-                    if (!filter->get(kShotcutAnimOutProperty)) {
-                        // Convert legacy fadeOut filters.
-                        filter->set(kShotcutAnimOutProperty, filter->get_length());
-                    }
-                    filter->set_in_and_out(filter->get_in(), out);
-                    changed = true;
-                    if (filterName == "fadeOutBrightness") {
-                        const char* key = filter->get_int("alpha") != 1? "alpha" : "level";
-                        filter->clear(key);
-                        filter->anim_set(key, 1, filter->get_length() - filter->get_int(kShotcutAnimOutProperty));
-                        filter->anim_set(key, 0, filter->get_length() - 1);
-                    } else if (filterName == "fadeOutMovit") {
-                        filter->clear("opacity");
-                        filter->anim_set("opacity", 1, filter->get_length() - filter->get_int(kShotcutAnimOutProperty), 0, mlt_keyframe_smooth);
-                        filter->anim_set("opacity", 0, filter->get_length() - 1);
-                    } else if (filterName == "fadeOutVolume") {
-                        filter->clear("level");
-                        filter->anim_set("level", 0, filter->get_length() - filter->get_int(kShotcutAnimOutProperty));
-                        filter->anim_set("level", -60, filter->get_length() - 1);
-                    }
-                } else if (!filter->get_int("_loader") && filter->get_out() == m_producer->get_out()) {
-                    filter->set_in_and_out(filter->get_in(), out);
-                    changed = true;
-
-                    // Update simple keyframes of non-current filters.
-                    if (filter->get_int(kShotcutAnimOutProperty) > 0
-                        && MAIN.filterController()->currentFilter()
-                        && MAIN.filterController()->currentFilter()->service().get_service() != filter.data()->get_service()) {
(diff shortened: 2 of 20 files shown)


src/mainwindow.h
```diff
@@ -109,6 +109,8 @@ class MainWindow : public QMainWindow
     void openFailed(QString);
     void aboutToShutDown();
     void renameRequested();
+    void serviceInChanged(int delta, Mlt::Service*);
```
bmatherly, line 112: I am not sure about the location of this. But MainWindow has access to all the panels that need to be connected to this event.

src/models/keyframesmodel.h
```diff
@@ -86,6 +86,8 @@ public slots:
     void reload();
     void onFilterChanged(const QString& property);
     void onFilterInChanged(int delta);
+    void trimFilterIn(int in);
```
bmatherly, line 89: Not sure about the location of these functions. The QML needs a function to call that is not the filter so that the adjustments can be made. Could also go in keyframesdock, for example.

bmatherly (commented):
I am open to feedback about where some functions/signals should live.

ddennedy:
This change is very heavy and will need a lot of scrutiny and testing. The explanation

> For advanced keyframes, all keyframes are relative to the start of the producer. So all keyframes need to be moved if the in point changes.

is either ambiguous or sounds counter to my design. The problem is "producer" can be either the original producer or the cut producer. The design is that advanced keyframes are relative to the original producer. That is, if I need a parameter to change when the media does X (as in correction or reaction), I want the keyframe to stay on that frame when the in point changes.

When I read the bug report and in particular https://forum.shotcut.org/t/inconsistent-handling-of-keyframes-when-splitting-a-clip/27404/6 and ignore kagsundaram because I have not vetted his replies yet, this seems like an awfully big change for what is described. To be honest, I started ignoring him in this thread at the mention of ripple and ripple-all-tracks when the poster does not include that in his description or screenshots, which I did reproduce. Your change does address that, but does it do more? One side effect of this change is that nothing is selected after a split.

bmatherly:
> This change is very heavy and will need a lot of scrutiny and testing.

I agree. I tried to break it into smaller pieces, but many things are necessarily interconnected. The final version of this patch will require a beta release with a special request for testing on this.

> The explanation is either ambiguous or sounds counter to my design. The problem is "producer" can be either the original producer or the cut producer. The design is that advanced keyframes are relative to the original producer. That is, if I need a parameter to change when the media does X (as in correction or reaction), I want the keyframe to stay on that frame when the in point changes.

I agree about ambiguity and I am not exactly sure how to word it.

The way I see it, an advanced keyframe should always be associated with the original frame that it was originally created with. However, keyframe positions are stored in the filter relative to the start of the filter. So, when the in point of a clip or attached filter changes, the keyframe must be adjusted in the opposite direction so that it still references the original frame that it was associated with.

>  I started ignoring him in this thread at the mention of ripple and ripple-all-tracks when the poster does not include that in his description or screenshots, which I did reproduce.

I agree that comment might be off-topic. I have just been focused on the need to adjust keyframes when a clip or filter "in" point changes.

> Your change does address that, but does it do more?

The more I tested, the more inconsistencies I found:

1) Trim clip in source player
   - did not adjust keyframes at all
   - fixed by this patch
2) Trim clip in timeline
   - keyframes are only adjusted for a filter that is being displayed in the keyframe panel
   - Filters not being displayed did not get adjusted
   - fixed by this patch
3) Split clip in timeline
   - keyframes are only adjusted for a filter that is being displayed in the keyframe panel
   - Filters not being displayed did not get adjusted
   - **I believe this is the bug reported in the forum**
   - fixed by this patch
4) Trim filter in keyframe panel
   - keyframes were properly adjusted before and that is not broken by this patch.
   - Now using common code for all these scenarios

> One side effect of this change is that nothing is selected after a split.

Correct. That could also be a separate commit or I could look for a better solution. The fundamental problem is that after a split, the keyframe panel did not refresh properly and it continued to show the filter from the unselected clip. So maybe more signals need to be connected. But upon inspection, I see that there is a firestorm of signals and having "nothing" selected reduced the signals being fired significantly. So this was an easy way out. Maybe the selection could be remembered and restored after the split.

ddennedy:
Thanks for the clarifications.

bmatherly:
@ddennedy will you consider this change for the next release? I think I could break it into some smaller commits if that would be helpful.

src/mainwindow.h
ddennedy: It is true that MainWindow is like the main controller and a patch bay for many signals. However, it can be preferable to make each component emit its own signal, and in MainWindow connect each of those. This is what it currently does for service in/out changed, but you added emissions to Mlt::Controller. Mlt::Controller is not a QObject and lacks signals, but GLWidget inherits it and adds that. However, if you add this signal to GLWidget you cannot emit it from Mlt::Controller. So, I think this stays the way you have it.

src/models/keyframesmodel.h
ddennedy: This is fine. It is a little arbitrary, but conceptually it can be argued these methods are model-oriented, and there is a `onFilterInChanged` slot here making it sensible to group them together.

src/mltcontroller.cpp
```diff
@@ -736,94 +736,26 @@ void Controller::next(int currentPosition)
...
-                    }
-                    filter->set_in_and_out(in, filter->get_out());
-                    changed = true;
-                } else if (!filter->get_int("_loader") && filter->get_in() == m_producer->get_in()) {
-                    filter->set_in_and_out(in, filter->get_out());
-                    changed = true;
-                }
-            }
+        int delta = in - m_producer->get_in();
+        if (!delta) {
```
ddennedy, line 740: Instead of checking only for 0 with an early return consider
```
if (delta > 0) {
    m_producer->set("in", in);
    adjustClipFilters(*m_producer, m_producer->get_in(), m_producer->get_out(), delta, 0);
    Controller::refreshConsumer();
}
```

src/mltcontroller.cpp
```diff
@@ -736,94 +736,26 @@ void Controller::next(int currentPosition)
...
-                                }
-                            }
-                        }
-                    }
-                }
-            }
-            if (changed)
-                Controller::refreshConsumer();
+        int delta = out - m_producer->get_out();
+        if (!delta) {
```
ddennedy, line 753: ditto

src/mltcontroller.cpp
ddennedy: Oh wait. a negative delta is OK right? Nevermind

ddennedy (changes requested):
When testing, I found filter out trimming is broken using any method: dragging trim handle, toolbar button, keyboard shortcut. Attempting to do so makes the clip duration wrong in Keyframes
image

It works for trim filter in except when using the toolbar button:
> KeyframesToolbar.qml:61: TypeError: Cannot assign to read-only property "in"

I reproduced these whether using an old project (pre MLT7) or new one with chains.

bmatherly:
Thanks for the testing. Fixed the toolbar buttons and fixed trim out.

ddennedy:
- When trimming in the source player, filters that were untrimmed (filter.in = producer.in and/or filter.out = producer.out) do not remain untrimmed.
- Tirmming the in point of a timeline clip using the keyboard shortcut <kbd>I</kbd> is not working correctly. Here I trimmed the middle clip which previously filled the blank spot. It looks OK on the Timeline but not in Keyframes
image
- If I ripple trim the in point of the timeline clip using shortcut <kbd>Shift</kbd>+<kbd>I</kbd> it behaves better but the keyframes curve is wrong, and I need to reselect the clip or filter to show it correctly.

ddennedy:
I do not want to include this in the 21.06 release. There is too much risk of regression, which is counter to the main purpose of this release.

bmatherly:
> When trimming in the source player, filters that were untrimmed (filter.in = producer.in and/or filter.out = producer.out) do not remain untrimmed.

I fixed this. I does not work properly in master, either

> Tirmming the in point of a timeline clip using the keyboard shortcut I is not working correctly

This and Shift + I are latent issues that exist in master. I made some improvement, but not perfect yet.

> I do not want to include this in the 21.06 release. There is too much risk of regression, which is counter to the main purpose of this release.

I understand the regression risk and the decision to delay is fine with me. The testing you have done is appreciated. Did you do those same tests on Master? Many of them fail in the same or different ways. So I guess the judgement is whether the regressions would be worse than the latent issues that are being fixed.

ddennedy:
No, I did not test those 3 on master. However, I did just test them on v21.03.21, and only the second one reproduces.

PR #1103, opened 2021-05-19
Add button to add files directly into the playlist (without playing it)

hbtalha:
# Motivation
In my first project with Shotcut what didn't work for me was the fact that every time I wanted to insert a media I had to switch to Windows Explorer and drag it into the playlist or if I add it through "Open File" it would play first (didn't enjoy the blasting sound of a video every time I needed to add one).

The feature turned out easy enough to implement as it was just calling existing methods.

# Preview
Screenshot_5

This is my first time ever trying to contribute to an open-source project, so please let me know of anything that I can do better in the next time or right now.

And thanks for this project. I really enjoying using Shotcut.

P.S: I use Qt Beautifier Artistic Style to automatically format the code and I 'm having a hard time trying to match (so far with no success) the coding brace style used the project. I would like to ask what is the name of the coding style used in project?


src/docks/playlistdock.ui
```diff
@@ -37,7 +37,7 @@
     <item>
      <widget class="QStackedWidget" name="stackedWidget">
       <property name="currentIndex">
-       <number>0</number>
+       <number>1</number>
```
ddennedy, line 40: Please be careful when editing a UI XML file that you do not alter the default page of a `QStackedWidget`. This must remain the same unless there is an explicit reason. Here it is a mistake.

src/docks/playlistdock.ui
```diff
@@ -59,14 +59,21 @@
            <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body&gt;
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'MS Shell Dlg 2'; font-size:7.8pt; font-weight:400; font-style:normal;&quot;&gt;
```
ddennedy, line 62: Also when editing a UI XML file with rich text like this you must revert changes like this that add font information. This font is not cross platform, and this change should not be included.

src/docks/playlistdock.ui
```diff
@@ -59,14 +59,21 @@
...
 &lt;p style=&quot; margin-top:0px; margin-bottom:16px; margin-left:-24px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;You can freely preview clips without necessarily adding them to the playlist or closing it.&lt;/p&gt;
 &lt;p style=&quot; margin-top:0px; margin-bottom:16px; margin-left:-24px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;To trim or adjust a playlist item &lt;span style=&quot; font-weight:600;&quot;&gt;Double-click&lt;/span&gt; to open it, make the changes, and click the &lt;span style=&quot; font-weight:600;&quot;&gt;Update&lt;/span&gt; icon.&lt;/p&gt;
 &lt;p style=&quot; margin-top:0px; margin-bottom:16px; margin-left:-24px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Drag-n-drop&lt;/span&gt; to rearrange the items.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
           </property>
          </widget>
         </item>
+        <item>
+         <widget class="QPushButton" name="addFilesToPlaylistButton2">
+          <property name="text">
+           <string>Add files to playlist</string>
```
ddennedy, line 73: The big button as wide as the panel is not acceptable. It should be added as an icon button.

For this we need to add a new icon - something along the lines of an import image. When dealing with icons in Shotcut, there are 3 icon themes. The dark and light fusion themes try to use something from the DefaultIcon collection, and the System theme uses one from the Oxygen collection. Our [Credits](https://shotcut.org/credits/) page has links to these. The DefaultIcon will directly work in the Fusion Light theme, and then we run an ImageMagick command to invert its color slightly less than pure white for the Fusion Dark theme. See `scripts/update-dark.sh` for the script I use. I can help you create that once we have decided. We generally use the 32x32 PNG icons from these. A lot of times the name alone cannot be used to locate the icon, and you need to browse the images. In some rare cases we need to create an icon from scratch or take an icon from one of the two collections and try to make it work under the other themes. I just checked and there is nothing obvious to use. In DefaultIcon, maybe `file-add.png` can be used, and in Oxygen I wonder if `insert-image.png` can be used.

src/docks/playlistdock.ui
```diff
@@ -127,6 +134,13 @@ p, li { white-space: pre-wrap; }
           </property>
          </widget>
         </item>
+        <item>
+         <widget class="QPushButton" name="addFilesToPlaylistButton">
+          <property name="text">
+           <string>Add files to playlist</string>
```
ddennedy, line 140: The button should not be added twice - once to each page. Rather, the new icon button should be added to the existing row of icons - perhaps between the minus and checkmark - within `horizontalLayout`.

ddennedy (changes requested):
Thanks for your contributions! Your idea is good and welcome, but some changes are needed. As for the coding style, I do not apply a code formatter, and I do not know if there is a name for it. I just write it like this using Qt Creator with its default options. I would call it "normal" and closely matches the code seen in the Qt docs and examples.

src/docks/playlistdock.cpp
```diff
@@ -1113,3 +1113,34 @@ void PlaylistDock::onProducerModified()
...
+
+        Settings.setOpenPath(QFileInfo(filenames.first()).path());
+
+        foreach(const QString& s, filenames)
+            urls << s;
+
+        mimeData->setUrls(urls);
+
+        onDropped(mimeData, m_view->currentIndex().row() + 1);
+    }
```
ddennedy, line 1145: Personally, I think this `if` block has too many blank lines. I use blank lines to make logical chunks of code within a method. Also, I tend to prefer to always add braces to `for` and `if` statements. It just makes tinkering with the code and extending it easier - at a later time if not now.

src/docks/playlistdock.cpp
```diff
@@ -1113,3 +1113,34 @@ void PlaylistDock::onProducerModified()
...
 }
+
+void PlaylistDock::on_addFilesToPlaylistButton2_clicked()
+{
+    on_addFilesToPlaylistButton_clicked();
+}
+
+void PlaylistDock::on_addFilesToPlaylistButton_clicked()
+{
+    QMimeData* mimeData = new QMimeData;
```
ddennedy, line 1124: This `mimeData` is not destroyed and leaked because `onDropped` does not. It is probably OK to simply declare this as  `QMimeData mimeData;` and then pass it like `onDropped(&mimeData, ...)`

hbtalha: OK, I will make the change.

src/docks/playlistdock.ui
hbtalha: OK

src/docks/playlistdock.ui
hbtalha: > ... the System theme uses one from the Oxygen collection. Our Credits page has links to these.

I went to the Oxygen webpage from the link provided in the Credits but I couldn't use it. I went to search an icon in the search box but the page didn't change.

> See scripts/update-dark.sh for the script I use. I can help you create that once we have decided.

Yes, I would appreciate this.

> In DefaultIcon, maybe MD-stream-video.png can be used

What about `file-add.png`?

src/docks/playlistdock.ui
hbtalha: Yes it is a mistake, I honestly don't know how that changed and I didn't notice it. 

I will revert the file to before the commit.

 May I suggest adding a `ui->stackedWidget->setCurrentIndex(0);` to the constructor (guaranteeing it will always be the default page).

src/docks/playlistdock.ui
hbtalha: Ok I will revert these changes. I am not sure how they were added. I only edited this file using Qt designer, that's why I didn't check it when committing. Are you able to give an idea about what happened so that I avoid it in the future.

I will revert the file to before the commit.

src/docks/playlistdock.ui
ddennedy: You can get Oxygen here: https://github.com/KDE/oxygen-icons

> What about file-add.png?

That is also an option if you do not mind something that looks like paper/document. I prefer something media-oriented.

> I can't seem to use ImageMagick to invert the DefaultIcon icon color.

Your results look fine to me. It can be difficult to see white on white.

src/docks/playlistdock.ui
ddennedy: That is the behavior of Qt Designer whether inside Qt Creator or not. There is no way to avoid it. You just need to review what you are committing.

src/docks/playlistdock.ui
hbtalha: > I prefer something media-oriented.

Ok, I will use `MD-stream-video.png` then.

hbtalha:
@ddennedy 
Preview of the changes

Screenshot_3
Screenshot_6
Screenshot_1

ddennedy (approved):

src/docks/playlistdock.ui
```diff
@@ -170,6 +170,23 @@ p, li { white-space: pre-wrap; }
...
+        </property>
+        <property name="toolTip">
+         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Add files to playlist&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+        </property>
+        <property name="text">
+         <string/>
+        </property>
+        <property name="icon">
+         <iconset theme="list-add-files" resource="../../icons/resources.qrc">
+          <normaloff>:/icons/oxygen/32x32/actions/list-add-files.png</normaloff>:/icons/oxygen/32x32/actions/list-add-files.png</iconset>
```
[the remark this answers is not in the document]
hbtalha: Ok, if you find any issue please let me know!

PR #1120, opened 2021-06-17
new GPS text filter UI

dany123:
[ for https://github.com/mltframework/mlt/pull/703 ]
This is the ui I came up with to fill out all filter properties. 

I took a very C-like procedural approach with this, I'm not familiar either with Qt or javascript so feel free to offer advice/fixes.
There are some todos in the code but none of them are major.
As I said in the backend code, I had issues with just declaring variables at the top and not losing their values after minimising the shotcut window so I use filter.set("x") to reliably store values. This is also why there's some parseInt(Number(x)) there - to guard against NaN and empty values. If they look ugly I'll remove them and clean up but only after understanding/fixing the issue.

I got used to the way it looks, but maybe it's a bit too busy? I'm more of a "give me more options at the expense of nice interface" kind of guy and I'm not sure how to cut down on the options, the filter is also quite complex so it needs quite a bit more fields than usual. The descriptions (on hover) are also quite long, but this is to counter the complexity, I feel like someone with no prior experience should understand everything just by reading the hover tips.

filter p1
filter p2

### DIFF
--- a/src/qml/filters/gpstext/meta.qml
+++ b/src/qml/filters/gpstext/meta.qml
@@ -19,22 +19,22 @@ import QtQuick 2.0
 import org.shotcut.qml 1.0
 
 Metadata {
-	type: Metadata.Filter
-	name: qsTr("GPS text")
-	mlt_service: 'gpstext'
-	qml: 'ui.qml'
-	vui: 'vui.qml'
-	keyframes {
-		allowAnimateIn: true
-		allowAnimateOut: true
-		simpleProperties: ['geometry']
-		parameters: [
-			Parameter {
-				name: qsTr('Position / Size')
-				property: 'geometry'
-				isSimple: true
-				isRectangle: true
-			}
-		]
-	}
+    type: Metadata.Filter
+    name: qsTr("GPS Text")
+    mlt_service: 'gpstext'
+    qml: 'ui.qml'
+    vui: 'vui.qml'
+    keyframes {
+        allowAnimateIn: true
+        allowAnimateOut: true
+        simpleProperties: ['geometry']
+        parameters: [
+            Parameter {
+                name: qsTr('Position / Size')
+                property: 'geometry'
+                isSimple: true
+                isRectangle: true
+            }
+        ]
+    }
 }
--- a/src/qml/filters/gpstext/meta.qml
+++ b/src/qml/filters/gpstext/meta.qml
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2021 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+	type: Metadata.Filter
+	name: qsTr("GPS text")
+	mlt_service: 'gpstext'
+	qml: 'ui.qml'
+	vui: 'vui.qml'
+	keyframes {
+		allowAnimateIn: true
+		allowAnimateOut: true
+		simpleProperties: ['geometry']
+		parameters: [
+			Parameter {
+				name: qsTr('Position / Size')
+				property: 'geometry'
+				isSimple: true
+				isRectangle: true
+			}
+		]
+	}
+}
--- a/src/qml/filters/gpstext/ui.qml
+++ b/src/qml/filters/gpstext/ui.qml
@@ -1,0 +1,815 @@
+/*
+ * Copyright (c) 2014-2021 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.12
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+import QtQuick.Dialogs 1.2
+import QtQuick.Window 2.1
+import Shotcut.Controls 1.0 as Shotcut
+import org.shotcut.qml 1.0 as Shotcut
+import QtQml.Models 2.12
+
+Item {
+	id: gpsTextRoot
+	width: 300
+    height: 800
+	
+	property url settingsOpenPath: 'file:///' + settings.openPath
+	Shotcut.File { id: gpsFile }
+	
+	// This signal is used to workaround context properties not available in
+	// the FileDialog onAccepted signal handler on Qt 5.5.
+	signal fileOpened(string path)
+	onFileOpened: settings.openPath = path
+	
+	Component.onCompleted: {
+		var resource = filter.get('gps.file')
+		gpsFile.url = resource
+		
+		filter.blockSignals = true
+	        
+		filter.set(textFilterUi.middleValue, Qt.rect(0, 0, profile.width, profile.height))
+		filter.set(textFilterUi.startValue, Qt.rect(0, 0, profile.width, profile.height))
+		filter.set(textFilterUi.endValue, Qt.rect(0, 0, profile.width, profile.height))
+		if (filter.isNew) {
+			var presetParams = preset.parameters.slice()
+			var index = presetParams.indexOf('argument')
+			if (index > -1)
+				presetParams.splice(index, 1)
+
+			if (application.OS === 'Windows')
+				filter.set('family', 'Verdana')
+			filter.set('fgcolour', '#ffffffff')
+			filter.set('bgcolour', '#00000000')
+			filter.set('olcolour', '#aa000000')
+			filter.set('outline', 3)
+			filter.set('weight', 10 * Font.Normal)
+			filter.set('style', 'normal')
+			filter.set(textFilterUi.useFontSizeProperty, false)
+			filter.set('size', profile.height)
+			
+			// Add default preset.
+			filter.set(textFilterUi.valignProperty, 'bottom')
+			filter.set(textFilterUi.halignProperty, 'left')
+			filter.set(textFilterUi.rectProperty, '10%/10%:80%x80%')
+			filter.savePreset(presetParams)
+		} else {
+			filter.set(textFilterUi.middleValue, filter.getRect(textFilterUi.rectProperty, filter.animateIn + 1))
+			if (filter.animateIn > 0)
+				filter.set(textFilterUi.startValue, filter.getRect(textFilterUi.rectProperty, 0))
+			if (filter.animateOut > 0)
+				filter.set(textFilterUi.endValue, filter.getRect(textFilterUi.rectProperty, filter.duration - 1))
+		}
+		filter.blockSignals = false
+		setControls()
+		
+		if (filter.isNew)
+			filter.set(textFilterUi.rectProperty, filter.getRect(textFilterUi.rectProperty))
+			
+		if (gpsFile.exists()) {
+			fileLabel.text = gpsFile.fileName
+			fileLabelTip.text = gpsFile.filePath
+		} else {
+			fileLabel.text = qsTr("No File Loaded")
+			fileLabel.color = 'red'
+			fileLabelTip.text = qsTr('No GPS file loaded.\nClick "Open" to load a file.')
+		}
+	}
+
+	FileDialog {
+		id: fileDialog
+		modality: application.dialogModality
+		selectMultiple: false
+		selectFolder: false
+		folder: settingsOpenPath
+		nameFilters: ['Supported files (*.gpx *.tcx)', 'GPS Exchange Format (*.gpx)', 'Training Center XML (*.tcx)']
+		onAccepted: {
+			gpsFile.url = fileDialog.fileUrl
+			gpsTextRoot.fileOpened(gpsFile.path)
+			fileLabel.text = gpsFile.fileName
+			fileLabel.color = activePalette.text
+			fileLabelTip.text = gpsFile.filePath
+			filter.set('gps.file', gpsFile.url)
+			gpsFinishParseTimer.restart()
+			filter.set('gps_start_text', '')
+			filter.set('gps_processing_start_time', 'yyyy-MM-dd hh:mm:ss');
+		}
+	}
+	
+	function smooth_value_to_index(val) {
+		switch (parseInt(val)) {
+			case 0: return 0;
+			case 1: return 1;
+			case 3: return 2;
+			case 5: return 3;
+			case 7: return 4;
+			case 11: return 5;
+			case 15: return 6;
+			case 31: return 7;
+			case 63: return 8;
+			case 127: return 9;
+            default: {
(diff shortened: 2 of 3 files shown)


dany123:
There's one annoying behaviour that I need to fix with javascript, when i filter.set a larger number ending in 0 it will switch to scientific notation so something like -8850800 becomes:
<property name="major_offset">-8.8508e+06</property>

bmatherly:
For status values like video and GPS start times, we tend to put those at the end below a white separator bar. See Compressor and Time Remap filters for example.

Can you explain "Insert GPS Field"? What is that for?

src/qml/filters/gpstext/ui.qml
```diff
@@ -0,0 +1,815 @@
+/*
+ * Copyright (c) 2014-2021 Meltytech, LLC
```
bmatherly, line 2: The date should just be 2021

src/qml/filters/gpstext/ui.qml
```diff
@@ -0,0 +1,815 @@
...
+Item {
+	id: gpsTextRoot
+	width: 300
+    height: 800
+	
+	property url settingsOpenPath: 'file:///' + settings.openPath
+	Shotcut.File { id: gpsFile }
+	
+	// This signal is used to workaround context properties not available in
+	// the FileDialog onAccepted signal handler on Qt 5.5.
```
bmatherly, line 36: I think we do not care about Qt 5.5 since Qt 5.12 is the oldest that is supported.

src/qml/filters/gpstext/ui.qml
```diff
@@ -0,0 +1,815 @@
...
+			if (filter.animateOut > 0)
+				filter.set(textFilterUi.endValue, filter.getRect(textFilterUi.rectProperty, filter.duration - 1))
+		}
+		filter.blockSignals = false
+		setControls()
+		
+		if (filter.isNew)
+			filter.set(textFilterUi.rectProperty, filter.getRect(textFilterUi.rectProperty))
+			
+		if (gpsFile.exists()) {
```
bmatherly, line 84: These should be handled in setControls()

src/qml/filters/gpstext/ui.qml
```diff
@@ -0,0 +1,815 @@
...
+			}
+		}
+	}
+
+    //set all textbox values at start or on refresh
+	function setControls() {
+		textArea.text = filter.get('argument');
+		textFilterUi.setControls();
+		
+		if (filter.isNew) {
```
bmatherly, line 162: All these filter.set() calls should be moved to Component.onCompleted. setControls should not change values on the filter.

src/qml/filters/gpstext/ui.qml
```diff
@@ -0,0 +1,815 @@
...
+		}
+		RowLayout {
+			Shotcut.ComboBox {
+				implicitWidth: 300
+				id: combo_smoothing
+				model: 
+					[   qsTr('0 (raw data)'), 					//0
+				   		qsTr('1 (interpolate missing data)'),	//1
+		                qsTr('3 points'),						//2
+		                qsTr('5 points (recommended)'),			//3
```
bmatherly, line 476: I do not think it is necessary to label something as recommended. But it is good to choose a default value that you think will work for most situations (as you have done).

src/qml/filters/gpstext/ui.qml
```diff
@@ -0,0 +1,815 @@
...
+	    	TextField {
+	    		id: gps_processing_start
+	    		text: 'yyyy-MM-dd hh:mm:ss'
+	            horizontalAlignment: TextInput.AlignRight
+	            //TODO: regex to validate date yyyy-MM-dd hh:mm:ss
+	    		implicitWidth: 128
+				Shotcut.HoverTip { text: qsTr('Insert a date and time formatted as: YYYY-MM-DD HH:MM:SS (all fields mandatory), UTC timezone - same as GPS (use #gps_datetime_now# in filter to get current time).') }
+	        	onEditingFinished: filter.set('gps_processing_start_time', gps_processing_start.text);
+	    	}
+			Shotcut.UndoButton {
```
bmatherly, line 550: There are multiple indentation alignment issues in this file that would be good to fix. Did you mix tabs and spaces?

bmatherly:
In general, I still do not understand major and minor offsets. And why can minor only be set in seconds but major also has seconds?

src/qml/filters/gpstext/ui.qml
```diff
@@ -0,0 +1,815 @@
...
+				}
+			}
+		}
+		
+		Label {
+			id: gps_sync_minor
+			text: qsTr('GPS minor offset')
+			Layout.alignment: Qt.AlignRight
+			leftPadding: 10
+			Shotcut.HoverTip { text: qsTr('This value is also added to gps time to sync with video time: ') }
```
bmatherly, line 420: Why is there a colon at the end of this tooltip? Maybe a typo?

dany123:
> For status values like video and GPS start times, we tend to put those at the end below a white separator bar. See Compressor and Time Remap filters for example.

I wonder if that's the best decision here, putting them at the bottom would place them outside screen (at least on my 1080p display I need to scroll to get to the end) and they are quite important for sync issues. 
 
> Can you explain "Insert GPS Field"? What is that for?

It's the list of available keywords as a drop-down, it's the equivalent "Insert field" button list from _Text: simple_, I had to do a drop down list as there are 14 options

image

src/qml/filters/gpstext/ui.qml
dany123: This is how it was in lut3d filter where I copied it from, I'm still not sure I understand what it's doing, Is it saving the last opened folder location?

src/qml/filters/gpstext/ui.qml
dany123: (I've moved them all except one of them which I'll do it in next commit after some more testing)
Question: when is _filter.isNew_ changing to false? I expected it to only execute once at the very begining of creating a new filter but it's not the case (after a window redraw or something it seems to enter it again).

src/qml/filters/gpstext/ui.qml
dany123: I used notepad++ before and switched to QT creator which seems to dynamically replace tabs with spaces. I changed it to tabs now and fixed most of the lines (there's still one remaining for next commit).

ddennedy: Shotcut source uses spaces while MLT is tabs.

src/qml/filters/gpstext/ui.qml
ddennedy: isNew only changes when reloading the filter UI. It is set once for each time the UI loads and indicates if the filter was added or already exists.

dany123:
> In general, I still do not understand major and minor offsets. And why can minor only be set in seconds but major also has seconds?

I'll try to explain in short: both are in seconds and both are independent of one another, technically you could remove any of them and still sync with no problems (of course, after lifting the 60s limit on the minor one) as in the backend I just add them together and use it as only one value in calculations. If there was an easy way to keep the total value into only one filter property I'd switch to it immediately, but for the UI, the 2 values are tied to the 2 separate ways to enter the ofset. (I assume the main problem is the 2 filter properties that are also sent to backend, not the fact that there are 2 Ui elements dealing with sync, or is it both?)

In the begining there was only one of them, the "major_offset" which is converted from seconds into days/hours/min/sec separately and printed individually so it is very clear how much offset there is. And it is enough for the best case where there the times are synced perfectly or they differ only by the timezone, so if the offset is 3600 seconds it will just show 00 01 00 00 and that's it. 

But, most of the time this is not the case, unless the device has internet access and auto time sync it will probably be manually set to up to 1 min accuracy (as an example a gopro will only let you set HH:MM, not seconds, so it is almost guaranteed to have some error). So what do you do? you start inputting 1 second into the second textArea, you check to see if it's better on the overlay, maybe try 5, 10, 15, etc, but this is not efficient as you have to click into the box, replace the old number then click outside (or hit tab) to finish editing and have the text update on overlay.
This is where the slider spinner of the "minor_offset" comes into action, as if you use the mouse scrollwheel on it (or just click and drag), you can immediately go through 2 minutes of gps data and have it continously update it over the video so you can quickly check where the speed/altitude or something else matches the video at that point (so speed 0 if you stopped, or max altitude if you are at the peak of some mountain/track).

In practice I almost always used the minor offset slider instead of entering data manually into the fields. This is why I'm defending this separate minor/major split.

src/qml/filters/gpstext/ui.qml
dany123: Oh, so QT creator was right, I'll switch to spaces in the next commit.

bmatherly:
>> For status values like video and GPS start times, we tend to put those at the end below a white separator bar. See Compressor and Time Remap filters for example.

> I wonder if that's the best decision here, putting them at the bottom would place them outside screen (at least on my 1080p display I need to scroll to get to the end) and they are quite important for sync issues.

There are tradeoffs in all UI decisions. I tent to prefer consistency in UI unless there is an exceptional reason to be inconsistent. If you move the status to the end under a white bar like the other filters, would it help to also display those statuses in a tooltip when you mouse over the file name? Maybe that would achieve consistency and convenience.

> It's the list of available keywords as a drop-down, it's the equivalent "Insert field" button list from Text: simple, I had to do a drop down list as there are 14 options

It is not consistent with the dynamic text filter, but I understand the exceptional need here. Obviously we will not put 14 buttons. I kind of prefer the approach. Maybe we will change the dynamic text filter to match this someday.

> In practice I almost always used the minor offset slider instead of entering data manually into the fields.

You should be able to use the wheel mouse to quickly "spin" or "step" the seconds value in the major offset. If the only reason to have two offsets is so that you can use two different controls, then let's find a control that can achieve both uses reasonably well.

The Timer filter uses the ClockSpinner.qml control. You could copy/paste that and modify it, or share the control and extend it to meet your needs. If you were to share it, you would need to:
1) Move the filter to the modules/Shotcut/Controls folder
2) Add an option to display/control days (needed for gps text)
3) Add an option to display/control miliseconds (needed for Timer)
4) Add a scroll wheel handler to increment seconds

src/qml/filters/gpstext/ui.qml
bmatherly: OK. You can probably just delete the comment. The comment is about the signal.

You are asking about the following line which saves the opened filed name in the Shotcut settings. It does this so that the next time the users tries to open a file Shotcut will start the file dialog in the same directory.

dany123:
> There are tradeoffs in all UI decisions. I tent to prefer consistency in UI unless there is an exceptional reason to be inconsistent. If you move the status to the end under a white bar like the other filters, would it help to also display those statuses in a tooltip when you mouse over the file name? Maybe that would achieve consistency and convenience.

That would work for me, seems like a good compromise.

> You should be able to use the wheel mouse to quickly "spin" or "step" the seconds value in the major offset. If the only reason to have two offsets is so that you can use two different controls, then let's find a control that can achieve both uses reasonably well.
> 
> The Timer filter uses the ClockSpinner.qml control. You could copy/paste that and modify it, or share the control and extend it to meet your needs. If you were to share it, you would need to:
> 
>     1. Move the filter to the modules/Shotcut/Controls folder
> 
>     2. Add an option to display/control days (needed for gps text)
> 
>     3. Add an option to display/control miliseconds (needed for Timer)
> 
>     4. Add a scroll wheel handler to increment seconds

Good idea, combining the controls should keep the functional part of both. 

I am not a fan of the ClockSpinner in general, I mean it's ok for something standard like HH:MM:SS to read, but for editing if I need to change something in the middle (say 01:00:45 to 01:23:45) I find it faster to delete everything and just type the entire string at once.
Instead, I like the DoubleSpinBox functionality and design so I'll give this a shot to see if it can do all the stuff + if it fits the design.

Is there a simple onWheelUp/Down event for the simple textArea I can fallback to if the spinbox doesn't work as expected?

bmatherly:
> But I also have some issues with the limited acces to wheel events so I went to my backup plan to add mouse wheel functionality to the original textareas and after some searching I found something that works and is very flexible (I mean I can do the full 59 seconds +1 logic that increases correctly the next field). Would this solution work for you?

It sounds promising and you are on the right track. We would want all the numeric fields to behave the same. So either duplicate the mouse event for each field, or make this into a new component.

dany123:
I've cleaned up the design quite a bit after applying the suggestions, only 3 rows at the top for the important stuff before the text comes in.
Should I keep the bold category labels? (GPS options, Text options, Advanced options - after correcting case)
I feel like there is a bit more spacing between my rows compared to other filters, any idea why? The "Text" near the big textbox is also misaligned for no reason.
There is a lot of space (margin/padding?) around the TextFilterUi, not sure if related to previous spacing issue, can I remove it?
Should I bother aligning items? (to allign the GPS offset and smoothing rows there's one element with 39 px width, it was 40 originally but the slight 1 px misalignment was quite annoying to look at)

Also, are multiple simple commits or single complex ones prefferred? 

2021-06-21 gpstext design v2

bmatherly:
> Should I keep the bold category labels? (GPS options, Text options, Advanced options - after correcting case)

That is up to you. There are other filters that make similar groupings. If you keep the category labels, maybe "Advanced" should be "Speed"?

> I feel like there is a bit more spacing between my rows compared to other filters, any idea why?

It looks like you have a GridLayout inside a GridLayout. Did you do that on purpose? I think it should not be necessary. My recommendation is to group each row of components together in the code (with an empty line between each group) and then look very carefully at the use of columnSpan, etc.

For the padding around the TextArea, you might need to play with the padding and anchors. Keep in mind that the code you copied might be trying to reserve some space for scrollbars to appear if/when they are needed.

> Also, are multiple simple commits or single complex ones prefferred?

You can keep adding commits to this review and we will squash them all before merging to master.

src/qml/filters/gpstext/ui.qml
dany123: Ok, I tried moving stuff around but had to revert to original version and just removed the comment now.

src/qml/filters/gpstext/ui.qml
dany123: Done in latest commit.

src/qml/filters/gpstext/ui.qml
dany123: Separated everything properly in latest commit.

src/qml/filters/gpstext/ui.qml
dany123: Ok, it makes sense, I had a console.log() in a different area (independent of component.onCompleted) that gave me the wrong impression.

dany123:
> If you keep the category labels, maybe "Advanced" should be "Speed"?

Speed it is, I had some extra fields I planned to put there not related to speed but I scratched them for now, I might switch back to Advanced options if I ever add them to backend.

>It looks like you have a GridLayout inside a GridLayout. Did you do that on purpose? I think it should not be necessary

At one point I tried aligning all buttons to the right using relative positioning, didn't work out and the grid was leftover. Changed in latest commit to a single rowlayout for that entire row.
I found out why the weird large spacings appeared: there was too much space in the main element (height: 800) so items spread around to fill it, lowered it to 700 and added a fillHeight element at the end and now it's good.

I also completely removed minor offset (will update backend soon). I'm not sure if it's worth to rename all "major_offset" references to something generic now.

bmatherly:
> I'm not sure if it's worth to rename all "major_offset" references to something generic now.

Maybe "time_offset" or just "offset"?

If we want to change it, it should be done now because it is difficult to change later after applications have started to depend on it.

dany123:
Good point, I'll better change it now rather than later.

src/qml/filters/gpstext/meta.qml
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+	type: Metadata.Filter
+	name: qsTr("GPS text")
```
austincbrooks, line 23: Just an observation... Every other filter in Shotcut uses title case, like "GPS Text".  This filter would stick out if it didn't.

dany123: Good catch, renamed.

dany123:
Is there an equivalent of `producer.duration / profile.fps` but for the original file duration?

bmatherly:
> Is there an equivalent of producer.duration / profile.fps but for the original file duration?

Maybe try producer.length / profile.fps
https://github.com/mltframework/shotcut/blob/master/src/qmltypes/qmlproducer.h#L36
This is a recent addition, so you may need to checkout the latest master and rebase

> Also I just noticed my tooltips are inconsistent regarding ending with a dot, should I add it to all sentences or remove it everywhere?

Prefer complete sentences with proper capitalization and punctuation.

dany123:
> Maybe try producer.length / profile.fps

This worked perfectly.

PR #1121, opened 2021-06-18
Deband filter - initial QML UI files

austincbrooks:
Documentation is up-to-date and located at:
https://forum.shotcut.org/t/introducing-a-new-filter-deband/28302


ddennedy:
Thanks, initial review briefly reading over the code looks good! We will schedule this for version 21.07.

src/qml/filters/deband/ui.qml
```diff
@@ -0,0 +1,440 @@
...
+            onValueChanged: setThreshold(thr1Param, value)
+        }
+        Shotcut.UndoButton {
+            onClicked: idThr1.value = thrToPct(thr1Default)
+        }
+
+        // Row split
+
+        Label {
+            text: qsTr('Cb/G Threshold')
```
bmatherly, line 244: Would be nice to get some tool tips on some of these. Will normal users know what "Cb/G" means?

src/qml/filters/deband/ui.qml
```diff
@@ -0,0 +1,440 @@
...
+
+        // Row split
+
+        Label {}
+        CheckBox {
+            id: idLink
+            text: qsTr('Link thresholds')
+            onClicked: filter.set(linkParam, checked)
+        }
+        Label {}
```
bmatherly, line 303: You can either use an Item{} here or set columnspan: 2 on idLink
https://doc.qt.io/qt-5/qml-qtquick-layouts-layout.html#columnSpan-attached-prop

src/qml/filters/deband/ui.qml
```diff
@@ -0,0 +1,440 @@
...
+
+        // Filler
+
+        Label {
+            Layout.columnSpan: 3
```
bmatherly, line 308: Are the fillers necessary? This could cause more people to require vertical scrolling depending on their screen resolution and panel layout. Maybe consider removing this filler and the one below.

src/qml/filters/deband/ui.qml
```diff
@@ -0,0 +1,440 @@
...
+        }
+
+        // Row split
+
+        Label {
+            text: qsTr('More information') + ': <a href="http://ffmpeg.org/ffmpeg-all.html#deband">FFmpeg deband</a>'
```
bmatherly, line 419: I appreciate what you are trying to do here. But we do not do this for other filters and I am not thrilled that it exposes some basic users to advanced things they may not understand. Can we extract the most important parts of the ffmpeg documentation and incorporate it into the tooltips? I could even support some paragraph text here if it would avoid the link to ffmpeg. I am open minded on this and mostly looking for brainstorm/discussion.

bmatherly (commented):
I agree with Dan. Let's target 21.07 for this. Should be a good addition.

bmatherly:
It would also be good to squash all these commits into one commit  before we merge into master.

src/qml/filters/deband/ui.qml
austincbrooks: The idea here was to prevent the checkbox text from spilling into the Undo column, which could create a ragged-edge appearance with long translated strings.  If Item{} is a suitable replacement, should I also use Item{} in place of the Label{} that's in the first column of the checkbox rows?

src/qml/filters/deband/ui.qml
austincbrooks: Agreed on vertical scrolling.  However, the reasoning was that the Randomize checkboxes were directly associated with the sliders above them.  If I put two more checkboxes below the lower Randomize checkbox with no filler, it could seem implied that those two checkboxes (Blur and All) were associated with the Slider as well.  But they're not.  They're independent of the sliders, and the filler was a way to imply disassociation.  I could see collapsing the upper filler.  But what are your thoughts on the middle filler between Randomize and Blur?

src/qml/filters/deband/ui.qml
austincbrooks: I'm fine with removing the entire block.  There is plenty of documentation on the forum that should be more accessible to beginning users.

austincbrooks:
My dev box is being rebuilt and I'm doing everything through the GitHub web site at the moment.  I'm not sure how to squash through it even after much Google searching.  Do you happen to know?

src/qml/filters/deband/ui.qml
bmatherly: Item is typically used if you just need a filler in a cell. If there is some other reason to use Label, that is fine, too.

src/qml/filters/deband/ui.qml
bmatherly: My advice would be to word the labels in a way that the text creates the association you are looking for. For example, "Randomize between zero and value" could be changed to "Randomize pixel range between zero and value". And the next one could be "Randomize direction between zero degrees and value".

For complex filters like this one, people will need to learn the settings through study and trial-and-error. So I don't know how much the grouping really helps. Hopefully people will suggest improvements to the labels as they start to understand the filter.

src/qml/filters/deband/ui.qml
bmatherly: Agreed. It would be great if you could also contribute a Documentation page on the forum for this filter.

bmatherly:
> I'm not sure how to squash through it even after much Google searching. Do you happen to know?

We can squash and merge in the GitHub UI. Or, I am willing to check out the branch and take care of it when you are done with the polish.

ddennedy:
I will squash upon merge.

src/qml/filters/deband/ui.qml
austincbrooks: I agree and thought a lot about this.  I decided to try making the labels friendly to novices, then put some techie stuff in the tool tips.
My theory is that most people wouldn't use a Deband filter in RGB mode anyway (no range compression from RGB sources), which frees the labels to focus on YCbCr mode.
I'm stretching "luma" to mean "contrast" in this context, but functionally it works.  The higher the contrast in the luma plane, the higher the "contrast" threshold must be to count as banding.
The other labels are hopefully self-explanatory.

src/qml/filters/deband/ui.qml
austincbrooks: Item {} caused layout issues due to collapsing to zero width.
I had to use Item { Layout.fillWidth: true } which hopefully is okay.
I replaced all occurrences of Label {} with Item.

src/qml/filters/deband/ui.qml
austincbrooks: Done.

src/qml/filters/deband/ui.qml
austincbrooks: Documentation exists in the forum at https://forum.shotcut.org/t/introducing-a-new-filter-deband/28302

I will update references to labels based on the outcome of this pull request.  Then it can be copied and pasted to a #docs wiki page.

austincbrooks:
All issues to my knowledge have been addressed.  Ready for another review round.

src/qml/filters/deband/ui.qml
```diff
@@ -0,0 +1,430 @@
...
+            id: idPreset
+            Layout.columnSpan: 2
+            parameters: allParams
+            onPresetSelected: setControls()
+        }
+
+        // Row split
+
+        Label {
+            text: qsTr('Contrast Threshold')
```
ddennedy, line 230: Per our [UI Conventions](https://shotcut.org/notes/ui-conventions/) field labels should not capitalize each word. I realize there is some inconsistency here, but I want to avoid it continuing, and this is a simple fix. The reason existing ones have not been corrected when I noticed is because I did not want to break translations for something superficial.

ddennedy (changes requested):

austincbrooks:
Easy enough.  Thanks for the clarification.

ddennedy (approved):

PR #1125, opened 2021-06-29
Ensure length is in time format (not frame)

### LINKED ISSUE #1124: Wrong conversion while fps change
All latest Shotcut versions.

Original project was made for FullHD 30 fps. Now I'm going to use 60 fps.
The problem is that some video files length is stored in number of frames. In our case 20 seconds 30 fps video has 600 frames.
If I'll try to change output video format, project becomes broken.  All video with length in frames will be twice shorter.
If same video file will be added after change to 60 fps it stores differently in project.
I've tried different versions of Shotcut, tried re-saves and other tweaks. Now the only way to edit xml manually, but project is big and that will take a lot of time...

[Test.zip](https://github.com/mltframework/shotcut/files/6730185/Test.zip)

bmatherly:
When dragging multiple files to the playlist, the first clip would
have length in time format, but subsequent clips would be in frame
format.

Fix #1124

### DIFF
--- a/src/mltcontroller.cpp
+++ b/src/mltcontroller.cpp
@@ -903,6 +903,7 @@ Producer* Controller::setupNewProducer(Producer* newProducer) const
 {
     // Call this function before adding a new producer to Shotcut so that
     // It will be configured correctly. The returned producer must be deleted.
+    Producer* ret = nullptr;
     QString serviceName = newProducer->get("mlt_service");
     if (serviceName == "avformat")
     {
@@ -911,7 +912,7 @@ Producer* Controller::setupNewProducer(Producer* newProducer) const
     }
     setImageDurationFromDefault(newProducer);
     lockCreationTime(newProducer);
-    newProducer->get_length_time(mlt_time_clock);
+    QString length = newProducer->get_length_time(mlt_time_clock);
 
     if (serviceName.startsWith("avformat"))
     {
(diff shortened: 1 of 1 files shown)


src/mltcontroller.cpp
```diff
@@ -934,10 +935,14 @@ Producer* Controller::setupNewProducer(Producer* newProducer) const
                 }
                 filter.reset(newProducer->filter(i));
             }
-            return chain;
+            ret = chain;
```
ddennedy, line 938: This changed caused a severe regression:
1. Open a clip from File > Open or Recent
2. Add a filter
3. Upon choosing a filter, Shotcut hangs

Hmm, I just experienced strong deja vu!

src/mltcontroller.cpp
```diff
@@ -934,10 +935,14 @@ Producer* Controller::setupNewProducer(Producer* newProducer) const
...
         }
     }
-    return new Mlt::Producer(newProducer);
+    if (ret) {
+        ret = new Mlt::Producer(newProducer);
+    }
+    ret->set("length", qUtf8Printable(length));
```
ddennedy, line 944: This causes a crash. If `ret` is NULL, it is not replaced with a new one, but the `ret->set()` crashes on NULL dereference.

src/mltcontroller.cpp
ddennedy: fixed in fe3f8355

src/mltcontroller.cpp
ddennedy: fixed in fe3f8355

PR #1128, opened 2021-07-06
Warn the user before exporting a frame from proxy file

bmatherly:
As suggested here:
https://forum.shotcut.org/t/add-warning-when-exporting-frame-with-proxy-on/28185

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -4112,6 +4112,20 @@ void MainWindow::onGLWidgetImageReady()
     if (Settings.playerGPU() || Settings.playerPreviewScale()) {
         MLT.setPreviewScale(Settings.playerPreviewScale());
     }
+    if (glw->imageIsProxy() && !image.isNull()) {
+        QMessageBox dialog(QMessageBox::Question,
+                           tr("Export frame from proxy?"),
+                           tr("This frame is from a lower resolution proxy instead of the original source.\n\n"
+                              "Do you still want to continue?"),
+                           QMessageBox::No | QMessageBox::Yes,
+                           this);
+        dialog.setDefaultButton(QMessageBox::Yes);
+        dialog.setEscapeButton(QMessageBox::No);
+        dialog.setWindowModality(QmlApplication::dialogModality());
+        if (dialog.exec() != QMessageBox::Yes) {
+            return;
+        }
+    }
     if (!image.isNull()) {
         SaveImageDialog dialog(this, tr("Export Frame"), image);
         dialog.exec();
--- a/src/glwidget.cpp
+++ b/src/glwidget.cpp
@@ -761,6 +761,21 @@ QImage GLWidget::image() const
     return QImage();
 }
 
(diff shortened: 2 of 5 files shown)


src/mainwindow.cpp
```diff
@@ -4096,6 +4096,21 @@ void MainWindow::on_actionExportEDL_triggered()
 void MainWindow::on_actionExportFrame_triggered()
 {
     if (!MLT.producer() || !MLT.producer()->is_valid()) return;
+    if (Settings.proxyEnabled()) {
```
ddennedy, line 4099: It is not ideal to check the setting because not all clips in a project have a proxy when proxy mode is on. Instead, check `producer.get_int(kIsProxyProperty)`. However, you might have to try to get the producer from the current shared frame; `MLT.producer()` is not the correct thing to check.

src/mainwindow.cpp
```diff
@@ -4096,6 +4096,21 @@ void MainWindow::on_actionExportEDL_triggered()
...
+        QMessageBox dialog(QMessageBox::Question,
+                           tr("Export from from proxy"),
```
ddennedy, line 4101: "from" is repeated.

ddennedy (changes requested):

src/mainwindow.cpp
bmatherly: Good idea. I will try that.

bmatherly:
Do you have an opinion about whether this should be a QuestionBox or a WarningBox?

ddennedy:
> Do you have an opinion about whether this should be a QuestionBox or a WarningBox?

I think either one can work.

src/mainwindow.cpp
```diff
@@ -4096,6 +4096,21 @@ void MainWindow::on_actionExportEDL_triggered()
...
+                           tr("This project is currently using proxy files. The exported frame will be taken from the lower resolution proxy instead of the original source.\n\n"
+                              "Do you still want to continue?"),
+                           QMessageBox::No | QMessageBox::Yes,
+                           this);
+        dialog.setDefaultButton(QMessageBox::Yes);
+        dialog.setEscapeButton(QMessageBox::No);
+        dialog.setWindowModality(QmlApplication::dialogModality());
+        int result = dialog.exec();
+        if (result == QMessageBox::No) {
```
ddennedy, line 4110: By checking for No, please be aware that closing the dialog via window title button will not return early here, and that could be unexpected.

src/mainwindow.cpp
bmatherly: That was pretty slick. Now it only prompts if the frame is actually from a proxy file. Thanks for the suggestion.

src/mainwindow.cpp
bmatherly: Done.

src/mainwindow.cpp
bmatherly: Fixed

ddennedy (approved):

PR #1130, opened 2021-07-08
Add hover tip to Speed parameter of Timer filter

emlun:
Parameter description to go with https://github.com/mltframework/mlt/pull/723 . See also: https://github.com/mltframework/shotcut/pull/1080#issuecomment-868833814


bmatherly:
LGTM

PR #1137, opened 2021-07-24
Fix clip trim in redo on timeline

bmatherly:
As reported here:
https://forum.shotcut.org/t/undo-trim-clip-in-may-shift-clips-bug-hasnt-been-completely-fixed/28487


bmatherly:
I will admit that I don't completely understand the change but I copied the logic from here:
https://github.com/mltframework/shotcut/blob/master/src/docks/timelinedock.cpp#L1032

PR #1141, opened 2021-07-30
use filesystem instead of sqlite for thumbnails

ddennedy:
NOTE: You might want to make a backup copy of your `db.sqlite3` in the Shotcut App Data directory before running this. It deletes the database after it converts it all to files the first time the thumbnails are used.

Check this out to remove using a database for storing the thumbnail and waveform cache. The database cannot be opened by multiple instances of Shotcut - certainly not on Windows. It hope it improves the performance of running multiple instances of Shotcut and fixes issues where waveforms are not updating.

### DIFF
--- a/src/database.cpp
+++ b/src/database.cpp
@@ -36,6 +36,7 @@ Database::Database(QObject *parent) : QObject(parent)
     m_deleteTimer.setInterval(kDeleteThumbnailsTimeoutMs);
     m_deleteTimer.setSingleShot(true);
     connect(&m_deleteTimer, &QTimer::timeout, this, &Database::deleteOldThumbnails);
+    thumbnailsDir(); // convert from db to filesystem if needed
 }
 
 Database &Database::singleton(QObject *parent)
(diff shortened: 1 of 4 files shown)


bmatherly:
Looks good to me.

In testing, the conversion worked properly for me and new waveform thumbnails were created successfully.

I am not sure how to measure performance improvement. Do you have any ideas?

I notice that clips without audio result in a 91kB file that is 1x1 px. I suppose that is on purpose to signal that the file has already been checked?

ddennedy:
To be honest, I have trouble reproducing the performance issues reliably. I do know that you can see database errors in the log - at least on Windows - so it is unable to reliably read and write the cache, which is obviously not good. I put this into a branch to test more easily on other machines before deciding to merge. So, I thought I'd make a PR to have you review it.

> I notice that clips without audio result in a 91kB file that is 1x1 px

That is not intentional but a byproduct of what happens when requesting waveform. I do not think it is a new behavior. We can probably prevent this by making the waveform task do nothing when audio_index = -1.

PR #1156, opened 2021-09-13
Ctl s w

bmatherly:
Improvements for:
https://forum.shotcut.org/t/entire-project-wiped-after-accidentally-hitting-ctrl-w-and-ctrl-s-at-the-same-time/29786

Now, when CTL + WS occurs, the project still closes, but the "save as" dialog pops up instead of saving an empty project into the previous project file.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -188,6 +188,7 @@ MainWindow::MainWindow()
     connect(ui->actionAbout_Qt, SIGNAL(triggered()), qApp, SLOT(aboutQt()));
     connect(this, &MainWindow::producerOpened, this, &MainWindow::onProducerOpened);
     connect(ui->mainToolBar, SIGNAL(visibilityChanged(bool)), SLOT(onToolbarVisibilityChanged(bool)));
+    ui->actionSave->setEnabled(false);
 
     // Accept drag-n-drop of files.
     this->setAcceptDrops(true);
@@ -2687,6 +2688,7 @@ void MainWindow::onProducerOpened(bool withReopen)
         Util::getHash(*MLT.producer());
         ui->actionPaste->setEnabled(true);
     }
+    ui->actionSave->setEnabled(true);
     QMutexLocker locker(&m_autosaveMutex);
     if (m_autosaveFile)
         setCurrentFile(m_autosaveFile->managedFileName());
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -2687,6 +2688,7 @@ void MainWindow::onProducerOpened(bool withReopen)
         Util::getHash(*MLT.producer());
         ui->actionPaste->setEnabled(true);
     }
+    ui->actionSave->setEnabled(true);
```
ddennedy, line 2691: Maybe we should change the save action's enabled state based on when the `QMainWindow::setWindowModified()` instead of this one case. There are 12 lines where that is set in `mainwindow.cpp`. However, the window-modified state is not super critical at the moment. It provides a window decoration to indicate and affects `continueModified()`. I would hate to put the tool into a situation where it will not save. What do you think?

ddennedy:
I tested this, and it fixes the reported problem. But also it still lets me save the "hidden" producer upon close when pressing these shortcuts in quick succession. It would be nice to prevent letting one save that.
I thought that is related to the save action still being enabled. But setting the save action enabled state every where `setWindowModified()` is called was not enough. Even a call to disable the save action early in `on_actionClose_triggered()` was not enough! It looks like Qt queues these actions in its `QEventLoop`. As well, checking for the "_hide" property at the beginning of `on_actionSave_triggered()` does not prevent it. It seems much of this is due to the asynchronous things that occur when making this dummy producer upon close instead of actually clearing the producer, which is quite a bit more work. So, I will accept this as avoiding the problem without introducing further risk. After all, I do want to make a release soon to fix a regression - and probably without a beta or release candidate.

src/mainwindow.cpp
bmatherly: > I would hate to put the tool into a situation where it will not save

This is my greatest fear and the reason I put that change in a separate commit. Also, the "action enable" change is not material for the fix (as you correctly observe below). I notice that some applications do not allow saving an empty file/project while others do. So I am not sure how important it is to disable the action. If you like the idea, I am willing to put more time into making a better implementation that is more comprehensive. But maybe wait until there will be a beta release.

bmatherly:
> Qt queues these actions in its QEventLoop

That is a good theory. In the future, we could look into intercepting events in the loop and detect situations where S and W end up in the queue at the same time.

PR #1166, opened 2021-10-16
WIP Markers

bmatherly:
Instructions:

* Press <kbd>M</kbd> to create a marker on the timeline at the current playhead position
* Click and drag to move markers
* <kbd>Ctrl</kbd>+click and drag to change the range of a marker (default is 1 frame)
* Right-click a marker to edit

TODO:

- [ ] Display marker information in the bubble status (in progress)
- [x] Click to seek to it
- [x] Edit keyboard shortcut (<kbd>M</kbd>)
- [x] Toolbar button
- [x] Delete keyboard shortcut (<kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>M</kbd>)
- [x] Export marker range

Out-of-scope but desired:

* Snap or skip to markers
* Create chapters from markers
* Ripple with timeline operations
* Display all markers in a **Markers** panel

Comments welcome on size, location, usability. I personally wonder if the marker handles are too small. Maybe the ruler height should be increased to allow the markers to be bigger.

### DIFF
--- a/src/commands/markercommands.cpp
+++ b/src/commands/markercommands.cpp
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2021 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "markercommands.h"
+#include <Logger.h>
+
+namespace Markers
+{
+
+DeleteCommand::DeleteCommand(MarkersModel& model, const Marker& delMarker, int index)
+    : QUndoCommand(0)
+    , m_model(model)
+    , m_delMarker(delMarker)
+    , m_index(index)
+{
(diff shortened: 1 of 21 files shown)


ddennedy:
I edited the description to prioritize the TODO list against the design doc. I will not include the out-of-scope items in the first release but "Create (YouTube) chapters" should be fairly easy using the JavaScript-based framework I made for exporting EDL.

bmatherly:
> I edited the description to prioritize the TODO list against the design doc.

Thanks.

One problem I have not solved for the keyboard shortcut is that currently I do not have a concept of a "selected" marker. I could infer a selected marker by looking at what marker is under the playhead - and ignoring a command if there is no marker under a playhead. However, I also do not currently prohibit overlapping markers.

For keyboard shortcuts, I propose to apply the keyboard shortcut command to the first marker that aligns with the playhead. Thoughts?

In addition to YouTube chapters, there was also a request to create file chapters for containers that support chapters. FFMpeg has an interface to specify chapters - and I think we could map markers over to the FFMpeg chapter metadata to support that eventually.

ddennedy:
I will start working on the toolbar button.

ddennedy:
Regarding the selected marker for edit and redundant markers. At first I was thinking that you would be prohibited from creating redundant markers because the marker command would automatically choose new or edit depending on if it exists. However, edit and dragging still lets you have more than one marker with the same start time. I think initially we should just make the command edit the first marker if finds. Later, we can add a popup menu listing them for choice.

bmatherly:
>I think initially we should just make the command edit the first marker if finds. Later, we can add a popup menu listing them for choice.

Agreed. I think overlapping will be uncommon.

ddennedy:
See #1167 where I fixed and added a number of things including export! I am willing to merge this as-is for a beta. You can try to add the bubble help by end of Oct. 17, but it can also be added during beta IMO. However, I would like you to review more of my changes on the origin branch before I merge by EOD Oct. 17.

PR #1173, opened 2021-10-29
Marker changes

ddennedy:
This is a proposed fix for the color rendering problem plus something else little I noticed with how the color is serialized (some large negative integer).

### DIFF
--- a/scripts/build-shotcut-msys2.sh
+++ b/scripts/build-shotcut-msys2.sh
@@ -1069,10 +1069,10 @@ function deploy
 
   log Copying some libs from Qt
   if [ "$DEBUG_BUILD" != "1" -o "$SDK" = "1" ]; then
-    cmd cp -p "$QTDIR"/bin/Qt5{Core,Gui,Multimedia,Network,OpenGL,Qml,QmlModels,QmlWorkerScript,Quick,QuickControls2,QuickShapes,QuickTemplates2,QuickWidgets,Sql,Svg,WebSockets,Widgets,WinExtras,Xml}.dll .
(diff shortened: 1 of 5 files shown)


ddennedy:
> But this should be in the API:

That is not so clear. The framework does not really care about the string representation of a color, and things work without the special string format. That is more the concern of XML or user interfaces. In this case, there can be people wanting to parse markers MLT XML from Shotcut in order to get things like chapters. The framework does have a parser for requesting an integer or color from a string that begins with '#'. I suppose we can change `mlt_properties_set_color` to use `mlt_property_set_string` instead, but not yet. I want to think about it more.

ddennedy:
>  I had avoided that because of all the angry warnings in the QT documentation about performance.

Another option is to use `Canvas` in QML, but some time a while ago Harald and I converted some of these to QQuickPaintedItems.

ddennedy:
Re Canvas, perhaps not, from the Qt docs:
> Note: In general large canvases, frequent updates, and animation should be avoided with the Canvas.Image render target. This is because with accelerated graphics APIs each update will lead to a texture upload. Also, if possible, prefer QQuickPaintedItem and implement drawing in C++ via QPainter instead of the more expensive and likely less performing JavaScript and Context2D approach.

bmatherly:
> I suppose we can change mlt_properties_set_color to use mlt_property_set_string instead, but not yet. I want to think about it more.

Or we could add a new function to set colors in string format - such as mlt_properties_set_color_str(). Personally, I think that set_color() should always set a string in # format. But I do not feel strongly about it and this is not a comment meant to block the review.

Regarding the QQuickPaintedItem implementation: I approve of what you propose. As a simplification, I think there could be just one Marker object with a property for left or right (or start/end, I suppose). The painterpath could be drawn appropriately based on the left/right property. But this is also just a comment and not meant to block the review.

bmatherly (approved):

bmatherly:
I just noticed on my Linux Machine that this change causes black triangles in the corner of the marker. Does this also happen for you?

image

ddennedy:
Oh yeah that is much easier to see on a light theme. The fix is really easy: just need to remove the `setOpaquePainting(true);`

bmatherly (approved):
Looks good!

PR #1181, opened 2021-11-10
Add StartupWMClass line to desktop file

silentJET85:
In some desktop environments, if someone pins Shotcut to their panel or dock, they will get a duplicate icon when they launch it. 

This is because the launcher icon is associated only with the initial shell script. When the Shotcut binary is called, a second icon shows up on the user's task-switcher or dock.

Adding this line to the .desktop file prevents this.


ddennedy:
thanks

PR #1182, opened 2021-11-11
Improve accuracy of delta portion of bubblehelp while trimming on Timeline and Keyframes

traprog:
timeline/Track.qml
- `else` clause for `if (delta != 0)` means the trim handle is being stuck due to snapping enabled. Accumulated delta `clip.originalX` needs to be adjusted in the same manner as when `timeline.trimClipIn` or `timeline.trimClipOut` returns `false`.

keyframes/keyframes.qml
- `else` clauses in `onTrimmingIn` and `onTrimmingOut` are equivalent to when `timeline.trimClipIn` or `timeline.trimClipOut` returns `false` in the Timeline.

Clip.qml
- Somehow `- originalX - delta` is double-counting or something. `delta` accumulation stops half way trimming. I feel like this `if` statement check of a clip becoming very thin is taken care of by `clip.originalX -= originalDelta` or `clip.originalX -= delta` in the files above.

### DIFF
--- a/src/qml/views/timeline/Track.qml
+++ b/src/qml/views/timeline/Track.qml
@@ -188,7 +188,7 @@ Rectangle {
                     } else {
                         clip.originalX -= originalDelta
                     }
-                } else
+                } else {
                     clip.originalX -= originalDelta
                 }
             }
--- a/src/qml/views/timeline/Track.qml
+++ b/src/qml/views/timeline/Track.qml
@@ -160,6 +160,8 @@ Rectangle {
                     } else {
                         clip.originalX -= originalDelta
                     }
+                } else {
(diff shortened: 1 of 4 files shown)


src/qml/views/timeline/Track.qml
```diff
@@ -186,6 +188,8 @@ Rectangle {
                     } else {
                         clip.originalX -= originalDelta
                     }
+                } else
```
ddennedy, line 191: Missing a '{' at the end of this line

traprog: Yeah, updated with Update Track.qml below.

ddennedy:
Thank you :pray:

PR #1188, opened 2021-11-22
Fix erratic keyframe parameter height zoom function

traprog:
It looks like a parameter's `minimum` and `maximum` properties are constantly changing as I move a keyframe while `zoomHeight` is `true`.
I made `minimum` and `maximum` property modification one-time event in response to a click on the zoom icon.

`zoomHeight` can be made to be a property of a parameter rather than of a parameter head as it was in the original.

### DIFF
--- a/src/qml/views/keyframes/Parameter.qml
+++ b/src/qml/views/keyframes/Parameter.qml
@@ -30,6 +30,11 @@ Item {
 
     signal clicked(var keyframe, var parameter)
 
+    function setLowestHighest(lh) {
+        minimum = lh ? model.lowest : model.minimum
+        maximum = lh ? model.highest : model.maximum
+    }
+
     function getKeyframeCount() {
         return keyframesRepeater.count
     }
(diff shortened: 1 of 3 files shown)


src/qml/views/keyframes/Parameter.qml
```diff
@@ -30,6 +30,11 @@ Item {
 
     signal clicked(var keyframe, var parameter)
 
+    function setLowestHighest(lh) {
```
bmatherly, line 33: Name suggestions:
Change "setLowestHighest" to "setMinMax"
Change "lh" to "zoomHeight"
thoughts?

bmatherly (changes requested):
Thanks for the contribution. One comment for your consideration.

src/qml/views/keyframes/Parameter.qml
traprog: Yeah, I have no problem with changing names.
Updated with two commits below.

PR #1191, opened 2021-11-28
Adjust markers with timline when ripple all tracks is used.

bmatherly:
Would like your feedback on this approach in case you think of a better approach. I considered creating marker commands and then making them children of the timeline commands. But that requires a common parent command and it seemed like it would be a lot more to change.


ddennedy:
I have not looked at this deeply yet or even tried it, but I am concerned that basing it on ripple-all-tracks is not what users will want. I believe most users avoid that option for its unexpected behavior but will still want markers to move when doing ripple operations. I am not sure what people who want this capability expect when moving clips. And I bet some people only want existing markers to ripple when doing ripple operations on V1 alone (except when they place a solid color background color on V1 and use V2 for the main video layer).

bmatherly:
> I am concerned that basing it on ripple-all-tracks is not what users will want.

That is a good comment and I value your insight on this. In the previous discussion, there were two ideas: ripple all tracks, or add a separate "ripple markers" button. Maybe a separate button makes more sense.

In my testing, this implement make sense in a few ways:
1) When you ripple remove a clip, it leaves a gaping hole in the timeline with no clips. So it stands to reason that any markers applied to that segment would be irrelevant.
2) When you ripple move a clip, it makes sense that all the markers would move proportionally
These could also be accomplished with a separate button. It would add an extra click, but it would also give the user more control.

src/commands/timelinecommands.cpp
```diff
@@ -109,18 +112,27 @@ void InsertCommand::redo()
...
             m_model.insertClip(m_trackIndex, clip, m_position, m_rippleAllTracks, false, lastClip);
+            shift += clip.get_playtime();
         }
     } else {
         ProxyManager::generateIfNotExists(clip);
         m_model.insertClip(m_trackIndex, clip, m_position, m_rippleAllTracks, m_seek);
+        shift = clip.get_playtime();
     }
     m_undoHelper.recordAfterState();
+    if (m_rippleAllTracks && shift > 0) {
```
ddennedy, line 123: Instead of using `m_rippleAllTracks` here, let's add a new member `m_rippleMarkers` and initialize it in the same way currently? Then, if decide make a distinct option for marker toggle, quite possibly in beta, it will be easier to change.

ddennedy:
For an icon, you can base it upon the ripple-all-tracks icon with a marker inside instead of an asterisk.

bmatherly:
I added a button. Here is what I came up with for an icon:

image
image
image

ddennedy:
Those look good.

ddennedy (approved):

PR #1194, opened 2021-12-01
Various keyframe improvement

traprog:
Slight adjustment to keyframe `DragY` bounds
- A keyframe can sink half way into the floor or go half way through the ceiling though Line 112
`var trackValue = Math.min(Math.max(0, 1.0 - parent.y / (parameterRoot.height - parent.height)), 1.0)`
is designed to allow a value between `parameterRoot.height - parent.height`.

Double-click no longer deleting keyframe
- Since a year ago or so when QT was upgraded, `trigger()` hasn't worked. Somehow it worked in older QT versions.

Ctrl and Alt during `onPositionChanged`
- If Ctrl and Alt key press checks are in `onPressed`, either key has to be pressed before a left mouse click. I don't think I have experienced that behavior in any other programs. Like in file explorers and forum text edit areas, I can typically be able to press Ctrl during a mouse movement to change the operation from moving to copying.
- I can also release Ctrl and Alt during a mouse movement, and X-axis or Y-axis movement lock is still intact. I also found it strange.

I didn't squash these commits because they look like separate issues. You can squash them upon merge because they look like a collection of tiny changes.

### DIFF
--- a/src/qml/views/keyframes/Keyframe.qml
+++ b/src/qml/views/keyframes/Keyframe.qml
@@ -35,8 +35,8 @@ Rectangle {
     property double trackValue: (0.5 - (value - minimum) / (maximum - minimum)) * (trackHeight - height - 2.0 * border.width)
     property double minDragX: activeClip.x - width/2
     property double maxDragX: activeClip.x + activeClip.width - width/2
-    property double minDragY: activeClip.y - width/2
-    property double maxDragY: activeClip.y + activeClip.height - width/2
+    property double minDragY: activeClip.y
+    property double maxDragY: activeClip.y + activeClip.height - width
     property bool inRange: position >= (filter.in - producer.in) && position <= (filter.out - producer.in)
 
     signal clicked(var keyframe)
--- a/src/qml/views/keyframes/Keyframe.qml
+++ b/src/qml/views/keyframes/Keyframe.qml
@@ -96,6 +96,12 @@ Rectangle {
                   drag.minimumY = minDragY
                   drag.maximumY = maxDragY
                }
+            }
+        }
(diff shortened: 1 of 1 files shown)


ddennedy:
> Slight adjustment to keyframe DragY bounds

Give an example how to reproduce.

> Double-click no longer deleting keyframe

accepted

> Ctrl and Alt during onPositionChanged

accepted

traprog:
> Give an example how to reproduce.
1. Open a video file.
2. Trim in and out on the source tab to make it shorter. 
This is preferable because a keyframe curve line disappears if a clip is too long. This has been discussed on the forum.
3. Move the clip to the timeline. This is optional.
4. Apply a brightness filter and click on the use keyframes button.
5. Move the advanced keyframe upward to make the value 2.0 or 200%.
6. The keyframe goes half way into the clip section, and snap back into the parameter section. The snap back distance is extactly the amount of `width/2`. The curve line is also close to unnoticeable because it's so close to the clip and parameter section boarder.
7. Likewise, an advanced keyframe snaps back up `width/2` to the parameter section if you move it downward to make the value 0.0.

The ceiling and floor were not good descriptions. What I meant were the boarder between the clip section and parameter section and the boarder between the parameter section and the dock background.

ddennedy:
> The keyframe goes half way into the clip section

I do not see that:
image

> and snap back into the parameter section

I do not understand the "snap" - maybe because I do not see the "half way into the clip section."

> The curve line is also close to unnoticeable because it's so close to the clip and parameter section boarder.

It looks OK in my screenshot in my opinion.

ddennedy:
> Move the advanced keyframe upward to make the value 2.0 or 200%.

Oh, sorry, in my test I had dragged the slider in the filter parameters UI. When I drag the keyframe I see what you mean.

PR #1195, opened 2021-12-02
Fix misc. typos

luzpaz:


ddennedy:
You are correcting a generated file where the code being corrected was not our own but rather included from a npm dependency.

luzpaz:
sorry about that.

PR #1197, opened 2021-12-06
Audiolevelgraph

bmatherly:
As suggested in this thread:
https://forum.shotcut.org/t/disco-mix-video-extract-featuring-simulated-spectrum-analyzer-leds/31054/11

Depends on https://github.com/mltframework/mlt/pull/756

### DIFF
--- a/src/qml/filters/audiolevelgraph/vui.qml
+++ b/src/qml/filters/audiolevelgraph/vui.qml
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2021 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


src/qml/filters/audiolevelgraph/vui.qml
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright (c) 2017-2020 Meltytech, LLC
```
ddennedy, line 2: Change copyright year to current year only in all new files even when copying and pasting a lot of the code from somewhere else.

bmatherly: done

PR #1201, opened 2021-12-15
Use new animation previous and next functions

bmatherly:
Utilize new error checking functionality.

Depends on https://github.com/mltframework/mlt/pull/758


ddennedy:
Hmm, this is going to require a new MLT release for the next Shotcut release. I try to avoid making and requiring a new MLT release every Shotcut release. Something like a new filter will not load dynamically if not present. Going to wait and see what else goes into MLT and its criticality.

bmatherly:
> Going to wait and see what else goes into MLT and its criticality.

This does not fix any known bugs - we have already worked around the behavior of the original functions. So I think we can wait with this change until there is another reason to release a new MLT version.

PR #1207, opened 2021-12-30
Improvements to Simple/Advanced keyframe interactions

bmatherly:
Implement consistent behaviors for interactions between simple and advanced keyframes as discussed here:
https://forum.shotcut.org/t/crop-rectangle-filters-corner-radius-appears-when-only-wanting-to-open-position-size-if-a-simple-keyframe-is-set-already/29184/7

### DIFF
--- a/src/models/keyframesmodel.cpp
+++ b/src/models/keyframesmodel.cpp
@@ -610,9 +610,7 @@ void KeyframesModel::removeAdvancedKeyframes()
 
 bool KeyframesModel::simpleKeyframesInUse()
 {
-    if (m_filter && m_metadata && (m_filter->animateIn() > 0 || m_filter->animateOut() > 0))
-        return true;
-    return false;
+    return m_filter && m_metadata && (m_filter->animateIn() > 0 || m_filter->animateOut() > 0);
 }
 
 void KeyframesModel::removeSimpleKeyframes()
--- a/src/models/keyframesmodel.cpp
+++ b/src/models/keyframesmodel.cpp
@@ -581,6 +581,30 @@ bool KeyframesModel::isKeyframe(int parameterIndex, int position)
     return false;
 }
 
+bool KeyframesModel::advancedKeyframesInUse()
+{
+    if (m_filter && m_metadata && m_filter->animateIn() <= 0 && m_filter->animateOut() <= 0)
+    for (int i = 0; i < m_metadata->keyframes()->parameterCount(); i++) {
+        if (m_filter->keyframeCount(m_metadata->keyframes()->parameter(i)->property()) > 0) {
+            return true;
+        }
+    }
+    return false;
+}
+
+void KeyframesModel::removeAdvancedKeyframes()
+{
+    if (m_filter && m_metadata && m_filter->animateIn() <= 0 && m_filter->animateOut() <= 0) {
+        for (int i = 0; i < m_metadata->keyframes()->parameterCount(); i++) {
+            QString name = m_metadata->keyframes()->parameter(i)->property();
+            if (m_filter->keyframeCount(name) > 0) {
+                m_filter->set(name, m_filter->get(name, 0));
+            }
+        }
+        reload();
+    }
+}
+
 void KeyframesModel::reload()
 {
     beginResetModel();
--- a/src/models/keyframesmodel.cpp
+++ b/src/models/keyframesmodel.cpp
@@ -237,15 +237,13 @@ void KeyframesModel::load(QmlFilter* filter, QmlMetadata* meta)
     m_metadataIndex.clear();
     m_filter = filter;
     m_metadata = meta;
-    if (m_filter && m_metadata)
+    if (m_filter && m_metadata && m_filter->animateIn() <= 0 && m_filter->animateOut() <= 0)
     for (int i = 0; i < m_metadata->keyframes()->parameterCount(); i++) {
-        if (!m_metadata->keyframes()->parameter(i)->isSimple() || (m_filter->animateIn() <= 0 && m_filter->animateOut() <= 0)) {
-            if (m_filter->keyframeCount(m_metadata->keyframes()->parameter(i)->property()) > 0) {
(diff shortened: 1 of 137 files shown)


src/models/keyframesmodel.cpp
```diff
@@ -599,12 +599,72 @@ void KeyframesModel::removeAdvancedKeyframes()
...
         }
         reload();
     }
 }
 
+bool KeyframesModel::simpleKeyframesInUse()
+{
+    if (m_filter && m_metadata && (m_filter->animateIn() > 0 || m_filter->animateOut() > 0))
+        return true;
+    return false;
```
ddennedy, line 615: This can be simplified to
`return m_filter && m_metadata && (m_filter->animateIn() > 0 || m_filter->animateOut() > 0);`

src/qmltypes/qmlfilter.cpp
```diff
@@ -531,6 +531,23 @@ void QmlFilter::setAnimateIn(int value)
...
+            for (int i = 0; i < m_metadata->keyframes()->parameterCount(); i++) {
+                QString name = m_metadata->keyframes()->parameter(i)->property();
+                Mlt::Animation anim = getAnimation(name);
+                if (anim.is_valid() && anim.key_count() > 0) {
+                    QString value;
+                    if (anim.key_count() > 1) {
+                        value = m_service.anim_get(qUtf8Printable(name), 1);
+                    } else {
+                        value = m_service.anim_get(qUtf8Printable(name), 0);
+                    }
```
ddennedy, line 545: I do not understand this logic. It gets the value for the first or second frame depending on the number of keyframes. That seems rather arbitrary. I think it should get the value at the current time (playhead) position: make it always look like I see it now. Same for `setAnimateOut()` below.

src/models/keyframesmodel.cpp
```diff
@@ -583,20 +581,102 @@ bool KeyframesModel::isKeyframe(int parameterIndex, int position)
...
+}
+
+void KeyframesModel::removeSimpleKeyframes()
+{
+    if (simpleKeyframesInUse()) {
```
ddennedy, line 620: `removeSimpleKeyframes()` is never called! And this is the only call `simpleKeyframesInUse()`. Both can be removed. If you want to keep them for completeness, then it should take an argument for the current frame position and get the values at that position.

src/models/keyframesmodel.cpp
```diff
@@ -583,20 +581,102 @@ bool KeyframesModel::isKeyframe(int parameterIndex, int position)
...
+
+void KeyframesModel::removeAdvancedKeyframes()
+{
+    if (m_filter && m_metadata && m_filter->animateIn() <= 0 && m_filter->animateOut() <= 0) {
+        for (int i = 0; i < m_metadata->keyframes()->parameterCount(); i++) {
+            QString name = m_metadata->keyframes()->parameter(i)->property();
+            if (m_filter->keyframeCount(name) > 0) {
+                m_filter->set(name, m_filter->get(name, 0));
+                for (auto& key : m_metadata->keyframes()->parameter(i)->gangedProperties()) {
+                    m_filter->set(key, m_filter->get(key, 0));
```
ddennedy, line 603: Use values at current frame position instead of first frame?

ddennedy (changes requested):

src/models/keyframesmodel.cpp
bmatherly: Done

src/qmltypes/qmlfilter.cpp
bmatherly: The goal is to set the value to the "middle" simple keyframe value. I think from the user's perspective, it should be "I no longer want to fade in, so make it always be the middle value". In this case, the middle value will be the keyframe at index 1 (index 0 is the "start" simple keyframe). the index 0 code is just a failsafe that I never triggered in testing. I updated the code to clarify the goal of those lines.

Your comment also made me realize that I was not using the correct index to choose the middle keyframe for the animOut case. So I fixed that logic.

src/models/keyframesmodel.cpp
bmatherly: It is called in QML. When the user has simple keyframes in use, and they click the advanced keyframes "clock" button, they are prompted to ask if they want to convert simple keyframes to advanced keyframes. This function is called if they choose "yes".

I was debating if this function should be in KeyframesModel or QmlFilter. I decided on Keyframes model since it has some other similar convenience functions. But I am open to suggestions.

src/models/keyframesmodel.cpp
bmatherly: I am willing to change that. But I am not sure it if would be better. 

This function is called when:
* Advanced keyframes are in use
AND
* The user grabs the simple keyframes handle
THEN
The user is prompted to ask if they want to clear simple keyframes with an option to cancel.

Why did they grab the handle if they have already done work to set advanced keyframes? It was probably a mistake and they will choose cancel. If it was not a mistake, did they carefully position the cursor on the exact frame that has a value that they want? Maybe not. Maybe the cursor is sitting arbitrarily between two advanced keyframes on a value that they never selected.

The current code (using keyframe 0) at least sets the parameter value to the first value the user had selected.

I am open to changing it, just wanted to give more perspective.

src/models/keyframesmodel.cpp
ddennedy: Sorry, now I found it. I was searching the PR web page but had a file makred as viewed, which was hiding it.

src/qmltypes/qmlfilter.cpp
ddennedy: Sorry, I was mistaken and had read it as get property value at position instead of an anim_get. This is fine.

src/models/keyframesmodel.cpp
ddennedy: I guess this is fine as-is and somewhat ambiguous either way.

ddennedy (approved):

PR #1208, opened 2022-01-03
translation and correction

fylip22:
translation and correction


ddennedy:
Sorry we do not take translations through GitHub but rather Transifex. If I accept this then it might overwrite recent changes made in transifex and the next time I pull from transifex it will overwrite yours.

PR #1220, opened 2022-01-23
Improve audio distortion when reversing a clip.

bmatherly:
Use eac3 and mov for all intermediate files for reverse. This
combination results in the best audio syncing and the least
artifacts when reversing a clip.

As reported here:
https://forum.shotcut.org/t/reversing-a-clip-on-linux-causes-the-audio-to-crackle/31713

This proposed combination was found through experimentation (trial and error).

ac3 and eac3 are the only audio formats that I can find that result in zero distortions in my tests. Even PCM has occasional distortion. I assume this is due to better seeking to AC3 frames. But I do not know why the seeking would be better for that format. So I chose eac3 @ 640kbps for all quality levels (good, better, best).

Of all the quality levels, "best" had the worst distortion - even when using eac3 for audio. In the decoding log, I see this error:

> [matroska,webm @ 0x55bd10939e80] parser not found for codec utvideo, packets or times may be invalid.
> [matroska,webm @ 0x55bd10939e80] parser not found for codec pcm_s32le, packets or times may be invalid.

In this change, I change intermediates for all quality levels to use MOV format. Using MOV, I do not see the above errors, and I do not hear any distortion in the audio.

Throwing this out for review in case anyone has any other insights worth trying. I do not know if this worked perfectly in the past and suffered a regression when FFMpeg was upgraded. I just know these combinations work pretty good in my tests.

### DIFF
--- a/src/widgets/avformatproducerwidget.cpp
+++ b/src/widgets/avformatproducerwidget.cpp
@@ -1122,11 +1122,12 @@ void AvformatProducerWidget::on_reverseButton_clicked()
         if (m_producer->get_int("video_index") == -1)
             meltArgs << "vn=1" << "video_off=1";
 
+        ffmpegArgs << "-f" << "mov" << "-codec:a" << "alac";
+
         switch (dialog.format()) {
         case 0:
             path.append("/%1 - %2.mp4");
             nameFilter = tr("MP4 (*.mp4);;All Files (*)");
-            ffmpegArgs << "-f" << "mov" << "-codec:a" << "alac";
             if (ui->scanComboBox->currentIndex()) { // progressive
                 ffmpegArgs << "-codec:v" << "dnxhd" << "-profile:v" << "dnxhr_hq" << "-pix_fmt" << "yuv422p";
             } else { // interlaced
@@ -1137,7 +1138,6 @@ void AvformatProducerWidget::on_reverseButton_clicked()
             meltArgs << "vpreset=medium" << "g=1" << "crf=11";
             break;
         case 1:
-            ffmpegArgs << "-f" << "mov" << "-codec:a" << "alac";
             meltArgs << "acodec=alac";
             if (ui->scanComboBox->currentIndex()) { // progressive
                 ffmpegArgs << "-codec:v" << "dnxhd" << "-profile:v" << "dnxhr_hq" << "-pix_fmt" << "yuv422p";
@@ -1151,9 +1151,7 @@ void AvformatProducerWidget::on_reverseButton_clicked()
             nameFilter = tr("MOV (*.mov);;All Files (*)");
             break;
         case 2:
-            ffmpegSuffix = "mkv";
-            ffmpegArgs << "-f" << "matroska" << "-codec:a" << "pcm_s32le" << "-codec:v" << "utvideo";
-            ffmpegArgs << "-pix_fmt" << "yuv422p";
+            ffmpegArgs << "-codec:v" << "utvideo" << "-pix_fmt" << "yuv422p";
             if (!ui->scanComboBox->currentIndex()) { // interlaced
                 meltArgs << "field_order=" + QString::fromLatin1(ui->fieldOrderComboBox->currentIndex()? "tt" : "bb");
             }


src/widgets/avformatproducerwidget.cpp
```diff
@@ -1151,9 +1151,7 @@ void AvformatProducerWidget::on_reverseButton_clicked()
             nameFilter = tr("MOV (*.mov);;All Files (*)");
             break;
         case 2:
-            ffmpegSuffix = "mkv";
-            ffmpegArgs << "-f" << "matroska" << "-codec:a" << "pcm_s32le" << "-codec:v" << "utvideo";
-            ffmpegArgs << "-pix_fmt" << "yuv422p";
+            ffmpegArgs << "-codec:v" << "utvideo" << "-pix_fmt" << "yuv422p";
```
ddennedy, line 1154: Hmm, Ut Video in MOV?!?

ddennedy (changes requested):
I tried to reproduce it by making a 30 second video with sine wave at 720p59.94 and reversing it using the three quality levels. I reproduced it on good (MP4) and better (MOV) but not best (MKV). Testing the branch, and now I have a problem with all 3! Here is "best" now

image

The convert and reverse job logs did not complain about utvideo in MOV.
With `ffmpegArgs << "-f" << "mov" << "-codec:a" << "pcm_s32le";` I do not experience a problem at any output quality including "good":

image


Please test and confirm my finding.

bmatherly:
Thanks for testing. That was a good test.

My immediate observation is that my test clip is 1080p30. I made a new clip that is 720p59 and eac3 is as bad as everything else. So there must be come coincidence between my 1080p30 clip and eac3.

> With ffmpegArgs << "-f" << "mov" << "-codec:a" << "pcm_s32le"; I do not experience a problem at any output quality

I  tested this, and I did not have good results with either test clip (frame rate). But I think it would be a good solution if this could be the format for all modes.

I actually see the fewest artifacts (but still some) with ffmpegArgs << "-f" << "mov" << "-codec:a" << "alac";

Besides the framerate, maybe there is some other fundamental difference between our test clips.

I have shared mine here if you have a chance to try reversing them. They were made with Shotcut and default export settings.
https://drive.google.com/file/d/1opydF-AOE_fey1MfANjHxhdz_xWGNp5o/view?usp=sharing
https://drive.google.com/file/d/1-5HljMKzu-JPP9qcZzcc2b5-RSq51tjV/view?usp=sharing

If you find that your test clips can reverse without artifacts, but mine can not, I would like to compare the differences between the test clips.

> The convert and reverse job logs did not complain about utvideo in MOV.

That was also my observation. Since the file is intermediate and deleted by Shotcut, it would not necessarily need to be a file that is playable by other players.

ddennedy:
The difference between our clips is that I exported the sine tone using the lossless/Ut Video preset, and it makes the export frame rate appear in Shotcut as 59.94.0063! In More Information it shows it as
```
r_frame_rate=19001/317
avg_frame_rate=19001/317
time_base=1/1000
```
So, this is a separate bug related to all Matroska and WebM outputs. With your test clip `countdown_720p59.mp4` I do get less artifacts when using alac over pcm_32le, but all of this leads me to believe this cannot be solved with a simple intermediate format change. I suspect various frame rates will yield different results, and ultimately we need to look deeper into the avformat producer.

ddennedy:
I tested this today using your 720p59.94 test clip and Shotcut v21.06 (before MLT changes for FFmpeg API deprecations and an upgrade from 4.2 to 4.4). It was basically the same problem, and export to matroska with non-integer frame rate has same inaccuracy.

ddennedy:
About Matroska frame rate, read https://gitlab.com/mbunkus/mkvtoolnix/-/wikis/Wrong-frame-rate-displayed
Way to take a player-centric or idealized (everything should support variable frame rate) view of things! Geez. We must demote usage of Matroska for intermediates and lossless. Also, ffmpeg 4.4 at the command line produces the same output including ffprobe stream details as MLT. MediaInfo skirts the issue by rounding to 3 decimal places.

ddennedy:
Since this is an "improvement" and not a comprehensive fix, I will accept alac/mov as an improvement for the next release 22.01. Please submit that.

I also want to change Convert and the lossless Matroska presets to AVI. For FFV1, HuffYUV, and Ut Video AVI seems roughly just as popular as Matroska for these even for digital preservation standards: https://en.wikipedia.org/wiki/FFV1
I am doing some testing to ensure a 2 hour long, large file holds up fine with all of these codecs.

bmatherly:
> all of this leads me to believe this cannot be solved with a simple intermediate format change

Yes. I draw the same conclusion.

> ultimately we need to look deeper into the avformat producer.

Agreed.

> I tested this today using your 720p59.94 test clip and Shotcut v21.06 (before MLT changes for FFmpeg API deprecations and an upgrade from 4.2 to 4.4). It was basically the same problem, and export to matroska with non-integer frame rate has same inaccuracy.

Good test. Thanks for doing that.

> We must demote usage of Matroska for intermediates and lossless. 
> I also want to change Convert and the lossless Matroska presets to AVI.

I tried AVI in my testing and I saw that same error:
`parser not found for codec utvideo, packets or times may be invalid.`
That is why I had suggested MOV instead. Can you keep an eye out for that error if you are testing intermediate files with AVI? To be clear, the warning message is printed when decoding the intermediate AVI file. Hopefully I was mistaken.

> I will accept alac/mov as an improvement for the next release 22.01. Please submit that.

Done

ddennedy:
We can continue discussion about Matroska or AVI here with respect to Convert and Reverse targets (not reverse intermediate) and lossless export presets. From my study, Quicktime is not a commonly used format for FFV1, HuffYUV, or Ut Video while AVI is. Here is what FFmpeg developer Nicholas George says about this message (written as if perhaps he added the log message):
https://lists.mplayerhq.hu/pipermail/ffmpeg-user/2011-November/003279.html
Indeed, when I open Ut Video or HuffYUV AVI in Shotcut from my IDE from a debug build, I do see these messages but not when I run a release package. Shotcut runs with a more verbose log level for MLT and FFmpeg from a debug build (see `main.cpp`). I strongly suspect for these I-frame only videos, there is not a problem at least for the video, but the audio I am not sure about. For the Ut Video, it is complaining about pcm_s24le, but consider that is uncompressed.

austincbrooks:
Very unfortunate about Matroska's interpretation of frame rate.

As for AVI, something we learned before is that Ut Video can signal BT.709 vs BT.601 while Huffyuv and FFV1 cannot.  Ut Video has dedicated FourCC for each colorspace that gets saved in the stream which can preserve the colorspace, whereas Huffyuv and FFV1 seem to signal colorspace with the container (which AVI doesn't support and the information is therefore lost).

I haven't run an updated colorspace test with newer FFmpeg, so potentially things have changed.  But this was a primary reason I used Ut Video so much over Huffyuv and FFV1 in the past.  Or rather, always stuffed them in Matroska to get the colorspace flag.

Ut Video's FourCC list:
http://umezawa.dyndns.info/archive/utvideo/utvideo-19.0.1-readme.en.html

ddennedy:
Thanks, @austincbrooks . I knew you would remember something I forgot. I just ran into the full range issues myself when trying different lossless codecs and muxers for a separate project. That left lossless H.264 as my best option, and I could also set its pix_fmt=yuv422p. So, one thing I am thinking about doing is keeping Matroska, and in MLT's avformat producer rounding the frame rate and then coercing it to one of the broadcast standard framerates based on `Util::normalizeFrameRate` in Shotcut. A question is whether to limit that activity to Matroska or not.

ddennedy:
```
--- src/modules/avformat/producer_avformat.c
+++ src/modules/avformat/producer_avformat.c
@@ -2250,8 +2250,32 @@ static int video_codec_init( producer_avformat self, int index, mlt_properties p
 			frame_rate.num = profile->frame_rate_num;
 			frame_rate.den = profile->frame_rate_den;
 		}
 
+		// Normalize broadcast frame rates for Matroska
+		if (self->video_format->iformat->name && !strcmp(self->video_format->iformat->name, "matroska,webm")) {
+			switch (lrint(100000.0 * frame_rate.num / frame_rate.den)) {
+			case 2997003:
+				frame_rate.num = 30000;
+				frame_rate.den = 1001;
+				break;
+			case 5994006:
+				frame_rate.num = 60000;
+				frame_rate.den = 1001;
+				break;
+			case 2397602:
+				frame_rate.num = 24000;
+				frame_rate.den = 1001;
+				break;
+			case 4795204:
+				frame_rate.num = 48000;
+				frame_rate.den = 1001;
+				break;
+			default:
+				break;
+			}
+		}
+
 		self->video_time_base = stream->time_base;
 		if ( mlt_properties_get( properties, "force_fps" ) )
 		{
 			AVRational force_fps = av_d2q( mlt_properties_get_double( properties, "force_fps" ), 1024 );
```

ddennedy:
Seems we may need the above code after all per [twitter conversation](https://twitter.com/ddennedy/status/1485808649121329153) with David Rice, an archival and preservation expert - as part of the main group standardizing on FFV1/Matroska for archiving. This leads to the Matroska site's notes on [DefaultDecodedFieldDuration](https://www.matroska.org/technical/notes.html) and its examples.

bmatherly:
> Seems we may need the above code after all per twitter conversation with David Rice, an archival and preservation expert - as part of the main group standardizing on FFV1/Matroska for archiving. This leads to the Matroska site's notes on DefaultDecodedFieldDuration and its examples.

Should I work on a patch for ffmpeg?

ddennedy:
I doubt FFmpeg will accept it and prefer to keep the simple mathematical conversion it has. Up to you if you want to try.

PR #1224, opened 2022-01-26
Add job priority menu option

bmatherly:
Staging this here for the next release and for discussion.

Added a menu option to choose "low" or "normal" priority for jobs

image

As suggested here:
https://forum.shotcut.org/t/melt-exe-priority-vs-alder-lake-on-windows-10/31706

### DIFF
--- a/src/jobs/abstractjob.cpp
+++ b/src/jobs/abstractjob.cpp
@@ -17,6 +17,7 @@
 
 #include "abstractjob.h"
 #include "postjobaction.h"
+#include "settings.h"
 #include <QApplication>
 #include <QTimer>
 #include <Logger.h>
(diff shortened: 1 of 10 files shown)


ddennedy (approved):
Thanks for this! I was thinking about doing it. I do not mind to include this in the 22.01 release because it is very low risk (no heavy logic change) and easily testable. Do you agree?

bmatherly:
> I do not mind to include this in the 22.01 release because it is very low risk (no heavy logic change) and easily testable. Do you agree?

Yes. I agree. I tested in Windows and Linux. Can you test on Mac? Please test export, convert, reverse and ffprobe jobs.

And be aware that this change adds translation strings.

I am eager for the 22.01 release because it fixes some good issues. What timeline do you expect for that release?

PR #1225, opened 2022-01-31
Add notes panel

bmatherly:
As suggested here:
https://forum.shotcut.org/t/notes-feature/20830
https://forum.shotcut.org/t/add-comments-to-project/31040

Posting for comments and ideas. Need polish before being merged (scroll bars, etc).
This implementation allows rich text:

image

The text is saved in the project properties of the xml file:

image

Topics for discussion:
* Is this a good idea at all?
* Use rich text or just plain text?
* Should the background color of the dock be forced to white regardless of the theme?
* Add more feature like bullet/numbered list?


ddennedy:
> Is this a good idea at all?

Yes, even though I do not plan to use it. Maybe I will eventually; who knows.

> Use rich text or just plane text?

One can choose to not do any formatting, but I guess the concern is what people might want to do with the text inside the XML if they do some XML processing outside Shotcut. That is very difficult to anticipate. Will one person want a fixed width font for some crude plain text formatting while another would find it ugly? If plain text, there will be a request to change the font and size.

> Should the background color of the dock be forced to white regardless of the theme?

I do not think so. Many people will complain about the big bright block of white on the default and most popular dark theme. We will get a request to make the background color configurable, probably very early on; however, it is quite usable without it and more like plain text as long as the text color follows the theme until formatted.

> Add more feature like bullet/numbered list?

That would be nice but also a slippery slope: next request will be for a way to tick off items and provide todo list functionality. See this GitHub editor's toolbar. :-P In its absence people will develop their own conventions like dash, asterisk, etc. And they can assign their own meaning to them - plain text approach. If you decide to add this, perhaps drop alignment or indent actions on the toolbar to make room for them.

bmatherly:
> One can choose to not do any formatting

I thought about this. But even without applying any formatting the rich text object adds HTML tags:

```
    <property name="shotcut:projectNote">&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
&lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
p, li { white-space: pre-wrap; }
&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Ubuntu'; font-size:11pt; font-weight:400; font-style:normal;&quot;&gt;
&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;No Formatting&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</property>
```

Maybe I could add a way to detect this and strip out HTML tags. Or I could add an option to switch between plain and rich text.

> but I guess the concern is what people might want to do with the text inside the XML if they do some XML processing outside Shotcut

I agree with this. I think people might like to write scripts to read or modify things like this. And escaped HTML tags will make that difficult.

> I do not think so. Many people will complain about the big bright block of white on the default and most popular dark theme.

I also agree with this.

> next request will be for a way to tick off items and provide todo list functionality. 

I also thought about this. I considered adding checkbox features to the notes panel. Then I also considered making a separate "TODO" panel with timeline timecodes. But now I think that someone could use markers to achieve something similar. They could create a marker at a timeline location with text that describes what is left to be done. And they could choose a certain color to indicate "todo marker".

I will spend some time to see how plain text only works out. Maybe that will be better.

timlau:
Just some crazy ideas, that pops up in my head.

* Markdown formatted notes, instead of html, there properly exist some QT makedown widget out there.
* Another approach for project notes, is to being able to add non media items to the playlist and just store locations in the .mlt and open then with the default OS viewer for that that filetype, then shotcut don't have to handle all the problems an Richtext editor will introduce

ddennedy:
Sorry @timlau, but I need to reject both of those ideas.

bmatherly:
> Markdown formatted notes, instead of html, there properly exist some QT makedown widget out there.

TextArea does have support for markup. But I'm not sure if I am willing to put in the effort to make an editor. We already have a RichText editor in Shotcut.

https://doc.qt.io/qt-5/qml-qtquick-textedit.html#textFormat-prop

timlau:
It was just crazy ideas, so no hard feelings :-D
Of cause no need to build a custom editor, when QT already has one.

ddennedy:
Maybe this should instead be a teleprompter for voiceover recording, and one can use it for notes as well by adding a foreward page:
https://qprompt.app/

bmatherly:
OK. I spiffed it up a little by adding options to switch between rich text and plain text. I also added scrollbars.

image

I am open to comments/suggestions.

bmatherly:
Rich text is now disabled (as discussed).

PR #1233, opened 2022-02-20
Add drag track headers to reorder

bmatherly:
This is my first attempt at drag-n-drop track reordering. The track headers have an added handle on the left size. When the user hovers the mouse over the handle, and expands and allows the user to drag the track header. Other track headers highlight with a green outline if they are a valid drop target.

drag_track

Feedback is welcome.

Also, any ideas about how to keep this code from exploding in size and complexity? Maybe i should pull out the track header column into its own file to reduce the levels of hirearchy.

### DIFF
--- a/src/qml/views/timeline/timeline.qml
+++ b/src/qml/views/timeline/timeline.qml
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2021 Meltytech, LLC
+ * Copyright (c) 2013-2022 Meltytech, LLC
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -18,6 +18,7 @@
 import QtQuick 2.12
 import QtQml.Models 2.12
 import QtQuick.Controls 2.12
+import QtQuick.Dialogs 1.3
 import QtGraphicalEffects 1.12
 import QtQuick.Window 2.12
 import QtQuick.Layouts 1.12
@@ -213,6 +214,8 @@ Rectangle {
                         id: trackHeaderRepeater
                         model: multitrack
                         TrackHead {
+                            id: trackHead
+                            property var trackIndex: index
                             trackName: model.name
                             isMute: model.mute
                             isHidden: model.hidden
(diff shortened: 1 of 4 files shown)


src/qml/views/timeline/timeline.qml
```diff
@@ -232,6 +235,103 @@ Rectangle {
...
+                                Behavior on width { PropertyAnimation {easing.type: Easing.InOutCubic} }
+                                onReleased: {
+                                    if (drag.active) {
+                                        dragItem.Drag.drop()
+                                    }
+                                }
+                                Rectangle {
+                                    id: dragItem
+                                    property var trackHead: trackHead
+                                    color: activePalette.dark
```
ddennedy, line 259: The palette dark color is rather undefined but I see it as a mid-gray. I think it should be the same as the the block that appears when you drag a clip to the timeline, which is white with 50% opacity. However, 50% might be too translucent here. So, I think this color is close enough. But I am wondering if this should be the same solid color or some slightly alternating shade like we do for the track background color?

src/qml/views/timeline/timeline.qml
```diff
@@ -232,6 +235,103 @@ Rectangle {
...
+                                    Drag.hotSpot.x: width / 2
+                                    Drag.hotSpot.y: height / 2
+                                    Drag.keys: ["trackHeader"]
+                                    Label {
+                                        anchors.fill: parent
+                                        horizontalAlignment: Text.AlignHCenter
+                                        verticalAlignment: Text.AlignVCenter
+                                        elide: Text.ElideRight
+                                        text: qsTr("Move track %1").arg(trackHead.trackName)
+                                        color: activePalette.light
```
ddennedy, line 271: I think the white on gray looks a little funny in the lighter themes, and maybe it should just follow the theme text color. The Overwrite/Insert block when dragging to the timeline uses black with a white outline, but I think that would look bad on this smaller size text.

src/qml/views/timeline/timeline.qml
```diff
@@ -232,6 +235,103 @@ Rectangle {
...
+                                    containsValidDrag = false
+                                }
+                                onDropped: {
+                                    if (drop.proposedAction == Qt.MoveAction &&
+                                        trackHead.isVideo == drop.source.trackHead.isVideo &&
+                                        trackHead.trackIndex != drop.source.trackHead.trackIndex) {
+                                        drop.acceptProposedAction()
+                                        movePrompt.fromTrack = drop.source.trackHead
+                                        movePrompt.visible = true
+                                    }
```
ddennedy, line 323: Overall, I like it and think its basically ready to merge. Only other comment is perhaps here when the drop is rejected we show a status message in the designated status message area.

src/qml/views/timeline/timeline.qml
bmatherly: Thank you for mentioning the Overwrite/Insert block. I changed the styling to match that - including the larger text size. I think it is much better. Let me know what you think.

Also, I changed the narrow handle to be transparent until the mouse hovers over it. So not it is less obvious that there is a reason to put the mouse there, but the color is not visible all the time - only when hovering/dragging. I think it restores the original visual appeal.

Additionally, I also set the current track to whichever track is being hovered over (as the overwrite drag does). So the white panel is always over a consistent color background.

Have a look and see what you think.

src/qml/views/timeline/timeline.qml
bmatherly: Added some status messages - good idea.

src/qml/views/timeline/timeline.qml
ddennedy: I think it might be a little better cosmetically except the only hint that this is draggable is the mouse pointer, which is subtle. I think a "Move %1" would be better (shorter than "Move track %1"). I was not asking to use the larger text size, which reduces the space for the text. Hmm... in fact, this rectangle width does not need to be limited to the header:
```
--- src/qml/views/timeline/timeline.qml
+++ src/qml/views/timeline/timeline.qml
@@ -248,3 +248,3 @@ Rectangle {
                                 drag.minimumY : -y - height / 2
-                                width: containsMouse ? trackHead.width : 8
+                                width: containsMouse ? root.width : 8
                                 Behavior on width { PropertyAnimation {easing.type: Easing.InOutCubic} }
@@ -267,3 +267,3 @@ Rectangle {
                                         anchors.fill: parent
-                                        text: trackHead.trackName
+                                        text: qsTr("Move %1").arg(trackHead.trackName)
                                         style: Text.Outline
```

But there is some clipping getting in the way of that until you drag, which might not be so bad. Maybe it is rather heavy to paint over the whole track when you accidentally hover into the track drag handle area.
Screenshot from 2022-02-21 11-17-05

ddennedy: For some reason, my change breaks the action.

bmatherly: Here is a suggestion that sets the width to the text width and sets the width before drag starts:

```
--- a/src/qml/views/timeline/timeline.qml
+++ b/src/qml/views/timeline/timeline.qml
@@ -246,7 +246,7 @@ Rectangle {
                                 drag.axis: Drag.YAxis
                                 drag.maximumY: trackHeaders.height - y - height / 2
                                 drag.minimumY : -y - height / 2
-                                width: containsMouse ? trackHead.width : 8
+                                width: containsMouse | drag.active ? Math.max(parent.width, dragItemText.contentWidth + 20) : 8
                                 Behavior on width { PropertyAnimation {easing.type: Easing.InOutCubic} }
                                 onReleased: {
                                     if (drag.active) {
@@ -256,16 +256,20 @@ Rectangle {
                                 Rectangle {
                                     id: dragItem
                                     property var trackHead: trackHead
+                                    anchors.left: dragMouseArea.left
+                                    anchors.top: dragMouseArea.top
+                                    anchors.bottom: dragMouseArea.bottom
+                                    width: dragMouseArea.width
                                     color: 'white'
                                     Drag.active: dragMouseArea.drag.active
-                                    Drag.hotSpot.x: width / 2
                                     Drag.hotSpot.y: height / 2
                                     Drag.keys: ["trackHeader"]
                                     opacity: dragMouseArea.containsMouse | dragMouseArea.drag.active ? 0.7 : 0
                                     Behavior on opacity { NumberAnimation {} }
                                     Text {
+                                        id: dragItemText
                                         anchors.fill: parent
-                                        text: trackHead.trackName
+                                        text: qsTr("Move %1").arg(trackHead.trackName)
                                         style: Text.Outline
                                         styleColor: 'white'
                                         font.pixelSize: Math.min(Math.max(parent.height * 0.7, 15), 30)
@@ -274,14 +278,12 @@ Rectangle {
                                     }
                                     states: [
                                         State {
-                                            when: dragMouseArea.drag.active
+                                            when: dragMouseArea.drag.active | dragMouseArea.containsMouse
                                             ParentChange { target: dragItem; parent: trackHeaders.parent }
-                                            PropertyChanges { target: dragItem; anchors.fill: undefined }
                                         },
                                         State {
-                                            when: !dragMouseArea.drag.active
+                                            when: !dragMouseArea.drag.active & !dragMouseArea.containsMouse
                                             ParentChange { target: dragItem; parent: trackHead }
-                                            PropertyChanges { target: dragItem; anchors.fill: dragMouseArea }
                                         }
                                     ]
                                 }
```

image

ddennedy: That looks good to me. If you like it too, go ahead and commit it.

ddennedy:
One problem I found is that when the track name is more than like 10 characters long, the drag never becomes valid and shows a green target rectangle.

bmatherly:
> One problem I found is that when the track name is more than like 10 characters long, the drag never becomes valid and shows a green target rectangle.

Fixed. Additional comments welcome.

PR #1235, opened 2022-02-21
Run scope processing at low priority

bmatherly:
Can you test this and see if this provides improved performance?


ddennedy:
No, but thanks for trying. If you still want to merge it because you feel it is better behavior in general, then please change the bare thread pointer to unique/scoped.

PR #1236, opened 2022-02-22
Add tootip to timeline clips

bmatherly:
Show clip name, comment and duration

As suggested here:
https://forum.shotcut.org/t/show-clips-comments-in-the-timeline/32131/3

This uses the standard tooltip which has a delayed popup after hover

image

### DIFF
--- a/src/docks/timelinedock.cpp
+++ b/src/docks/timelinedock.cpp
@@ -1098,7 +1098,7 @@ bool TimelineDock::blockSelection(bool block)
 void TimelineDock::onProducerModified()
 {
     // The clip name may have changed.
-    emitSelectedChanged(QVector<int>() << MultitrackModel::NameRole);
+    emitSelectedChanged(QVector<int>() << MultitrackModel::NameRole << MultitrackModel::CommentRole);
 }
 
 void TimelineDock::replace(int trackIndex, int clipIndex, const QString& xml)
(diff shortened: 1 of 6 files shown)


ddennedy (changes requested):
There are a few problems with it:
1. It shows up for "blank"s
2. When I edit the comments in properties, the tooltip is not updated.
3. A long comment makes a very big tooltip and it should probably be elided while considering not only line length but number of lines.
3. It shows when dragging a clip along with the time "bubble help", which is distracting as they are in different places and look different.

Consider adding an option in the timeline menu, options sub-menu that defaults to on.

bmatherly:
> It shows up for "blank"s

fixed

> When I edit the comments in properties, the tooltip is not updated.

fixed

> A long comment makes a very big tooltip and it should probably be elided while considering not only line length but number of lines.

I am not sure what to do about this. The ToolTip control does not support elided text. I am thinking about writing a javascript function to manually truncate/elide lines that are longer than 50 characters and to limit the comment to 3 lines.

Thoughts?

> It shows when dragging a clip along with the time "bubble help", which is distracting as they are in different places and look different.

fixed

> Consider adding an option in the timeline menu, options sub-menu that defaults to on.

I am not sure if this is worthy of a menu option. Do you think that some users will not like this feature? If so, maybe we should just reject the idea. I do not feel strongly about the idea. It just seems like an easy thing to change.

ddennedy:
I think the eliding function idea is good. We can leave the menu option out for now and add it later if it seems desirable.

bmatherly:
I think this was closed by accident?

ddennedy:
Yes an accident

bmatherly:
I added a manual "elide" to the comment to limit the size.

image

ddennedy (approved):

PR #1239, opened 2022-03-06
Implement new box blur filter based on new MLT box_blur

bmatherly:
Hide the old filter for backwards compatibility.

Depends on this: https://github.com/mltframework/mlt/pull/776

Provides h/v values in the range of 0-100% (which really maps to a radius of 0-10% of image width in the MLT filter)

image

### DIFF
--- a/src/qml/filters/blur/ui_box_blur_2.qml
+++ b/src/qml/filters/blur/ui_box_blur_2.qml
@@ -1,0 +1,151 @@
+/*
+ * Copyright (c) 2022 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


src/qml/filters/blur/ui_box_blur_2.qml
```diff
@@ -0,0 +1,151 @@
...
+                filter.resetProperty('vradius')
+            }
+            onPresetSelected: {
+                setControls()
+                initializeSimpleKeyframes()
+            }
+        }
+
+        Label {
+            text: qsTr('Horizontal Radius')
```
ddennedy, line 85: I have had to correct your labels a few times. Not every word of a field label is capitalized like the names of Actions (buttons and menu-items). See for reference https://shotcut.org/notes/ui-conventions/

ddennedy:
You do not need to do anything with my comment because I renamed these parameters to be the same as the old filter: Width and Height. Also, keep in mind in the future that the names of the Keyframes parameters should match the field labels.

bmatherly:
> You do not need to do anything with my comment because I renamed these parameters to be the same as the old filter: Width and Height. Also, keep in mind in the future that the names of the Keyframes parameters should match the field labels.

Thanks. I appreciate your attention to detail.

PR #1250, opened 2022-04-16
MainWindow Constructor Refactoring

DJE98:
I refactored the MainWindow Constructor for better readability.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -204,14 +204,7 @@ MainWindow::MainWindow()
 
     setupMenuView();
 
-    // connect video widget signals
-    Mlt::GLWidget* videoWidget = (Mlt::GLWidget*) &(MLT);
-    connect(videoWidget, SIGNAL(dragStarted()), m_playlistDock, SLOT(onPlayerDragStarted()));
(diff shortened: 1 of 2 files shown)


ddennedy:
Thank you for trying to contribute, but I am not yet convinced navigating back and forth between 11 new functions is more readable - especially as the code grows and changes making their placement within the file more distributed. That is why I am content with simply grouping with line breaks and comment headings. I do not blindly agree with "long method is bad" practice. To be honest, I have been meaning to refactor this class into separate areas of concerns. Things related to the project/document, profile/video-mode, and settings menu could be factored out, and this change could help or get in the way.

ddennedy:
Are you interested in trying to refactor the entire MainWindow class? If not what is something you are interested in doing?

DJE98:
Yes, I'm interested in refactoring the entire MainWindow class step by step. In my view, the constructor has some inconsistencies. Sometimes there were commented blocks and sometimes people has already introduced functions like setupSettingsMenu for familiar work. I understand your general idea behind it, but in a modern developmental environment you can easily step back and forth. So you can easily gain an overview and step into the functions later. Also, more granular functions are helping to separate concerns more easily later on.

ddennedy:
Thank you

PR #1254, opened 2022-04-30
Format code style using Artistic Style

bmatherly:
* Add Artistic Style project file (.astylerc)
* Use the file to format all source files

The Artistic Style project file conforms to the QT coding standards.
The project file parameters are recommended at the bottom of this
page:
https://wiki.qt.io/Qt_Coding_Style

To apply the formatting, run this command from the root folder:
   astyle --project=.astylerc --recursive "./*.cpp,*.h"

Posting this here for discussion. I do not feel strongly about this topic and it is OK with me if we discard this PR. But I notice that my code is generally non-conforming. So having a tool like this would help my coding style.

If we like this tool, I am willing to add some automation
 * Add to build scripts
 * Fail builds if code does not conform
 * Add cmake targets
 * etc


.astylerc
```diff
@@ -0,0 +1,11 @@
+--style=kr 
+--indent=spaces=4 
+--align-pointer=name
```
bmatherly, line 3: I dislike this one, but I am willing to comply for consistency. This parameter places the "*" by the variable name rather than the type. e.g: "const char *string"

CuteLogger/include/AbstractAppender.h
```diff
@@ -24,22 +24,24 @@
 
 class CUTELOGGERSHARED_EXPORT AbstractAppender
 {
-  public:
```
bmatherly, line 27: I can add an exception for the CuteLogger directory if we want.

src/spatialmedia/box.cpp
```diff
@@ -1,19 +1,19 @@
 /*****************************************************************************
```
bmatherly, line 1: I can add an exception for the spatialmedia directory if we want.

ddennedy:
I do not want to add this to the build scripts and fail the build. The build script now is primarily for nightly CI, they cannot commit, and they generally should not change the code or you get that update problem when using it for development. You can instead add a new workflow in GitHub Actions that is run whenever there is a commit pushed or a pull request to check the style and report something.

.astylerc
ddennedy: I started using the Qt style because that is the default in Qt Creator, and I quickly liked it. Why are you not using Qt Creator yet?! This is one place where I differed, and I think that came from another C++ convention I was trying to follow. I still prefer to align the asterisk with the identifier in MLT. I am willing to conform to the Qt and MLT standard here and OK with it as-is.

CuteLogger/include/AbstractAppender.h
ddennedy: Yes, please exclude CuteLogger.

src/spatialmedia/box.cpp
ddennedy: Yes, let's exclude spatialmedia as well.

.astylerc
bmatherly: > Why are you not using Qt Creator yet?!

Because I am old and stubborn - I have 20 years of muscle memory inertia in Eclipse. I probably need to take the hit and learn something new.

bmatherly:
> I do not want to add this to the build scripts and fail the build.

I agree with that. I do not have a strong preference about automation - I am just offering to implement something if you would like since it is an activity that I enjoy.

> The build script now is primarily for nightly CI, they cannot commit, and they generally should not change the code or you get that update problem when using it for development.

I also agree with this. You and I often submit directly to master without making a PR. So there is no opportunity to catch new code style infractions before the commit goes to master. Unless we want to change that policy and require staging PRs for all commits, I think we will just have occasional "style cleanup" commits. 

My primary plan is to run astyle before checking in code. As a result, I will probably submit code style fixes at arbitrary times.

ddennedy:
In addition to a GitHub Action to check code style on commit to master and pull request, you can setup a custom target in CMake (see codespell). Then, you can also add that as a build step in Qt Creator, or simply make that target periodically.

CuteLogger/include/AbstractAppender.h
bmatherly: Done

src/spatialmedia/box.cpp
bmatherly: Done

bmatherly:
> a GitHub Action to check code style on commit to master and pull request

I will probably work on this in the future. I will need to add astyle to the docker image.

> you can setup a custom target in CMake

This is done

> Then, you can also add that as a build step in Qt Creator, or simply make that target periodically.

If you are not already familiar, you can configure qtcreator to use the .astylerc file that is in the root of the source directory:
https://doc.qt.io/qtcreator/creator-beautifier.html

ddennedy:
>  I will need to add astyle to the docker image.

You do not need to run our docker image if you make a workflow simply to run astyle. Our image is heavy. You can study Actions to learn more. There might even be an action in the marketplace for it.

bmatherly:
> You do not need to run our docker image if you make a workflow simply to run astyle. Our image is heavy. You can study Actions to learn more. 

I created an action and added it to this PR. I am not sure why it did not run. Maybe because the PR already existed when I pushed it. https://github.com/mltframework/shotcut/pull/1254/commits/c6c3e3a1ec30f6e53986f097c437f85deb78841b

It worked in my test repository: https://github.com/bmatherly/shotcut/pull/1

ddennedy:
As a result of this I now get a bunch of these warnings :-(
https://github.com/KDE/clazy/blob/1.11/docs/checks/README-connect-not-normalized.md

bmatherly:
> As a result of this I now get a bunch of these warnings :-(
> https://github.com/KDE/clazy/blob/1.11/docs/checks/README-connect-not-normalized.md

I notice that the QT coding style automatically wraps lines if they exceed a maximum limit. I also notice that some connect() function calls were broken into multiple lines in this change. Maybe clazy does not handle connect() calls that are line wrapped.

PR #1256, opened 2022-05-01
Format QML code style using qmlformat

bmatherly:
Add a cmake custom target to fix QML formatting
Apply the formatting to all qml files

qmlformat applies the [QT QML Coding Conventions](https://doc-snapshots.qt.io/qt6-dev/qml-codingconventions.html)

In general, I think this makes the code more verbose by expanding many things that are currently on one line line.

I am willing to add a GitHub action job for this if we are interested. I am also OK discarding this PR - I do not feel strongly about it. But I offer this here in case it is interesting to pursue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,3 +67,8 @@ add_custom_target(astyle COMMAND
   astyle --project=.astylerc --ignore-exclude-errors --recursive "./*.cpp,*.h"
 )
 
+file(GLOB_RECURSE QML_SRC "src/qml/*.qml")
+add_custom_target(qmlformat COMMAND
+  ${CMAKE_PREFIX_PATH}/bin/qmlformat -i ${QML_SRC}
+)
+
--- a/src/qml/filters/alpha_adjust/ui.qml
+++ b/src/qml/filters/alpha_adjust/ui.qml
@@ -15,36 +15,37 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+import QtQml.Models 2.12
 import QtQuick 2.12
 import QtQuick.Controls 2.12
 import QtQuick.Layouts 1.12
 import Shotcut.Controls 1.0 as Shotcut
-import QtQml.Models 2.12
 
 Item {
-    width: 200
-    height: 50
     property string paramOperation: '2'
     property string paramThreshold: '3'
     property string paramAmount: '4'
     property string paramInvert: '5'
 
+    width: 200
+    height: 50
     Component.onCompleted: {
         if (filter.isNew) {
-            filter.set(paramOperation, 0)
-            filter.set(paramThreshold, 0.5)
-            filter.set(paramAmount, 0.5)
+            filter.set(paramOperation, 0);
+            filter.set(paramThreshold, 0.5);
+            filter.set(paramAmount, 0.5);
         }
-        var current = filter.getDouble(paramOperation)
+        var current = filter.getDouble(paramOperation);
         for (var i = 0; i < operationModel.count; ++i) {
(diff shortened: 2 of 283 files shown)


CMakeLists.txt
```diff
@@ -57,3 +57,8 @@ add_custom_target(codespell COMMAND
   -S export-edl.js,CuteLogger,drmingw,node_modules,moc_*,index.theme,*.ts,three.min.js,three.js,jquery.js,rangy-*.js,Makefile,shotcut.pro,*.desktop
 )
 
+add_custom_target(qmlformat COMMAND
+  bash -c [[for f in $(find . -name *.qml); do ~/Qt/5.15.2/gcc_64/bin/qmlformat -i ${f}; done]]
```
bmatherly, line 61: I would need to find a better way to locate qmlformat. Maybe there is an environment variable that CMake has access to.

src/qml/filters/alpha_adjust/ui.qml
```diff
@@ -15,36 +15,37 @@
...
     property string paramInvert: '5'
 
+    width: 200
+    height: 50
     Component.onCompleted: {
         if (filter.isNew) {
-            filter.set(paramOperation, 0)
-            filter.set(paramThreshold, 0.5)
-            filter.set(paramAmount, 0.5)
+            filter.set(paramOperation, 0);
```
bmatherly, line 34: Looking through the changes, I like most things. The change that would require the most adjustment for me is adding semicolons after function calls.

ddennedy:
The thing I am most worried about is if the code reorganization breaks many translations because the line delta is too high. Secondarily, I know QML is rather flexible, and I did not follow all conventions to make things the way I prefer. It is a pretty big change; I don't know yet.

bmatherly:
> The thing I am most worried about is if the code reorganization breaks many translations because the line delta is too high. 

The new action changes are going to create a lot of translation changes. Maybe this would be a good time to also take this hit.

> Secondarily, I know QML is rather flexible, and I did not follow all conventions to make things the way I prefer. It is a pretty big change; I don't know yet.

This is a still a fair comment, and I do not feel strongly.

ddennedy:
You made a good point regarding translations. Did you use the `--normalize` option?

CMakeLists.txt
ddennedy, line 61: There must be a better way to get this path. If not from a CMake variable, use environment variable `$QTDIR/bin/qmlformat`. Also, in `src/CMakeLists.txt` there is a line `file(GLOB_RECURSE QML_SRC "qml/*")` that finds all the QML sources instead of relying on bash and find, which is not going to work on Windows.

bmatherly:
> Did you use the --normalize option?

I do not see a normalize option

```
Usage: /home/brian/Qt/5.15.2/gcc_64/bin/qmlformat [options] filenames
Formats QML files according to the QML Coding Conventions.

Options:
  -h, --help               Displays help on commandline options.
  --help-all               Displays help including Qt specific options.
  -v, --version            Displays version information.
  -V, --verbose            Verbose mode. Outputs more detailed information.
  -n, --no-sort            Do not sort imports.
  -i, --inplace            Edit file in-place instead of outputting to stdout.
  -f, --force              Continue even if an error has occurred.
  -l, --newline <newline>  Override the new line format to use (native macos
                           unix windows).

Arguments:
  filenames                files to be processed by qmlformat

```

CMakeLists.txt
bmatherly: I am willing to improve this (if you have not done it already). I had mostly posted this for discussion and consideration. If you think we will go forward, I'll spiff this up.

bmatherly: Done

PR #1258, opened 2022-05-02
Support keyframes in lowpass audio filter

rkfg:
Discussion: https://forum.shotcut.org/t/audio-filter-keyframes/31628

This PR implements keyframes support for lowpass filter parameters. If it looks and works fine I can do the same to the other audio filters that don't support animation yet.

### DIFF
--- a/src/qml/filters/audio_lowpass/ui.qml
+++ b/src/qml/filters/audio_lowpass/ui.qml
@@ -21,31 +21,54 @@ import QtQuick.Controls 2.12
 import QtQuick.Layouts 1.12
 import Shotcut.Controls 1.0 as Shotcut
 
-Item {
+Shotcut.KeyframableFilter {
+    property string cutoffProperty: '0'
+    property string stagesProperty: '1'
+    property string wetnessProperty: 'wetness'
     width: 350
     height: 125
+    keyframableParameters: preset.parameters
+    startValues: [2637, 1, 0]
+    middleValues: [2637, 1, 1]
+    endValues: [2637, 1, 0]
     Component.onCompleted: {
         if (filter.isNew) {
             // Set default parameter values
-            filter.set('0', 2637)
-            filter.set('1', 1)
-            filter.set('wetness', 1.0)
+            filter.set(cutoffProperty, 2637)
+            filter.set(stagesProperty, 1)
+            filter.set(wetnessProperty, 1.0)
             filter.savePreset(preset.parameters)
         }
         setControls()
     }
 
     function setControls() {
-        sliderCutoff.value = filter.getDouble('0')
-        sliderStages.value = filter.get('1')
-        sliderWetness.value = filter.getDouble('wetness') * sliderWetness.maximumValue
+        var position = getPosition()
+        blockUpdate = true
+        sliderCutoff.value = filter.getDouble(cutoffProperty, position)
+        cutoffKeyframesButton.checked = filter.animateIn <= 0 && filter.animateOut <= 0 && filter.keyframeCount(cutoffProperty) > 0
+        sliderStages.value = filter.getDouble(stagesProperty, position)
+        stagesKeyframesButton.checked = filter.animateIn <= 0 && filter.animateOut <= 0 && filter.keyframeCount(stagesProperty) > 0
+        sliderWetness.value = filter.getDouble(wetnessProperty, position) * sliderWetness.maximumValue
+        wetnessKeyframesButton.checked = filter.animateIn <= 0 && filter.animateOut <= 0 && filter.keyframeCount(wetnessProperty) > 0
+        blockUpdate = false
+        enableControls(isSimpleKeyframesActive())
+    }
 
+    function enableControls(enabled) {
+        sliderCutoff.enabled = sliderStages.enabled = sliderWetness.enabled = enabled
+    }
+
+    function updateSimpleKeyframes() {
+        updateFilter(cutoffProperty, sliderCutoff.value, cutoffKeyframesButton, null)
+        updateFilter(stagesProperty, sliderStages.value, stagesKeyframesButton, null)
+        updateFilter(wetnessProperty, sliderWetness.value / sliderWetness.maximumValue, wetnessKeyframesButton, null)
     }
 
     GridLayout {
         anchors.fill: parent
         anchors.margins: 8
-        columns: 3
+        columns: 4
 
 
         Label {
@@ -54,9 +77,15 @@ Item {
         }
         Shotcut.Preset {
             id: preset
-            parameters: ['0', '1', '2', 'wetness']
-            Layout.columnSpan: 2
-            onPresetSelected: setControls()
+            parameters: [cutoffProperty, stagesProperty, wetnessProperty]
+            Layout.columnSpan: parent.columns - 1
+            onBeforePresetLoaded: {
+                resetSimpleKeyframes()
+            }
+            onPresetSelected: {
+                setControls()
+                initializeSimpleKeyframes()
+            }
         }
 
         Label {
@@ -69,14 +98,18 @@ Item {
             minimumValue: 5
             maximumValue: 21600
             suffix: ' Hz'
-            value: filter.getDouble('0')
-            onValueChanged: {
-                filter.set('0', value)
-            }
+            onValueChanged: updateFilter(cutoffProperty, value, cutoffKeyframesButton, getPosition())
         }
         Shotcut.UndoButton {
-                onClicked: sliderCutoff.value = 2637
+            onClicked: sliderCutoff.value = 2637
+        }
+        Shotcut.KeyframesButton {
+            id: cutoffKeyframesButton
+            onToggled: {
+                enableControls(true)
+                toggleKeyframes(checked, cutoffProperty, sliderCutoff.value)
             }
+        }
 
         Label {
             text: qsTr('Rolloff rate')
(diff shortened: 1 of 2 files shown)


bmatherly:
Nice work. I tested this and I do not see any bugs. A couple of comments for consideration:

The rolloff control works in integer increments. When you try to manipulate the slider it "jumps" from one integer to the next. But when you animate the keyframes, the animation will allow the slider to go to values  between integers (even though the text box only shows integers). This is a little strange because the keyframes can set values that the user can not set. I wonder if we should just make the control a double with one decimal.

After using this filter, I think that all three parameters should be controlled by the simple keyframes. There is no disadvantage to this. If the user only wants to animate the wetness, they can choose to only change the wetness control.

ddennedy:
> I think that all three parameters should be controlled by the simple keyframes. There is no disadvantage to this.

We talked about this on the forum. I had suggested that the convention is that simple keyframes controls all keyframable parameters. However, this and some other audio filters has a "wetness" parameter that is special. It mixes the modified signal with the filter's input. There is one (small) disadvantage. With a wetness parameter, it can be convenient to setup the other parameters as needed statically and then use the simple keyframes to kind of fade in and out of the effect. This is facilitated by making the start and end defaults equal 0 wetness. What do you think considering that use case?

src/qml/filters/audio_lowpass/ui.qml
```diff
@@ -21,9 +21,10 @@ import QtQuick.Controls 2.12
 import QtQuick.Layouts 1.12
 import Shotcut.Controls 1.0 as Shotcut
 
-Item {
+Shotcut.KeyframableFilter {
     width: 350
     height: 125
+    keyframableParameters: ['0', '1', 'wetness']
```
ddennedy, line 27: Since you are setting this variable from `KeyframableFilter` you should also initialize `startValues`, `middleValues`, and `endValues`. This especially important to give convenient default values for simple keyframes.

ddennedy: Also, usually when the code gets more complicated like this, and MLT property names are abstract like `0` and `1` we like to make QML properties to give them a nicer name:
```
property string cutoffProperty: '0'
property string stagesProperty: '1'
```

ddennedy: Another thing I often do is since they are equivalent:
`keyframableParameters: preset.parameters`

src/qml/filters/audio_lowpass/ui.qml
```diff
@@ -54,9 +69,17 @@ Item {
...
             id: preset
-            parameters: ['0', '1', '2', 'wetness']
-            Layout.columnSpan: 2
-            onPresetSelected: setControls()
+            parameters: ['0', '1', 'wetness']
+            Layout.columnSpan: parent.columns - 1
+            onBeforePresetLoaded: {
+                filter.resetProperty('0')
+                filter.resetProperty('1')
+                filter.resetProperty('wetness')
```
ddennedy, line 77: You should replace these lines with `resetSimpleKeyframes()` from `KeyframableFilters`. Semantically they are not all simple keyframes in this filter, but it does the same thing by iterating over `keyframableParameters`.

ddennedy (changes requested):

ddennedy:
> The rolloff control works in integer increments.

The ladspa parameter is integers [1, 10]. See `melt -query filter=ladspa.1891`.

> the animation will allow the slider to go to values between integers (even though the text box only shows integers). This is a little strange

That is a moderate problem between curves and our SliderSpinner with a very small range of integers. 

> I wonder if we should just make the control a double with one decimal.

Since it is inherently an integer parameter, I do not think we should do this. I think we should either live with the quirk or remove `isCurve: true`.

src/qml/filters/audio_lowpass/ui.qml
rkfg: > Since you are setting this variable from `KeyframableFilter` you should also initialize `startValues`, `middleValues`, and `endValues`. This especially important to give convenient default values for simple keyframes. I recommend defaults for wetness to be start=0 middle=1 and end=1.

I'll add those parameters but I don't understand the "defaults" part. It seems that these values are initialized by `KeyframableFilter` anyway if simple keyframes are active: https://github.com/mltframework/shotcut/blob/493c9c0373a462c3e0f3b158876867608c3a44d1/src/qml/modules/Shotcut/Controls/KeyframableFilter.qml#L81-L88
I tried to set the values to
```
    startValues: [0, 0, 0]
    middleValues: [0, 0, 1]
    endValues: [0, 0, 1]
```
and
```
    startValues: [0, 0, 0]
    middleValues: [0, 0, 0]
    endValues: [0, 0, 0]
```
and I don't see any difference. So probably simply initializing the arrays is enough?

rkfg:
All fixed I hope. I also replaced `'wetness'` with a property constant for uniformity.

src/qml/filters/audio_lowpass/ui.qml
```diff
@@ -21,31 +21,52 @@ import QtQuick.Controls 2.12
...
-Item {
+Shotcut.KeyframableFilter {
+    property string cutoffProperty: '0'
+    property string stagesProperty: '1'
+    property string wetnessProperty: 'wetness'
     width: 350
     height: 125
+    keyframableParameters: preset.parameters
+    startValues: [0, 0, 0]
+    middleValues: [0, 0, 0]
```
ddennedy, line 32: The middle value for wetness should be 1 to allow simple keyframes to fade in and/or out wetness by default.

src/qml/filters/audio_lowpass/ui.qml
ddennedy: No, no other change is required. I just tested with the following, and it works as I intended:
```
    startValues: [0, 0, 0]
    middleValues: [0, 0, 1]
    endValues: [0, 0, 0]
```

rkfg: Okay, it's a bit hard to understand from the code. I'll change it then!

src/qml/filters/audio_lowpass/ui.qml
rkfg: Should be good now!

ddennedy (approved):

bmatherly:
> With a wetness parameter, it can be convenient to setup the other parameters as needed statically and then use the simple keyframes to kind of fade in and out of the effect. This is facilitated by making the start and end defaults equal 0 wetness. What do you think considering that use case?

It is fine. I do not feel strongly about the final implementation. I think that in general we lack consistency and documentation in what the simple keyframes control. So it is up to each user to experiment and discover how they work from one filter to another. I was thinking that we could move toward a future where all keyframable parameters are part of the simple keyframes.

ddennedy:
> I was thinking that we could move toward a future where all keyframable parameters are part of the simple keyframes.

i thought I had read that you did a review of all filters and found this to be the case. Excluding this one, is my recollection correct?

bmatherly:
> i thought I had read that you did a review of all filters and found this to be the case. 

Thanks for the reminder. I just double checked and I find that to be true. The only exception is the gradient filter which supports simple keyframes, but no advanced keyframes.

So I guess we are in a more consistent state than I thought - and this filter would move us away from that. 

Again, I do not feel strongly. I just wanted to make sure we were considering the "consistency" angle.

src/qml/filters/audio_lowpass/ui.qml
```diff
@@ -21,31 +21,52 @@ import QtQuick.Controls 2.12
...
+        blockUpdate = false
+        enableControls(isSimpleKeyframesActive())
+    }
+
+    function enableControls(enabled) {
+        sliderCutoff.enabled = sliderStages.enabled = sliderWetness.enabled = enabled
+    }
 
+    function updateSimpleKeyframes() {
+        updateFilter(wetnessProperty, sliderWetness.value / sliderWetness.maximumValue, wetnessKeyframesButton, null)
```
ddennedy, line 63: I decided that we need to include all keyframable params in simple params. The consistency is more important than some nifty mode of operation. Besides, with the defaults, it might still fade in and out nicely with even if you forget to update the frequency at start and end.

src/qml/filters/audio_lowpass/ui.qml
```diff
@@ -21,31 +21,52 @@ import QtQuick.Controls 2.12
...
+Shotcut.KeyframableFilter {
+    property string cutoffProperty: '0'
+    property string stagesProperty: '1'
+    property string wetnessProperty: 'wetness'
     width: 350
     height: 125
+    keyframableParameters: preset.parameters
+    startValues: [0, 0, 0]
+    middleValues: [0, 0, 1]
+    endValues: [0, 0, 0]
```
ddennedy, line 33: The initial values for the non-wetness params in these arrays should be the same as the defaults below and thus:
```
     startValues: [2637, 1, 0]
     middleValues: [2637, 1, 1]
     endValues: [2637, 1, 0]
```

ddennedy (changes requested):

rkfg:
Thanks, now I understand those simple keyframes and all the associated parameters. Really, they're much more complex than the regular keyframes! Changes pushed, I tried different combinations with default start/middle/end values right after the filter is added and also modified them manually in the project. They animate nicely now and I fully agree that the default behavior now works as expected + it's possible to animate all parameters at once changing them at different positions.

PR #1259, opened 2022-05-02
Increase non-curve keyframe track height

rkfg:
I'm not 100% sure it's not just my issue but I tried my usual scaling (`QT_SCALE_FACTOR=1.2` as I have a 2k 24" monitor), the default 1 and also 2 (too big for my monitor but just to confirm it), it looks the same every time. I use Debian Testing, not sure if it's like that on Windows.

The problem is that non-curve keyframe tracks have fixed height that cuts the low part of the icons. This height can't be increased by the user, it looks bad:

This PR increases the height so it looks better:

### DIFF
--- a/src/qml/views/keyframes/Keyframes.js
+++ b/src/qml/views/keyframes/Keyframes.js
@@ -16,7 +16,7 @@
  */
 
 function trackHeight(isCurves) {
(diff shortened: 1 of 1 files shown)


ddennedy:
thanks

PR #1263, opened 2022-05-07
Support keyframes in highpass audio filter

rkfg:
This is almost a 100% copy of #1258 except the default start/middle/end parameters. Makes the highpass audio filter keyframeable.

### DIFF
--- a/src/qml/filters/audio_highpass/meta.qml
+++ b/src/qml/filters/audio_highpass/meta.qml
@@ -7,4 +7,33 @@ Metadata {
     name: qsTr("High Pass")
     mlt_service: 'ladspa.1890'
     qml: 'ui.qml'
+    keyframes {
+        allowAnimateIn: true
+        allowAnimateOut: true
+        simpleProperties: ['0', '1', 'wetness']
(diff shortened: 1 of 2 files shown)


src/qml/filters/audio_highpass/meta.qml
```diff
@@ -7,4 +7,33 @@ Metadata {
     name: qsTr("High Pass")
     mlt_service: 'ladspa.1890'
     qml: 'ui.qml'
+    keyframes {
+        allowAnimateIn: true
+        allowAnimateOut: true
+        simpleProperties: ['wetness']
```
ddennedy, line 13: This needs to list all of the simple keyframes properties. Looks like I overlooked that in the low pass filter. I will fix that, but you need to fix it for these new PRs.

ddennedy (changes requested):

src/qml/filters/audio_highpass/meta.qml
rkfg: Should be good to go now!

PR #1264, opened 2022-05-07
Support keyframes in reverb audio filter

rkfg:
This filter animation is based on my previous experience. However, it doesn't have the wetness parameter so for simple keyframes I used the minimum/default/minimum values for all sliders. It doesn't sound nice to be honest but for the sake of consistency the defaults are like that. The rest is straightforward.


rkfg:
Anything I should fix here?

PR #1278, opened 2022-05-30
Qml deprecations

bmatherly:
Significantly cleans up QML errors/warnings

Staging this here in case we want to wait for the next beta period.

### DIFF
--- a/src/qml/modules/Shotcut/Controls/GradientControl.qml
+++ b/src/qml/modules/Shotcut/Controls/GradientControl.qml
@@ -47,7 +47,7 @@ RowLayout {
                stopIndex == colorList.length - 1 ? parent.width - width :
                stopIndex * (parent.width / (colorList.length - 1)) - (width / 2)
             y: 0
-            color: colorList[stopIndex] != undefined ? colorList[stopIndex] : "gray"
+            color: typeof colorList[stopIndex] !== 'undefined' ? colorList[stopIndex] : "gray"
             border.color: "gray"
             border.width: 1
             width: 10
--- a/src/qml/modules/Shotcut/Controls/GradientControl.qml
+++ b/src/qml/modules/Shotcut/Controls/GradientControl.qml
@@ -47,7 +47,7 @@ RowLayout {
                stopIndex == colorList.length - 1 ? parent.width - width :
                stopIndex * (parent.width / (colorList.length - 1)) - (width / 2)
             y: 0
-            color: colorList[stopIndex]
+            color: colorList[stopIndex] != undefined ? colorList[stopIndex] : "gray"
             border.color: "gray"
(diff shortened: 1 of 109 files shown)


src/qml/modules/Shotcut/Controls/GradientControl.qml
```diff
@@ -47,7 +47,7 @@ RowLayout {
                stopIndex == colorList.length - 1 ? parent.width - width :
                stopIndex * (parent.width / (colorList.length - 1)) - (width / 2)
             y: 0
-            color: colorList[stopIndex]
+            color: colorList[stopIndex] != undefined ? colorList[stopIndex] : "gray"
```
ddennedy, line 50: `undefined` is not a keyword in JavaScript, and maybe it could be defined as a global variable somewhere to something with a type. Moreover, this is not using a strict comparison operator `!==`. This should be 
`typeof colorList[stopIndex] !== 'undefined' ? colorList[stopIndex] : "gray"`

You can search our code to see where I have used it previously.

src/qml/views/timeline/Track.qml
```diff
@@ -66,25 +66,25 @@ Rectangle {
...
-            fadeIn: model.fadeIn
-            fadeOut: model.fadeOut
-            hash: model.hash
-            speed: model.speed
-            audioIndex: model.audioIndex
+            fadeIn: model.fadeIn != undefined ? model.fadeIn : 0
+            fadeOut: model.fadeOut != undefined ? model.fadeOut : 0
+            hash: model.hash != undefined ? model.hash : 0
+            speed: model.speed != undefined ? model.speed : 1
+            audioIndex: model.audioindex != undefined ? model.audioIndex : 0
```
ddennedy, line 87: See all these usages of `undefined` here as well.

src/qml/views/timeline/timeline.qml
```diff
@@ -470,11 +470,11 @@ Rectangle {
                                 model: timeline.selection
                                 Rectangle {
                                     property var clip: trackAt(modelData.y).clipAt(modelData.x)
-                                    property var track: trackAt(clip.trackIndex + dragDelta.y)
-                                    x: clip.x + dragDelta.x
-                                    y: track.y
+                                    property var track: clip == undefined ? trackAt(clip.trackIndex + dragDelta.y) : 0
+                                    x: clip == undefined ? clip.x + dragDelta.x : 0
```
ddennedy, line 474: more `undefined`

ddennedy (changes requested):

src/qml/modules/Shotcut/Controls/GradientControl.qml
bmatherly: done

PR #1280, opened 2022-06-02
Fix filters disappear when changing clip properties

bmatherly:
This happens to clips on the timeline when a clip property is
changed.

As reported here:
https://forum.shotcut.org/t/please-test-the-beta-for-version-22-06/33941/13

However, the problem occurs with any filter, not only Crop:Source

Steps to reproduce:
1) Open a clip and drag to the timeline
2) Add a filter to the clip
3) In the clip properties panel, change the duration or the speed
4) Look at the properties panel and observe that the filter is still applied
5) Click the "Undo" arrow
6) Look at the properties panel and observe that the filter is still missing   <--- BUG

This change fixes this specific test case for me. However, I am posting this as a a review because I do not know why the logic was written that way originally.

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -1312,12 +1312,10 @@ UpdateCommand::UpdateCommand(TimelineDock &timeline, int trackIndex, int clipInd
     , m_trackIndex(trackIndex)
     , m_clipIndex(clipIndex)
     , m_position(position)
-    , m_isFirstRedo(true)
     , m_undoHelper(*timeline.model())
     , m_ripple(Settings.timelineRipple())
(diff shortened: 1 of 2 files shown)


ddennedy (changes requested):
Gave this a quick test by opening a timeline project, selecting a clip, change Properties > Aspect ratio, and Undo did not undo that change. The problem is that the project has already changed by the time called `m_undoHelper.recordBeforeState()` in `UpdateCommand::redo()`. So, it is not really recording the before state. The UpdateCommand is rather special how it works has a member variable in `TimelineDock`.

ddennedy:
I fixed this in f5570b7f5

PR #1283, opened 2022-06-04
Add speed range option for audio alignment adjustment

bmatherly:
Is it too late to add this before the release candidate? A forum user made a good case for this additional user configuration (Speed adjustment range). It adds two translation strings.

image

### DIFF
--- a/src/dialogs/alignaudiodialog.cpp
+++ b/src/dialogs/alignaudiodialog.cpp
@@ -39,6 +39,7 @@
 #include <QHeaderView>
 #include <QIcon>
 #include <QLabel>
+#include <QLocale>
 #include <QPainter>
 #include <QPushButton>
 #include <QStyledItemDelegate>
@@ -110,13 +111,11 @@ class ClipAudioReader : public QObject
 {
     Q_OBJECT
 public:
-    ClipAudioReader(QString producerXml, AlignmentArray &referenceArray, int index, int in, int out,
-                    bool calculateSpeed)
+    ClipAudioReader(QString producerXml, AlignmentArray &referenceArray, int index, int in, int out)
         : QObject()
         , m_referenceArray(referenceArray)
         , m_reader(producerXml, &m_clipArray, in, out)
         , m_index(index)
-        , m_calculateSpeed(calculateSpeed)
     {
         connect(&m_reader, SIGNAL(progressUpdate(int)), this, SLOT(onReaderProgressUpdate(int)));
     }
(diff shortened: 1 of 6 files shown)


src/dialogs/alignaudiodialog.cpp
```diff
@@ -246,6 +245,26 @@ AlignAudioDialog::AlignAudioDialog(QString title, MultitrackModel *model,
...
     glayout->addWidget(m_trackCombo, row++, 1, Qt::AlignLeft);
+    // Speed combo box
+    glayout->addWidget(new QLabel(tr("Speed adjustment range")), row, 0, Qt::AlignRight);
+    m_speedCombo = new QComboBox();
+    m_speedCombo->setToolTip("Larger speed adjustment ranges take longer to process.");
+    m_speedCombo->addItem("0%", QVariant(0));
+    m_speedCombo->addItem("0.1%", QVariant(0.001));
+    m_speedCombo->addItem("0.5%", QVariant(0.005));
+    m_speedCombo->addItem("1%", QVariant(0.01));
+    m_speedCombo->addItem("5%", QVariant(0.05));
```
ddennedy, line 256: Why are these strings in a combo? A small problem is that they are not locale sensitive, but all other numeric values in Shotcut UI are. But also, I am curious why not a spinner with range [0, 5]?

ddennedy:
To answer your question: it is not too late.

src/dialogs/alignaudiodialog.cpp
bmatherly: Good point about the locale. I can fix that.

I am open minded about a spinner vs combo. I need to give options for a wide range and a narrow range. But I do not want users to labor over whether they should choose 1.1% or 1.2%. What I really want is just some vague ranges like "None", "Narrow", "Normal", "Wide", "Very Wide". But people will need to know what range those actually map to in order to decide if the range will work for their use case (the 5% is to support PAL to Film conversion).

Maybe I could make the strings use natural language with a percent in parenthesis:
* None (0%)
* Narrow (0.1%)
* Normal (0.5%)
* Wide (1%)
* Very Wide (5%)

Those strings could be translatable to cover the locale. Or I could construct the string from translated string and locale conversion.

ddennedy: I do think the word descriptions will help. Please do like `tr("Narrow") + QString(" (%L1%)").arg(0.1)` to keep the numeric stuff out of the translations. I am trying to prevent introducing things into the translatable strings that translators can mess up. Other languages might have different symbols and conventions for parenthesis and percent, but nearly elsewhere in the code we do not make the % translatable.

bmatherly: Thanks for the example. That worked perfectly.

PR #1284, opened 2022-06-06
defish0r plugin UI

hypov8:
updated UI filter for defish0r plugin.

 * -changed Interpolation quality slider to integer
 * -changed UI presets(scale + A/R) from slider to radio buttons
 * -user set sliders are hidden when scale or A/R presets are used
 * -added buttons for pre-configured cameras. eg gopro..
 * -changed UI names to be more intuitive
 * -added UI values into global varables, incase plugin changes.

original source from forum user TwitchyMcJoe. 
https://forum.shotcut.org/t/using-defish0r-as-a-filter-code-for-a-useful-defisheye-effect/181

i could not join forum for some reason. so added it here.

files posted here in case i have done something wrong
https://github.com/hypov8/shotcut_fisheye


ddennedy:
Thank you for your contribution, but there are quite a few issues with this code but there are now 2 lens-oriented filters in libavfilter that ought to be considered instead of this frei0r plugin.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,650 @@
...
+        var i = 0
+
+        do {
+            v_ret = fn_getCameraListName(i, iDX_CAMERA)
+            if (v_ret !== ''){
+                if (a_list.indexOf(v_ret) === -1)
+                    a_list.push(v_ret) //append camera  if unique
+            }
+            else
+                v_more = false
```
ddennedy, line 206: Use braces for the `else` block if you needed them for the then clause.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,650 @@
...
+            onValueChanged: filter.set(idx0_amount, value)
+        }
+        Shotcut.UndoButton {
+            onClicked: uID0_ratioSlider.value = v0_amount
+        }
+
+        // Row 4: Interpolation Quality
+        Label {
+            text: qsTr('Quality')
+            Shotcut.HoverTip { text: qsTr('Resample quality\n 1=no interpolation\n 2=fast linear\n 7=slowest') }
```
ddennedy, line 437: When you have to enumerate values in a tooltip that means this field should be a combo and not a numeric input.

ddennedy (changes requested):

hypov8:
is the  menu on the right better. using presets
radiobuttons with labels take up to much space, so i used combo

ui_v2_01

src/qml/filters/fisheye/meta.qml
```diff
@@ -0,0 +1,27 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+ 
+
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+    type: Metadata.Filter
+    name: qsTr("DeFisheye")
```
ddennedy, line 24: Thanks for all of your updates. Let's make another round of some cosmetic improvements. "Defisheye" is not really a word. Change this to "Fisheye Remover" please.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,655 @@
...
+            //
+            //Focused
+            // close match gopro studio, trimed left/right edges, centre zoom slightly, makes motion look faster
+            //
+            //Linear
+            // minimal distortion at frame centre. edges have increased croped to suit
+
+
+            //HERO3
+            case  0: return [0.736, true, lensValue(4), scaleValue(1), 0.500, qualityValue(2), aspectValue(1), 0.500, 'HERO-3 1080', 'Wide', 'focused'] //todo: language?
```
ddennedy, line 77: Change all of the "HERO-3" to simply "HERO3" and so on.
After we decide about presets make words like "Wide" translatable with `qsTr()`.
Change "focused" to "Focus" to be consistent?

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,655 @@
...
+                id: scaleModel 
+                ListElement { text: qsTr('Scale to Fill'); value: 0.0 }
+                ListElement { text: qsTr('Keep Center Scale'); value: 0.333 }
+                ListElement { text: qsTr('Scale to Fit'); value: 0.666 }
+                ListElement { text: qsTr('Manual Scale'); value: 1.0 }
+            }
+            textRole: 'text'
+            onActivated: {
+                filter.set(scaePresetParam, scaleModel.get(currentIndex).value)
+                scaleShowSlider = ((currentIndex+1)==scaleCombo.count)? true : false //show user input?
```
ddennedy, line 501: More of these unnecessary `? true : false` things. When I ask for a change please apply it everywhere.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,655 @@
...
+        }
+        Shotcut.Preset {
+            id: preset
+            parameters: defaultParameters
+            onPresetSelected: setControlData()
+        }
+        Row {width: 6} //blank
+
+        //line seperator #1
+        Rectangle{Layout.columnSpan: 3; Layout.fillWidth: true; height: 1; color: activePalette.text; opacity: 0.3}
```
ddennedy, line 365: Remove this horizontal line. There are way too many lines in this UI. Lines are visual distraction. I allow some below because there is preset vs. manual so those rows are grouped.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,655 @@
...
+                text: qsTr('Apply')
+                implicitWidth: 80
+                onClicked: setCameraData()
+            }
+        }
+
+        //line seperator #5
+        Rectangle{
+            visible: cameraShowNew
+            Layout.columnSpan: 3; Layout.fillWidth: true; height: 1; color: activePalette.text; opacity: 0.3}
```
ddennedy, line 649: Remove this horizontal line.

ddennedy:
> is the menu on the right better. using presets

Hmm, if this is the only number of combinations, I prefer the presets. However, once you add more cameras it may be too much as a flat list. But honestly, I will be very surprised if anyone adds more if you do not. We just do not get much contribution let alone presets, which is like the easiest contribution to make! Still thinking about it.

src/qml/filters/fisheye/ui.qml
hypov8: i did not realize i could just omit that and it would resolve as bool

hypov8:
i have been looking at avfilter. the 2 filters i found are lenscorrection and v360.
lenscorrection seems to be a duplicate, it seems to have some color shift bug here
i will see if v360 can do a better job on the gopro

not sure where this lensfun is, or how i can implement its camera database etc

ddennedy:
v360 is for spherical video.
FFmpeg needs to be compiled with the lensfun library in order to use it. It is more than just a database. [Here](https://www.danielplayfaircal.com/blogging/ffmpeg/lensfun/v360/lenscorrection/fisheye/dodgeball/2020/03/24/correcting-lens-distortion-with-ffmpeg.html) is a blog post about it. Adding a library has build cross-platform implications with respect to Shotcut. We can move forward on this for now.

hypov8:
i have been messing with the defish0r binary. i added a simplified 'elastic scale' filter to the plugin so superview stretch can be fixed.
this fixes the pinching in the middle of the frame.
no extra processing is needed and quality is less effected than running the 2 filters.
i also added a 'letterbox' option to clear the curved pixels at top/bottom
it still shows some distortion but some of it seems to be from a camera/image issue, not the filter 
do you think this would be useful? i would also need to update the UI.
i was also thinking to look at an independent scale (x and y), so height can be stretched to fit frame, but this might be to many options

plugin_example_v2


thanks for link, i was actually looking at that to get the v360 filter working. also using [this](https://www.mltframework.org/plugins/FilterAvfilter-v360/) and [this](https://ffmpeg.org/ffmpeg-filters.html#v360)
v360 does a fairly good job at removing the fisheye, all though setting are a bit fiddly. still got to do more testing.
‘og’ = Orthographic. it has some other modes too
v360_orthographic_01

ddennedy:
You can submit your frei0r changes here: https://github.com/dyne/frei0r
They do seem useful as long as they utilize the same interpolation kernels as the remainder of defish0r.

hypov8:
UI changes. compare collapsed and expanded, with all optional values

ui_v3_02

src/qml/filters/fisheye/meta.qml
```diff
@@ -0,0 +1,27 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+ 
+
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+    type: Metadata.Filter
+    name: qsTr("Fisheye Remover")
```
ddennedy, line 24: Upon reviewing the new screenshot it dawned on me that I should have suggested adding "Remover" to the filter name since this add or remove. Please change it to simply "Fisheye"

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,827 @@
...
+
+
+    GridLayout {
+        columns: 3
+        anchors.fill: parent
+        anchors.margins: 8
+
+        // Row 1 Preset
+        Label {
+            text: qsTr('Presets')
```
ddennedy, line 419: This might be a British vs. U.S. English thing, but it should be consistent. The U.S. English UI that is the source of all translations including the British uses singular "Preset" and not the plural. This should be consistent.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,827 @@
...
+            implicitWidth: 120
+            model: ListModel {
+                id: qualityModel
+                ListElement { text: qsTr('1. Nearest neighbor'); value: 0.0 }
+                ListElement { text: qsTr('2. Bilinear'); value: 0.166 }
+                ListElement { text: qsTr('3. Bicubic smooth'); value: 0.333 }
+                ListElement { text: qsTr('4. Bicubic sharp'); value: 0.500 }
+                ListElement { text: qsTr('5. Spline 4x4'); value: 0.666 }
+                ListElement { text: qsTr('6. Spline 6x6'); value: 0.833 }
+                ListElement { text: qsTr('7. Lanczos 16x16'); value: 1.0 }
```
ddennedy, line 502: Nowhere else in the Shotcut UI do we number the combobox list items. It might be your preference, but these should not be numbered to be consistent.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,827 @@
...
+                    ListElement { text: qsTr('3. Equiarea'); value: 0.666 }
+                    ListElement { text: qsTr('4. Stereographic'); value: 1.0 }
+                }
+                textRole: 'text'
+                onActivated: {
+                    filter.set(lensTypeParam, lensModel.get(currentIndex).value)
+                }
+            }
+            CheckBox {
+                text: qsTr('Dynamic stretch') //stretch view
```
ddennedy, line 539: Again, how is it "Dynamic?" Do you mean nonlinear?

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,827 @@
...
+                leftPadding: 6
+                Shotcut.HoverTip { text: qsTr(  'Using a simplified version of \'Elastic Scale\' image will be stretched/squished to fix camera scaling between 4:3 and 16:9\n'+
```
ddennedy, line 541: I prefer not to mention Elastic Scale here. You can simply replace "Using a simplified version of \'Elastic Scale\'" with "The" as those words do not add much and makes the person have to see what that is all about.

src/qml/filters/fisheye/ui.qml
```diff
@@ -0,0 +1,827 @@
...
+            CheckBox {
+                text: qsTr('Crop')
+                leftPadding: 6
+                Shotcut.HoverTip { text: qsTr('Remove distorted edges') }                
+                id: cropCheckBox
+                padding: 0
+                checked: cropDefault
+                visible: fisheyeRemoveButton.checked
+                onCheckedChanged: { if (blockUpdate) return; filter.set(cropParam, checked) }
+            }
```
ddennedy, line 641: We need a way to hide the Crop, Stretch, and Y ratio UI elements when the frei0r filter is not recent enough. There are Linux and BSD distributions that only include release frei0r plugins, and those are only released around every 2 years! A simple way to do this is by checking the presence of a new non-zero (due to how MLT works) default parameter value from the frei0r plugin. For example, the new property '10' has a frei0r/MLT default value of 1.0. In older versions of this plugin, MLT returns 0.0 when this property is requested. So, before defaults are set in `Component.onCompleted` check if this property value is > 0 and set a boolean QML property that you refer to in this object's `visible` property. You do not need to worry about using unavailable properties in the rest of the logic except where it will lead to a divide by zero. If you request the value of a non-existing parameter with `filter.getDouble()` it returns 0.

ddennedy (changes requested):

src/qml/filters/fisheye/ui.qml
hypov8: i was stuck with what to name it.
i will put it as 'Non-Linear scale' with the slider name as 'Scale'

ddennedy:
Please change all of the ComboBoxes to have an `implicitWidth: 150`. On my Linux and macOS desktop, some of the text items are truncated and looks ugly otherwise. Should it default to Remove instead of Add since that is the most common use case?

Someone who helps a lot with testing requested that I require all new filters to support slice threading. Slice threading is a technique where the image is split up into horizontal bands and each assigned to a thread that runs the core image processing function - `f0r_update()` in terms of frei0r. This is already done in MLT, but must be turned on for the individual MLT frei0r filter like so:

```
--- src/qml/filters/fisheye/ui.qml
+++ src/qml/filters/fisheye/ui.qml
@@ -392,7 +392,8 @@ Item {
     }
 
     Component.onCompleted: {
+        filter.set('threads', 0)
        if (filter.getDouble(scaleYParam) > 0.0)
             pluginShowNew = true        
           
         filter.blockSignals = true
```

0 = automatically determines thread count. When I do this here, the image does not look correct. If you run it with that, you will see what I mean. Is this something you want to try to fix? If so, then you should understand that MLT calls `f0r_update()` with multiple threads per frame where the height is the slice height, and the image pointers are pointing to the first row of a slice within the whole image. I will probably not require it after all the work you have put into it, but if you run that experiment you can observe the performance benefit despite the artifacts to weigh your decision. The performance benefit is very dramatic as you use interpolation methods higher than bilinear.

hypov8:
im having a look at what i can do.
is there any way to know the thread index that is running(current Y offset) when it enters `f0r_update()`?

ddennedy:
> is there any way to know the thread index that is running(current Y offset) when it enters `f0r_update()`?

Not currently, but one thing we can do that is a hack is to override the unused time parameter on certain plugins such as this that do nothing with it. Below is the simplest change in MLT to facilitate that:

```--- src/modules/frei0r/frei0r_helper.c
+++ src/modules/frei0r/frei0r_helper.c
@@ -60,6 +60,7 @@ static int f0r_update_slice( int id, int index, int count, void *context )
 	int slice_offset = index * slice_height * ctx->width;
 	uint32_t *input = ctx->inputs[0] + slice_offset;
 	uint32_t *output = ctx->output + slice_offset;
+	ctx->time = index;
 	ctx->f0r_update(ctx->frei0r, ctx->time, input, output);
 	return 0;
 }
```

ddennedy (approved):

ddennedy:
Regarding slice threading, the private instance member `map` is a big problem because it will be created at the slice dimension instead of the frame dimensions. The only way I can see to cleanly add slice threading is to port the filter to MLT. Alternatively, instead of slice threading, one can use OpenMP within the filter. This is what the bigsh0t set of "360:" filters in Shotcut use if you want to see how they perform or study its code to learn how to use OpenMP.

hypov8:
i got most of it working. needed to get original video dimensions. i have been using `profile.height` in qml

had trouble getting mlt to compile. has to get msys2 etc. make was missing a library in command. -liconv
 im on win7 using MSV to code/debug defish0r.

the distortion map i left alone and i change the pointer back to beginning of the input image. not sure how safe this is? but mlt just moved it forward for slice.
then just offset the distortion map when each slice in run. it must have a full input image as defish scales vertically, past slice dimension

currently got a bug that wont let me fetch parms `filter.getDouble`

it does seem like it has to many limitations to be using thread slicing tho

ddennedy:
> it does seem like it has to many limitations to be using thread slicing tho

I do not understand what you mean by "limitations" except perhaps build and development issues, but it does read as frei0r plugin changes too ugly to be accepted for merging. BTW, I do not know if MLT compiles with MSVC, never tried and only ever heard about failure from those who tried. In any case, I am going to merge this now, and we should instead pursue usage of OpenMP to accelerate this. I might since there are some plugins Shotcut uses that are incompatible with slice threading.

ddennedy:
> `f0r_update()` gets called multiple times for preview

That can be very normal. When playback is paused every time the video frame needs to be refreshed because a parameter of any filter changed the video frame is reconstructed. The system is not optimized to only refresh one filter as that would require saving the state of the frame's data prior to each filter, which would consume even more memory. It might be called more than necessary, however, and the reasons and their solutions are complicated.

> it seems to make more calls to f0r_construct() than f0r_destruct()

That happens due to reference counting. Do you have much experience with that? A frame in a queue holds a reference to a producer, which holds a reference to the filter in addition to Shotcut's own reference to the filter. This makes things complex, and you may not see a destruction immediately when you expect it. There is likely to be some reference leaks or unnecessarily saved MLT objects in Shotcut code.

PR #1288, opened 2022-06-12
Minor best-practice update

Velocities:
Just an adjustment to a set of multiple if, else if and else statements.

### DIFF
--- a/src/models/playlistmodel.cpp
+++ b/src/models/playlistmodel.cpp
@@ -308,16 +308,19 @@ QVariant PlaylistModel::data(const QModelIndex &index, int role) const
             Mlt::Producer parent(producer->get_parent());
             int width = THUMBNAIL_WIDTH;
             QImage image;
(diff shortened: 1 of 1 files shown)


bmatherly:
Does this compile and run?  A quick google search suggests that it is not possible to make a switch statement on a QString?

Also, the proposed change makes the code longer. So I am unsure what the improvement is or what best practice is being followed.

Docteh:
Interestingly enough they also filed another similar pull request within the same hour: https://github.com/yuzu-emu/yuzu/pull/8456

Here is the feedback:

> `switch` statements can only be used with enum or integral types, not string types. If this does actually work, it's because of a compiler extension that we shouldn't be relying on (e.g. this will not compile on MSVC)

Velocities:
I only just realized this lol, sorry for the mistake (I'm relatively novice to programming).

PR #1296, opened 2022-06-30
cmake: Fix CuteLogger install directory for mingw

Biswa96:
libCuteLogger.dll should be installed in bin directory.
Without this, libCuteLogger.dll is installed in lib directory.
Here DLL are runtime targets but SO are library target.


ddennedy:
Rejected because our build is made for our usage and scripts, which does not use a bin directory on windows.

Biswa96:
Is it possible to accept a patch which will add `WINDOWS_DEPLOY` option to change bin dir like mlt library?

ddennedy:
Yes

PR #1298, opened 2022-07-07
new translations added

mohab8437:


ddennedy:
Please use https://www.transifex.com/projects/p/shotcut/ as mentioned [here](https://shotcut.org/howtos/contribute/). Otherwise, your changes will be overwritten the next time I pull all translations from that service before release.

PR #1301, opened 2022-07-17
Fix moving track does not change the blend order

bmatherly:
As reported here:
https://forum.shotcut.org/t/moving-tracks-does-not-change-the-hierarchy-of-tracks/34689

It is not enough to assign each transition a new b_track.
The transitions mush also be reconnected in the right order

I would appreciate some review on this since I am not an expert on tractors and fields.

### DIFF
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -3044,32 +3044,34 @@ void MultitrackModel::moveTrack(int fromTrackIndex, int toTrackIndex)
     // Take a copy of the track
     QScopedPointer<Mlt::Producer> producer(m_tractor->multitrack()->track(fromMltIndex));
 
-    // Save the transition b tracks and calculate the new mix mode indices to be reapplied later.
-    QVector<int> bTracks;
-    QMap<int, QString> modes;
+    // Take a copy of all the transitions
+    // Save the transitions in order of their new track index
+    QMap<int, Mlt::Transition> videoTransitions;
+    QMap<int, Mlt::Transition> audioTransitions;
     QScopedPointer<Mlt::Service> service(m_tractor->producer());
     while (service && service->is_valid()) {
         if (service->type() == mlt_service_transition_type) {
             Mlt::Transition t((mlt_transition) service->get_service());
             int newBTrack = t.get_b_track();
-            if (service->get("mlt_service") != QString("mix") ) {
-                // Figure out how the blend modes will be moved to match the track move
-                if (newBTrack == fromMltIndex) {
-                    newBTrack = toMltIndex;
-                } else {
-                    if (newBTrack > fromMltIndex) {
-                        newBTrack--;
-                    }
-                    if (newBTrack >= toMltIndex) {
-                        newBTrack++;
-                    }
+            if (newBTrack == fromMltIndex) {
+                newBTrack = toMltIndex;
+            } else {
+                if (newBTrack > fromMltIndex) {
+                    newBTrack--;
                 }
-                modes[newBTrack] = t.get("1");
+                if (newBTrack >= toMltIndex) {
+                    newBTrack++;
+                }
             }
-            bTracks.push_back(newBTrack);
+            if (service->get("mlt_service") != QString("mix") ) {
+                videoTransitions[newBTrack] = *service;
+            } else {
+                audioTransitions[newBTrack] = *service;
+            }
         }
         service.reset(service->producer());
     }
+
 
     if (toTrackIndex > fromTrackIndex) {
         beginMoveRows(QModelIndex(), fromTrackIndex, fromTrackIndex, QModelIndex(), toTrackIndex + 1);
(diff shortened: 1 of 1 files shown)


src/models/multitrackmodel.cpp
```diff
@@ -3044,29 +3044,30 @@ void MultitrackModel::moveTrack(int fromTrackIndex, int toTrackIndex)
     // Take a copy of the track
     QScopedPointer<Mlt::Producer> producer(m_tractor->multitrack()->track(fromMltIndex));
 
-    // Save the transition b tracks and calculate the new mix mode indices to be reapplied later.
-    QVector<int> bTracks;
-    QMap<int, QString> modes;
+    // Take a copy of all the transitions
+    // Save the transitions in order of their new track index
+    QMap<int, Mlt::Transition *> videoTransitions;
+    QMap<int, Mlt::Transition *> audioTransitions;
```
ddennedy, line 3050: We should avoid using pointers so much. `Mlt::Transition` has copy constructors from another `Transition` as well as `Service`. This avoids needing to `delete` and the casting:
```
--- src/models/multitrackmodel.cpp
+++ src/models/multitrackmodel.cpp
@@ -3046,8 +3046,8 @@ void MultitrackModel::moveTrack(int fromTrackIndex, int toTrackIndex)
 
     // Take a copy of all the transitions
     // Save the transitions in order of their new track index
-    QMap<int, Mlt::Transition *> videoTransitions;
-    QMap<int, Mlt::Transition *> audioTransitions;
+    QMap<int, Mlt::Transition> videoTransitions;
+    QMap<int, Mlt::Transition> audioTransitions;
     QScopedPointer<Mlt::Service> service(m_tractor->producer());
     while (service && service->is_valid()) {
         if (service->type() == mlt_service_transition_type) {
@@ -3064,9 +3064,9 @@ void MultitrackModel::moveTrack(int fromTrackIndex, int toTrackIndex)
                 }
             }
             if (service->get("mlt_service") != QString("mix") ) {
-                videoTransitions[newBTrack] = new Mlt::Transition((mlt_transition) service->get_service());
+                videoTransitions[newBTrack] = *service;
             } else {
-                audioTransitions[newBTrack] = new Mlt::Transition((mlt_transition) service->get_service());
+                audioTransitions[newBTrack] = *service;
             }
         }
         service.reset(service->producer());
@@ -3108,27 +3108,25 @@ void MultitrackModel::moveTrack(int fromTrackIndex, int toTrackIndex)
     // Add the transitions back in the new positions
     for (auto bTrack : audioTransitions.keys()) {
         int aTrack = 0;
-        Mlt::Transition *aTransition = audioTransitions[bTrack];
-        if (aTransition) {
+        Mlt::Transition aTransition = audioTransitions[bTrack];
+        if (aTransition.is_valid()) {
             m_tractor->plant_transition(aTransition, aTrack, bTrack);
-            delete aTransition;
         }
-        Mlt::Transition *vTransition = videoTransitions[bTrack];
-        if (vTransition) {
+        auto vTransition = videoTransitions[bTrack];
+        if (vTransition.is_valid()) {
             if (bTrack == 1) {
-                vTransition->set("disable", 1);
+                vTransition.set("disable", 1);
             } else {
                 // video transitions mix with track 1
                 // (except track 1 which mixes with track 0)
                 aTrack = 1;
-                vTransition->set("disable", 0);
-                QString mode = vTransition->get("1");
+                vTransition.set("disable", 0);
+                QString mode = vTransition.get("1");
                 if (mode.isEmpty()) {
-                    vTransition->set("1", "normal");
+                    vTransition.set("1", "normal");
                 }
             }
             m_tractor->plant_transition(vTransition, aTrack, bTrack);
-            delete vTransition;
         }
     }
 ```

File: src/models/multitrackmodel.cpp [hunk 2]
```diff
@@ -3087,39 +3088,48 @@ void MultitrackModel::moveTrack(int fromTrackIndex, int toTrackIndex)
...
-                if (!modes[bTrack].isEmpty()) {
-                    t.set("1", modes[bTrack].toUtf8().constData());
-                } else {
-                    t.clear("1");
+    // Remove the track and add it in the new position
+    m_tractor->remove_track(fromMltIndex);
+    m_tractor->insert_track(*producer, toMltIndex);
+
+    // Add the transitions back in the new positions
+    for (auto bTrack : audioTransitions.keys()) {
```
ddennedy, line 3109: `clazy` integration into Qt Creator gives me 2 warnings about this line:
> - [allocating an unneeded temporary container](https://github.com/KDE/clazy/blob/1.11/docs/checks/README-container-anti-pattern.md) (I still prefer a C++ range loop over using an iterator, and this is a short list of ints.)
> - [range loop might detach container](https://github.com/KDE/clazy/blob/1.11/docs/checks/README-range-loop-detach.md) (I personally tend not to use `qAsConst` to resolve this.)

Here is one approach that still makes a copy of the keys:
```
-    for (auto bTrack : audioTransitions.keys()) {
+    const auto keys = audioTransitions.keys();
+    for (auto &bTrack : keys) {
```
Alternatively using an iterator:
```
-    for (auto bTrack : audioTransitions.keys()) {
+    for (auto i = audioTransitions.keyBegin(); i != audioTransitions.keyEnd(); i++) {
+        int bTrack = *i;
```

ddennedy (changes requested):
Functionally, it is correct and working properly but requesting some code tidying.

src/models/multitrackmodel.cpp
bmatherly: Done. I prefer the iterator.

src/models/multitrackmodel.cpp
bmatherly: Done

PR #1306, opened 2022-07-25
UI for new gps_graphic filter + small updates to gps_text

dany123:
For gps_text (v2):
-removed the timezone parameter (now it's calculated via javascript directly)
-combined more stuff in getControls (renamed from gps_setControls)
-fixed dark/light icons
-moved video_speed location to the top, I think it makes sense to group it there
-added support for tracks passing the 180 meridian (but still not 100% if it passes both 0 and 180)

For gps_graphic:
-New filter, quite complex, the ui got quite big (again lol, I'm a big fan of having all the options) but I'm open to suggestions.
-there are a lot of preset params so I did a weird thing with them (sorry Brian, I know I said I'd follow other filter patterns) - I do think it improves readability a lot and it's not unnecessarily complex

image
image

### DIFF
--- a/src/qml/filters/gpsgraphic/meta.qml
+++ b/src/qml/filters/gpsgraphic/meta.qml
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2021 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.0
+import org.shotcut.qml 1.0
+
+Metadata {
+    type: Metadata.Filter
+    name: qsTr("GPS Graphic")
+    mlt_service: 'gpsgraphic'
+    qml: 'ui.qml'
+    vui: 'vui.qml'
+    allowMultiple: true
+    keyframes {
+        allowAnimateIn: true
+        allowAnimateOut: true
+        simpleProperties: ['geometry']
+        parameters: [
+            Parameter {
+                name: qsTr('Position / Size')
+                property: 'geometry'
+                isRectangle: true
+            }
+        ]
+    }
+}
--- a/src/qml/filters/gpsgraphic/meta.qml
+++ b/src/qml/filters/gpsgraphic/meta.qml
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Meltytech, LLC
+ * Copyright (c) 2022 Meltytech, LLC
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/src/qml/filters/gpsgraphic/ui.qml
+++ b/src/qml/filters/gpsgraphic/ui.qml
@@ -1,0 +1,1448 @@
+/*
+ * Copyright (c) 2018-2022 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import QtQuick 2.12
+import QtQuick.Controls 2.12
+import QtQuick.Layouts 1.12
+
+import QtQuick.Dialogs 1.2
+import QtQuick.Window 2.1
+import Shotcut.Controls 1.0 as Shotcut
+import org.shotcut.qml 1.0 as Shotcut
+import QtQml.Models 2.12
+
+Item {
+    id: gpsGraphRoot
+    width: 350
+    height: 850
+
+    property string rectProperty: "rect"
+    property rect filterRect: filter.getRect(rectProperty)
+    property var allParameters: [
+        'time_offset', 'smoothing_value', 'speed_multiplier',
+        'graph_data_source', 'graph_type', 'trim_start_p', 'trim_end_p', 'crop_mode_h', 'crop_left_p', 'crop_right_p', 'crop_mode_v', 'crop_bot_p', 'crop_top_p',
+        'color_style', 'color.1', 'color.2', 'color.3', 'color.4', 'color.5', 'color.6', 'color.7', 'color.8', 'color.9', 'color.10',
+        'show_now_dot', 'now_dot_color', 'show_now_text', 'angle', 'thickness', rectProperty, 'show_grid', 'legend_unit', 'draw_individual_dots',
+        'bg_img_path', 'bg_scale_w'
+    ]
+    property var default_colors: ["#00aaff", "#00e000", "#ffff00", "#ff8c00", "#ff0000"]
+    property string color_white: "#ffffff"
+    property string default_now_dot: "#00ffffff"
+    property string default_rect: '10%/10%:30%x30%'
+    property int js_tz_offset: 0
+
+    property bool _disableUpdate: true
+    property url settingsOpenPath: 'file:///' + settings.openPath
+
+    Shotcut.File { id: gpsFile }
+    Shotcut.File { id: bgFile }
+    signal fileOpened(string path)
+    onFileOpened: settings.openPath = path
+
+    Component.onCompleted: {
+        filter.blockSignals = true
+
+        var resource = filter.get('gps.file')
+        gpsFile.url = resource
+        if (filter.isNew) {
+            var usedParams
+            var _ = undefined //skip params faster
+
+         /* usedParams.push( set_gps_file_params(offset, smooth, speed) )
+            usedParams.push( set_graph_data_params(src, type,  startp, endp,  modeh, leftp, rightp,  modev, botp, topp) )
+            usedParams.push( set_graph_style_params(style,  nowdot, nowtext,  angle, thick, rect,  grid, legendunit, drawdots) )
+            usedParams.push( set_graph_background_params(bgpath, scalew, opacity) )
+            usedParams.push( set_graph_colors(cdot, c1, c2, c3, c4, c5, c6, c7, c8, c9, c10) )       */
+
+            usedParams = set_graph_data_params(0,0,  0,100,  0, 0, 100,  0, 0, 100)
+            usedParams.push( set_graph_style_params(1, 1, 0, 0, 5,  default_rect, 0, _, 0) )
+            usedParams.push( set_graph_colors(color_white, default_colors[0], color_white) )
+            filter.savePreset(usedParams, "2D map: full map progress line")
+
+            usedParams = set_graph_data_params(0,1, _,_, 0, 90, 100,  0, 90, 100)
+            filter.savePreset(usedParams, "2D map: follow @90%")
+
+            usedParams = set_graph_data_params(0,1, _,_, 0, 95, 100,  0, 95, 100)
+            filter.savePreset(usedParams, "2D map: follow @95%")
+
+            usedParams = set_graph_data_params(0,1, _,_, 0, 98, 100,  0, 98, 100)
+            filter.savePreset(usedParams, "2D map: follow @98%")
+
+            //1d graphs: speed, altitude, hr
+            usedParams = set_graph_data_params(1,0, _,_, 0, 0, 100,  0, 0, 100)
+            usedParams.push( set_graph_style_params(1,  1, 1,  0, _, _,  1, 'm', 0) )
+            usedParams.push( set_graph_background_params( '!' ) )
+            filter.savePreset(usedParams, "1D graph: Altitude")
+
+            usedParams = set_graph_data_params(1,1, _,_, 0, 90, 100,  0, 0, 100)
+            usedParams.push( set_graph_style_params(1,  1, 1,  0, _, _,  0, 'm', 0) )
+            usedParams.push( set_graph_background_params( '!' ) )
+            filter.savePreset(usedParams, "1D graph: Altitude, follow @90%")
+
+            usedParams = set_graph_data_params(2,0, _,_, 0, 0, 100,  0, 0, 100)
+            usedParams.push( set_graph_style_params(8,  1, 1,  0, _, _,  1, 'bpm', 0) )
+            usedParams.push( set_graph_background_params( '!' ) )
+            filter.savePreset(usedParams, "1D graph: HR")
+
+            usedParams = set_graph_data_params(2,1, _,_, 0, 90, 100,  0, 0, 100)
+            usedParams.push( set_graph_style_params(8,  1, 1,  0, _, _,  0, 'bpm', 0) )
+            usedParams.push( set_graph_background_params( '!' ) )
+            filter.savePreset(usedParams, "1D graph: HR, follow @90%")
+
+            usedParams = set_graph_data_params(3,0, _,_, 0, 0, 100,  0, 0, 100)
+            usedParams.push( set_graph_style_params(1,  1, 1,  0, _, _,  1, 'km/h', 0) )
+            usedParams.push( set_graph_background_params( '!' ) )
+            filter.savePreset(usedParams, "1D graph: Speed")
+
+            usedParams = set_graph_data_params(3,1, _,_, 0, 90, 100,  0, 0, 100)
+            usedParams.push( set_graph_style_params(1,  1, 1,  0, _, _,  0, 'km/h', 0) )
+            usedParams.push( set_graph_background_params( '!' ) )
+            filter.savePreset(usedParams, "1D graph: Speed, follow @90%")
+
+
+            usedParams = set_graph_data_params(0,1, _,_, 0, 99, 100,  0, 99, 100)
+            usedParams.push( set_graph_style_params(3, 1, 0, 0, 5, _, _, _, 1) )
+            filter.savePreset(usedParams, "Map style: show GPS points as dots @99%")
+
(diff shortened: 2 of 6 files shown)


bmatherly:
I was not able to get any text to appear with my test gpx file. Do you have an example file that I can use (and maybe some instructions of how to use it)?

src/qml/filters/gpsgraphic/meta.qml
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2021 Meltytech, LLC
```
bmatherly, line 2: 2022 now :)

src/qml/filters/gpsgraphic/ui.qml
```diff
@@ -0,0 +1,1448 @@
...
+            Shotcut.Button {
+                icon.name: 'media-playback-pause'
+                Shotcut.HoverTip { text: qsTr('Sync start of GPS to current video time.\nTip: use this if you recorded the moment of the first GPS fix.') }
+                implicitWidth: 20
+                implicitHeight: 20
+                onClicked: { set_sec_offset_to_textfields(filter.get('auto_gps_offset_now')) }
+            }
+            Shotcut.Button {
+                icon.name: 'document-save'
+                Shotcut.HoverTip { text: qsTr('Save the current offset as default to use in other GPS filters.') }
```
bmatherly, line 563: This does not match how we have settings in other filters/fields. In other (similar) situations, we just save the setting every time the field is edited. Then, we automatically restore it when the field is populated for the first time. Do you think that would work for you here? Essentially, we could just remove this button and always save the value for the user when they edit the field.

bmatherly (changes requested):
I might be able to provide more feedback if I can get the filter to draw something on my video. Please provide an example file if you have one.

src/qml/filters/gpsgraphic/ui.qml
```diff
@@ -0,0 +1,1448 @@
...
+            Label {
+                text: qsTr(' h:')
+                Layout.alignment: Qt.AlignRight
+            }
+            TextField {
+                id: offset_hours
+                text: '0'
+                horizontalAlignment: TextInput.AlignRight
+                validator: IntValidator {bottom: 0; top: 59;}
+                implicitWidth: 20
```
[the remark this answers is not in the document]
dany123: I think it's a scaling thing, on my screen I can fit easily 2 digits (this was my intention):
image

I just use the mouse wheel to edit them. 
I don't know what other design choice to make for better usage, the clasic ddd:hh:mm:ss would be a step down in my opinion.
It's already pretty long, but should I increase the width by 50% for each one?

src/qml/filters/gpsgraphic/ui.qml
dany123: Functionally that could work, but I think users would just be confused and pressing reset would set it to 0 or to the saved state? 

This is how I use it (or at least why I think it's useful): have 2-3 filter instances: one with map, one with altitude, one with speed. You vaguely sync one of them and copy that value to the others. After a few minutes you notice the speed doesn't really sync with a stop you make so you tweak a few more second: now you just click the save button then the restore one for the other 2 instances (in reality I actually ended with 5 graphs and 1 text a few days ago so really used this).

dany123:
> I was not able to get any text to appear with my test gpx file. Do you have an example file that I can use
 (and maybe some instructions of how to use it)?

Here is a very short random GPS track. I don't think there's any instructions to give, if I load the file in the filter I can already see the map drawn (but it's 1px thick and white, because of the default style, the "future" is thin). Go with the "2D map: full map progress line" preset and then change color style to "One color" and increase thickness to 10 to make sure you see the map over the video. I have a black color as a background and can see it immediately.

[20220721201416.zip](https://github.com/mltframework/shotcut/files/9212773/20220721201416.zip)

src/qml/filters/gpsgraphic/ui.qml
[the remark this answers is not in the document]
dany123: I think removing the text allows me to increase width by 50% to 30px and still look decent:
image

src/qml/filters/gpsgraphic/meta.qml
dany123: Fixed.

src/qml/filters/gpsgraphic/ui.qml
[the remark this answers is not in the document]
bmatherly: I think you could use fontMetrics.advanceWidth("XX") + 5 to have a solution that responds to font size changes.

https://doc.qt.io/qt-6/qml-qtquick-fontmetrics.html#advanceWidth-method

src/qml/filters/gpsgraphic/ui.qml
bmatherly: Your use case makes sense. And I think we should support it. But I think we should try to find another way. The "save" icon opens a file dialog in other places in Shotcut. So this behavior might surprise some people.

Have you see the ClockSpinner control used in the Timer filter? I think we could adapt that to add days:
https://github.com/mltframework/shotcut/blob/master/src/qml/filters/timer/ClockSpinner.qml
Then, we could modify the control to support copy/paste. 

In order to support your use case, the user could select the ClockSpinner from one filter, hit CTL-C, then switch to another filter and hit CTL-V.

This proposal would have the benefit of making the line more compact. 

How would you feel about trying to implement this? 

I hope you are not discouraged by my comments or feel like they are nit-picks. I am interested in this filter and just want to help make it as good as it can be.

bmatherly:
> [20220721201416.zip](https://github.com/mltframework/shotcut/files/9212773/20220721201416.zip)
This one works for me. Thanks. I will keep it around for testing.

src/qml/filters/gpsgraphic/ui.qml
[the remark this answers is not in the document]
dany123: I''m not sure why but fontmetrics returns less than I expected, 6px per char. This image is how it looks for me with 2x what it returns. I'd rather go with a fixed 30, what do you think? 2.5x could also get 30 but I don't know what it returns on other PCs, might look ridiculous.

image

Maybe I'm using it wrong? I added a :

```
    FontMetrics {
        id: fontMetrics
        font.family: "Arial"
    }
```

Into the root UI element then used it directly as:
`implicitWidth: fontMetrics.advanceWidth("99")*2`

src/qml/filters/gpsgraphic/ui.qml
dany123: I'm not a fan of the ClockSpinner control, I find it very hard to change/increment anything specific inside (to add 1 min to it it requires just too many steps/clicks/typing). 
I find the mouse wheel up/down for each division extremely important and I really used it every time I synced the GPS files so far. Yes, the control would solve the need for the save/restore and make the line more compact, but I'd rather remove the save functionality and the days division than the individual minutes/seconds boxes. I think the latest update with removing of the "days", "m:" etc texts and replacing them with ":" makes the line compact enough to not worry about the length anymore.

src/qml/filters/gpsgraphic/ui.qml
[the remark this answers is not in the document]
bmatherly: Maybe you are not specifying the correct font. Consider trying:

```
FontMetrics {
    id: fontMetrics
    font: offset_days.font
}
```

I find that I usually need to add a few pixels for padding. Maybe try:

`implicitWidth: fontMetrics.advanceWidth("99") * application.devicePixelRatio + 6`

Play with the "6" to find a good padding.

src/qml/filters/gpsgraphic/ui.qml
bmatherly: > I'm not a fan of the ClockSpinner control,

Ok. It was worth a try.

I still do not support the inconsistent use of the save button. How about, instead of saving it as a setting, use standard copy/paste buttons to copy/paste the values from the clipboard? That would be more intuitive than "saving" the values, I think.

src/qml/filters/gpsgraphic/ui.qml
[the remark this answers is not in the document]
dany123: It still returns 6 per char for me. I vote we go with a fixed 30.

src/qml/filters/gpsgraphic/ui.qml
dany123: Adding copy + paste is just a bit too many buttons and looks cluttered for my taste, but I also understand the save button issue, it would be very unintuitive therefore unused.
Now that I think about it, the whole copy-paste thing is not that important. I think it's best I remove it and wait to see if users complain/have some other ideas about it. In all honesty it's barely cutting down a few seconds of manual change so not vital by any means.

bmatherly: > Now that I think about it, the whole copy-paste thing is not that important. 

I can support that. After we get this merged, we can continue to make improvements.

src/qml/filters/gpsgraphic/ui.qml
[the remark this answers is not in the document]
bmatherly: The 30 works OK on my screen for two digits. But is not enough for triple digit days. Unless you want to add months/years, you might want to support 3 or 4 digits for days.

Also, you should add validators for the fields. Hours should be between 0-23. Minutes between 0-59. Seconds 0-59

image

src/qml/filters/gpstext/ui.qml
```diff
@@ -31,12 +31,13 @@ Item {
...
     property url settingsOpenPath: 'file:///' + settings.openPath
     Shotcut.File { id: gpsFile }
+    property int js_tz_offset: 0
     
     signal fileOpened(string path)
     onFileOpened: settings.openPath = path
     
     Component.onCompleted: {
-        var resource = filter.get('gps.file')
+        var resource = filter.get('resource')
```
bmatherly, line 40: For backwards compatibility, you should check for "gps.file" if "resource" returns an empty string. Else, existing projects will break when they are opened with this new version.

src/qml/filters/gpstext/ui.qml
```diff
@@ -293,16 +270,16 @@ Item {
                     recompute_time_offset()
                 }
             }
-            Label {
-                text: qsTr('days:')
-                Layout.alignment: Qt.AlignRight
-            }
+//            Label {
```
bmatherly, line 273: I prefer to remove unused code.

src/qml/filters/gpsgraphic/meta.qml
bmatherly, line 2: 2022

src/qml/filters/gpsgraphic/ui.qml
```diff
@@ -0,0 +1,1423 @@
...
+                currentIndex: 0
+                textRole: 'text'
+                Shotcut.HoverTip { text: qsTr('+ : Adds time to video (use if GPS is ahead).\n - : Subtracts time from video (use if video is ahead).') }
+                onActivated: {
+                    if (_disableUpdate) return
+                    filter.set('majoroffset_sign', sign_val.get(currentIndex).value)
+                    recompute_time_offset()
+                }
+            }
+//            Label {
```
bmatherly, line 456: I prefer to remove unused code.

src/qml/filters/gpsgraphic/ui.qml
```diff
@@ -0,0 +1,1423 @@
...
+    Component.onCompleted: {
+        filter.blockSignals = true
+
+        var resource = filter.get('resource')
+        gpsFile.url = resource
+        if (filter.isNew) {
+            var usedParams
+            var _ = undefined //skip params faster
+
+         /* usedParams.push( set_gps_file_params(offset, smooth, speed) )
```
bmatherly, line 65: I prefer to remove unused code.

src/qml/filters/gpsgraphic/ui.qml
```diff
@@ -0,0 +1,1423 @@
...
+            Shotcut.UndoButton {
+                onClicked: {
+                    if (_disableUpdate) return
+                    reset_legend_unit();
+                    filter.set('legend_unit', legend_unit.text);
+                }
+            }
+        }
+
+//        Label {
```
bmatherly, line 1141: I prefer to remove unused code.

src/qml/filters/gpsgraphic/ui.qml
```diff
@@ -0,0 +1,1423 @@
...
+            }
+            Shotcut.UndoButton {
+                onClicked: {
+                    slider_scaleW.value = 1
+                    filter.set('bg_scale_w', 1)
+                }
+            }
+        }
+
+//        Label {
```
bmatherly, line 1322: I prefer to remove unused code.

src/qml/filters/gpsgraphic/ui.qml
```diff
@@ -0,0 +1,1423 @@
...
+        }
+    }
+
+    Connections {
+        target: filter
+        onChanged: {
+            var newValue = filter.getRect(rectProperty)
+            if (filterRect !== newValue) {
+                filterRect = newValue
+//                filter.set(rectProperty, newValue)
```
bmatherly, line 1339: I prefer to remove unused code.

bmatherly (changes requested):
I think we are very close. We are down to nitpicks now. Thanks for tolerating my comments. Regarding the unused code (commented out), maybe you have intentions to add more features later. Can you stash those in another  branch somewhere? The reason I prefer to remove unused code is because I end up being the one to maintain all these filters as bugs are found or the underlying framework needs to change. Unused code makes the maintenance more time consuming for me.

src/qml/filters/gpsgraphic/ui.qml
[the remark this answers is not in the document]
dany123: I do have validators for all the fields but for some reason they don't revert back invalid values. I can see the invalid value doesn't go through (mouse up/down reverts to the previously valid value). I wonder if there's some text(string) vs int interaction I'm not aware of.

`                validator: IntValidator {bottom: 0; top: 59;}`

For the days I'm not sure what to do, I expect this field to be 0 for most "real" use cases but I can see 2 sort of common exceptions: 
-date of the device gets reset to something "default" like 1st of january <manufactured year> which would add some hundreds of years, or even 1xxx
-video (or image) has no date so it's read as epoch 0 which adds... 19.000+ days

I would personally consider both of these cases exceptions and not fully support them (visually I mean) because 5 digits wastes a lot of space for a corner case. I say a decent middle ground is 40px for days to make it slightly different from the others and induce the idea of "this field is different". I really don't want to add months/years, functionally they would serve no real purpose.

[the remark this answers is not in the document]
bmatherly: > I wonder if there's some text(string) vs int interaction I'm not aware of.

I found this helpful idea: https://stackoverflow.com/a/59113560/4355458

This worked for me:

```
                onTextChanged:
                {
                    if(!acceptableInput)
                        offset_hours.undo()
                }
```
But your hours validator is actually wrong. The top should be 23, not 59.

> I would personally consider both of these cases exceptions and not fully support them 

I think you are underestimating our users creativity. Users will add the filters to tracks, color/transparent clips, pictures that they have downloaded off the internet, etc. Putting the filter on another track is particularly helpful because it allow them to apply other affects to the text or graph without applying the effect to the video. They are going to expect all these scenarios to work. Let's just do 4 digits for the days. That should get us pretty far and hopefully avoid bug reports.

I would also encourage you to test some other use cases that are not the specific use case that you have in mind. I notice that when I apply the filter to a track, the filter starts from the beginning with each clip that it encounters on the track. This will be unexpected by users. But I do not know if we have to fix that as part of this PR. But, minimally, the filter should work well on a transparent clip on another track.

Another idea I have to help with the time calculation (not required for this PR):
We could add a status section that shows the detected start time of the clip and the gps file. That would allow the user to see what the expected offset is between the items. Something to think about.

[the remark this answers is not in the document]
dany123: > I found this helpful idea: https://stackoverflow.com/a/59113560/4355458
> But your hours validator is actually wrong. The top should be 23, not 59.

Good catch & find.

>Let's just do 4 digits for the days

Ok, works for me.

> Putting the filter on another track is particularly helpful because it allow them to apply other affects to the text or graph without applying the effect to the video. 

This will require tweaks from their part as it's not how the filter can work: it needs to "sync" to something that will provide a time back to it (which is normally the "current" clip underneath) so this is not directly supported simply because of how the entire filter works. I did encounter a case myself when I wanted to do this and I had to visually re-sync the gps (as it was now starting at epoch 0). I don't see another way around this except maybe making a dummy copy of the real clip and making the clip copy transparent somehow.

>I notice that when I apply the filter to a track, the filter starts from the beginning with each clip that it encounters on the track. 

This is actually intentional (unless I misunderstood your case) and I think the correct way to it. In your case the filter starts from the begining because you're reusing the same image/clip which means it will re-sync to that previous "location". 
But think of this scenario: you do a travel vlog in Paris and you add a map showing the current location. Every time you show this one clip near the eiffel tower you definitely want the map to show the same real location near the eiffel tower.

The previous gps_text filter actually messed this up when applied to the timeline by adding the current timeline time to the actual clip time ( mlt_frame_original_position() vs mlt_frame_get_position() ) so the same clip shown at minute 0 and then again at minute 10 would show different values (and I actually thought I had a major bug because locally it was perfectly synced but it somehow drifted when showing another clip, so for a short <1min export this was almost invisible, but for a 30min video it was quite absurd - but only if applied to the timeline, directly on clip was fine).

I'm going to wait for user feedback on variations for this one as I assume people will have specific use cases that I can't even imagine right now and will decide what to support on a case by case/popularity basis.

> We could add a status section that shows the detected start time of the clip and the gps file

Hmm, I actually removed this even from the original gps_text (I left it as a hidden feature when hovering over the chosen .gpx file (see attached img)), I found out in practice that I never ever used it and even when I had a sync issue it was more confusing than just trying to find a "stop" in the video and sync it to a 0km/h speed in gps. It's actually more confusing than it looks because of timezones and the fact that most clocks drift if not connected to the internet.
image
vs
image

If you think it's useful enough I can add it back from diff. I kind of think the even longer UI might scare people and it looks better without.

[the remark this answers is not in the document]
bmatherly: > I'm going to wait for user feedback on variations for this one as I assume people will have specific use cases that I can't even imagine right now and will decide what to support on a case by case/popularity basis.

I think that is a good approach

> If you think it's useful enough I can add it back from diff. I kind of think the even longer UI might scare people and it looks better without.

I think it is probably useful  because it will help the user to understand why their offset is 150 days or whatever might look strange to them. But let's tackle that in another PR.

Thanks for your continued discussion and consideration in this PR. Let's fix the validation, clean up the nitpicks and get this one merged. I hope you will continue to make improvements in future PRs.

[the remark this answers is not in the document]
dany123: > I think it is probably useful because it will help the user to understand why their offset is 150 days or whatever might look strange to them. But let's tackle that in another PR.

All the data is already there so this is just a matter of restoring the old lines. I've added them back quickly.

>Thanks for your continued discussion and consideration in this PR. Let's fix the validation, clean up the nitpicks and get this one merged. I hope you will continue to make improvements in future PRs.

No problem, and sure, I'll hopefully have at least one update with the planned stuff and maybe other user suggestions.

bmatherly (approved):

PR #1313, opened 2022-08-05
Fix keyframes can have the same frame position and cross over one another after snapped to cursor

traprog:
Steps to reproduce:
1. Apply a filter to a clip.
2. Snap two neighboring keyframes to the cursor.
3. These keyframes can move passed each other afterward.

The fix I came up with is to check if the position is less than `model.minimumFrame` or greater than `model.maximumFrame` and adjust the position accordingly.

### DIFF
--- a/src/qml/views/keyframes/Keyframe.qml
+++ b/src/qml/views/keyframes/Keyframe.qml
@@ -129,6 +129,10 @@ Rectangle {
             // Snap to cursor
             if (keyX > cursorX - 10 && keyX < cursorX + 10) {
                 keyPosition = Math.round((cursorX) / timeScale) - (filter.in - producer.in)
+                if (keyPosition < model.minimumFrame)
+                    keyPosition = model.minimumFrame
+                else if (keyPosition > model.maximumFrame)
+                    keyPosition = model.maximumFrame
                 parent.x = cursorX - (parent.width / 2)
             }
             var trackValue = Math.min(Math.max(0, 1.0 - parent.y / (parameterRoot.height - parent.height)), 1.0)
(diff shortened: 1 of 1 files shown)


bmatherly:
Thanks for reporting this and proposing a solution. With your proposal, I am still able to crossover the keyframes in some cases. Can you check if this proposal works for you?

```
            if (keyX > cursorX - 10 && keyX < cursorX + 10 && cursorX > minDragX && cursorX < maxDragX) {
                keyPosition = Math.round((cursorX) / timeScale) - (filter.in - producer.in)
                parent.x = cursorX - (parent.width / 2)
            }
```

traprog:
With your proposal, I can't mess up keyframes if I move the first one toward the second, but I can if I move the second toward the first while the playhead is on the first one. That is to say, `cursorX > minDragX` isn't working.

One solution is to adjust `minDragX` and `maxDragX` by `parent.width/2`:
```
            if (keyX > cursorX - 10 && keyX < cursorX + 10 
                && cursorX > minDragX + parent.width/2 && cursorX < maxDragX + parent.width/2) {
                keyPosition = Math.round((cursorX) / timeScale) - (filter.in - producer.in)
                parent.x = cursorX - (parent.width / 2)
            }
```
Overall, I think your proposal is better because my proposal doesn't adjust `parent.x` in addition to your experience of not working in some cases. Even though it's literally 1 pixel visually, there is a misalignment between frontend and backend there.

I'll push this to my branch if you like the solution. I can't really mess up keyframes on my quick test. That should erase my proposal and replace it with this, right?

bmatherly:
> I'll push this to my branch if you like the solution. I can't really mess up keyframes on my quick test. That should erase my proposal and replace it with this, right?

Yes Please.

traprog:
I'm not doing this right. There are two commits there instead of one. Git command line is clearly not working for me to do actual pushes now. 
Can you manipulate this branch to make it have one commit? It says maintainers can edit and access this branch.

PR #1314, opened 2022-08-07
Fix timeline zoom slider doesn't update after opening of a timeline project file

traprog:
This bug appeared after 783b81120cd872186217ec1176831bb9f8bec0d5.

Create a timeline project, change the zoom level, and save the project. Close Shotcut and Relanuch Shotcut. Open the project through Open File. The zoom slider remains the same at 1.01.


bmatherly:
Thanks!

PR #1315, opened 2022-08-07
Fix trimclip regression

traprog:
I think this is a pretty serious regression after I made a commit about `trimClipIn` and `trimClipOut` here 5ba3005e450a20789a1e21873cf7bfabc22b6e71.

`removeTransitionByTrimInValid` and `removeTransitionByTrimOutValid` are triggered if `delta` is large enough to exceed the length of the neighboring transition, even if I am enlarging an existing transition through trimming.

If I'm moving my mouse slowly, `trimClipIn` or `trimClipOut` is called with small chunks of `delta`s, so this bug doesn't happen. Only if I move my mouse quickly, `delta` becomes large and the transition gets removed.

The fix makes sure that `removeTransitionByTrimInValid` and `removeTransitionByTrimOutValid` return `true` only if the mouse movement direction and m_isMakingTransition status match.

### DIFF
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -2123,7 +2123,8 @@ bool MultitrackModel::removeTransitionByTrimInValid(int trackIndex, int clipInde
         if (clipIndex > 1) {
             // Check if there is a transition and its new length is 0 or less.
             if (isTransition(playlist, clipIndex - 1)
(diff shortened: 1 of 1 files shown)


traprog:
In versions prior to 22.06.23, quick mouse movements will erase transitions. So, strictly speaking, this isn't a regression. It was subtle in prior versions but now is obvious in 22.06.23.

PR #1317, opened 2022-08-13
Fix selection changes on add or erase transition on a different track

traprog:
Steps to reproduce:
1. Place 4 clips on V1
2. Place 4 clips on V2.
3. Select all 4 clips on V2.
4. Make a transition by trimming between any two clip on V1.
5. Selection comes down from V2 to V1.

I don't think this is intended, especially now Shotcut has the ability to select multiple timeline clips.

The solution I came up with is to copy selection if tracks are different. Only if they are the same, `i.x()` adjustment with `n` is made.


bmatherly:
Thanks!

PR #1318, opened 2022-08-13
Fix duplicate after 2 clip moves, 2 undos, and 1 redo

traprog:
The current redo implementation stores information in producers. This makes difficult to deal with consecutive moves with the same selection.

I made a bunch of `QList` member variables, so that information stays within each undo command object.

This is highly reliant on the assumption that the order the producers are inserted into `newSelection` is the same as all the integers are inserted into the `QList`s.

It's working well on my tests and clears all the reports posted in the forum's v22.01 beta test thread that lead to reverting the initial attempt.

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -492,18 +492,18 @@ void MoveClipCommand::redo()
         }
     }
     // First, remove each clip.
+    int i = 0;
     for (auto &clip : m_selection) {
         if (!m_redo) {
             // On the initial pass, remove clips while recording info about them.
             QScopedPointer<Mlt::ClipInfo> info(m_model.findClipByUuid(MLT.uuid(clip), trackIndex, clipIndex));
             if (info && info->producer && info->producer->is_valid() && info->cut) {
-                info->producer->set(kNewTrackIndexProperty, qBound(0, trackIndex + m_trackDelta,
-                                                                   m_model.trackList().size() - 1));
-                info->producer->pass_property(*info->cut, kPlaylistStartProperty);
-                info->producer->set(kTrackIndexProperty, trackIndex);
-                info->producer->set(kClipIndexProperty, clipIndex);
-                info->producer->set(kShotcutInProperty, info->frame_in);
-                info->producer->set(kShotcutOutProperty, info->frame_out);
+                m_newTrackIndexList << qBound(0, trackIndex + m_trackDelta, m_model.trackList().size() - 1);
+                m_playlistStartList << info->cut->get_int(kPlaylistStartProperty);
+                m_trackIndexList << trackIndex;
+                m_clipIndexList << clipIndex;
+                m_inList << info->frame_in;
+                m_outList << info->frame_out;
                 newSelection.insert(info->cut->get_int(kPlaylistStartProperty), info->producer);
                 if (m_markerOldStart < 0 || m_markerOldStart > info->start) {
                     // Record the left most clip position being moved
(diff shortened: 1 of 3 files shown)


src/commands/timelinecommands.cpp
```diff
@@ -492,18 +492,18 @@ void MoveClipCommand::redo()
         }
     }
     // First, remove each clip.
+    int i = 0;
     for (auto &clip : m_selection) {
         if (!m_redo) {
             // On the initial pass, remove clips while recording info about them.
             QScopedPointer<Mlt::ClipInfo> info(m_model.findClipByUuid(MLT.uuid(clip), trackIndex, clipIndex));
             if (info && info->producer && info->producer->is_valid() && info->cut) {
-                info->producer->set(kNewTrackIndexProperty, qBound(0, trackIndex + m_trackDelta,
```
bmatherly, line 500: The following constants seem to be unused after this change:
* kNewTrackIndexProperty
* kClipIndexProperty
* kShotcutInProperty
* kShotcutOutProperty

So we could remove them from shotcut_mlt_properties.h

src/commands/timelinecommands.cpp
```diff
@@ -492,18 +492,18 @@ void MoveClipCommand::redo()
...
-                                                                   m_model.trackList().size() - 1));
-                info->producer->pass_property(*info->cut, kPlaylistStartProperty);
```
bmatherly, line 502: Question for @ddennedy:

I see the following properties being used in other parts of the code:
* kPlaylistStartProperty
* kTrackIndexProperty

I do not fully understand the purpose of those producer properties. Should they continue to be set here since they seem to be accessed in other parts of the code?

src/commands/timelinecommands.cpp
```diff
@@ -513,29 +513,32 @@ void MoveClipCommand::redo()
...
     }
     // Next, add the save clips to their new locations.
+    i = 0;
     for (auto &clip : m_selection) {
-        auto toTrack = clip.get_int(kNewTrackIndexProperty);
-        auto start = clip.get_int(kPlaylistStartProperty);
-        clip.set_in_and_out(clip.get_int(kShotcutInProperty), clip.get_int(kShotcutOutProperty));
+        auto toTrack = m_newTrackIndexList[i];
+        auto start = m_playlistStartList[i];
+        clip.set_in_and_out(m_inList[i], m_outList[i]);
```
bmatherly, line 533: Maybe kPlaylistStartProperty and kTrackIndexProperty need to be set on the clip here. I am unsure.

bmatherly (changes requested):
Testing went good for me. I have some questions about the properties.

src/commands/timelinecommands.cpp
traprog: Yeah, They can be removed. 
I thought that it'd be better to do it in another commit because strictly speaking these deletions contribute nothing to fixing the current bug.

src/commands/timelinecommands.cpp
ddennedy: Yes these must be updated when a clip moves.

src/commands/timelinecommands.cpp
traprog: We can set them by passing values from `QList`s to producer properties.
But they won't be updated during undos because `MoveClipCommand::undo()` uses `UndoHelper`, and I don't see any lines in `UndoHelper` that contain `kPlaylistStartProperty` and `kTrackIndexProperty`.

src/commands/timelinecommands.cpp
traprog: I restored lines that set values to these properties.
I think here is fine instead of later.

traprog: I just checked about `kTrackIndexProperty`.

When a user moves a clip from V2 (track index 0) to V1 (track index 1), kTrackIndexProperty remains 0 because it only registers a value retrieved from `m_model.findClipByUuid`, not a value calculated by `trackIndex + m_trackDelta`.
`kTrackIndexProperty` is relatively new entry in `shotcut_mlt_properties.h`, so the only affected place is `glaxnimateproducerwidget.cpp`

I think this is a separate issue. I haven't even looked into the issue of `kPlaylistStartProperty` and `UndoHelper`. Maybe this one is fine since it's been around like 5 years. We can merge this pull request after additional critiques if there is any.

src/commands/timelinecommands.cpp
bmatherly: It is fine as a separate commit, but I would like that commit to  be part of this PR so that we don't forget.

traprog: OK. I don't mind deletions being part of this pull request.
It's going to be better to squash all the commits, though. I don't want "Restore kTrackIndexProperty kPlaylistStartProperty" a part of the commit history.

src/commands/timelinecommands.cpp
bmatherly: I do not have any further comments, but I would like to see a thumbs up from Dan before we resolve this comment.

ddennedy (changes requested):
I do not have any specific change at this time. However, I think this change is risky and requires much more analysis and testing before merging. So, I want to block merging until I am comfortable with it.

ddennedy (approved):

PR #1321, opened 2022-08-14
Allow playlist auto change on open multiple files

traprog:
`if (m_model.rowCount() == 1)` was added 3 years ago. I assumed that some sort of crash would happen if it's changed. But I find nothing bad happening when this is executed.


ddennedy:
git blame shows this code was added in 260172329 with an interesting comment. Please read and test that ("Fix crash trimming playlist item.")

PR #1333, opened 2022-09-07
Append shortcuts to the end of action tooltips

bmatherly:
I forgot that we had discussed adding shortcuts to the tooltips. You can choose to merge this before the release, after the release, or not at all.


ddennedy:
It is a little weird that some toolbar icons show them but not others, and I worry that will draw another bug report. What do you think?

bmatherly:
> It is a little weird that some toolbar icons show them but not others, and I worry that will draw another bug report. What do you think?

I agree that will probably happen. But that will not bother me. I think this change is good for the users and this patch moves us in a positive direction. We will work on getting full coverage over time as a lower priority.

PR #1335, opened 2022-09-11
Fix clip move duplicated regression

traprog:
This converts `QMultiMap` to `QList` in `MoveClipCommand`.

`Poducer`s in `m_selection` is now ordered by how a user has selected clips on the timeline, not by playlist start positions.

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -451,7 +451,7 @@ void MoveClipCommand::redo()
 {
     LOG_DEBUG() << "track delta" << m_trackDelta;
     int trackIndex, clipIndex;
-    QMultiMap<int, Mlt::Producer> newSelection;
+    QList<Mlt::Producer> newSelection;
 
     if (!m_redo) {
         if (m_selection.size() > 1)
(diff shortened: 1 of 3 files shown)


traprog:
This regression fix causes another problem.
When multiple clips are moved beyond the upper and lower boundaries of the timeline, the result is totally dependent on which clip was selected first, second, third, and so forth.

ddennedy:
I am inclined to revert commit badd569 as it is less of a problem than the fixes.

ddennedy:
Another way to address the original bug is to reset the undo history on this operation.

traprog:
Best course of action is to close this pull request without merging, revert badd56971557ad3334a2ffd1e4701edb699e44d2, and implement your solution.

Even if I were to come up with another solution, the amount of changes in badd56971557ad3334a2ffd1e4701edb699e44d2 is so small that it doesn't make a difference if I start from here or before all the changes were made.

ddennedy:
Thank you for your understanding.

ddennedy:
Hey @traprog I just reverted my change to "reset the undo history on this operation" as it is more hostile than this bug. Imagine having 20 commands in History including some before and after a clip move and then choosing something fairly old in your history - possibly by accident. Zap there goes possibly a bunch of work! But when I introduced that I also added merging of multiple move commands for the same clip into one. As I was testing after reverting that, I realized this merging of move clip commands fixes the original bug!

traprog:
Yeah, I agree with that decision. Clearing all the history is way too much.

I'll check the nightly when it's available on Github Actions to see what's going on with the original duplicate bug.

traprog:
I'm still getting the original duplicate bug on Windows 11 with nightly 22.09.22 if I follow the steps described here:
https://forum.shotcut.org/t/clip-duplicate-after-2-moves-2-undos-and-1-redo/30904

Not a whole lot of people is reporting this bug in the forum, so this bug can be looked into again in future release cycles.
Maybe it's hard to naturally encounter this bug.

PR #1344, opened 2022-09-28
Update the Simplified Chinese translation

rerbin:
更新简体中文翻译


ddennedy:
Thank you for trying to help, but we only accept translations through transifex.com where people can collaborate without learning to use a  specialized tool.
https://shotcut.org/howtos/contribute/

PR #1349, opened 2022-10-05
Fix keyframes are lost when trimming transition

bmatherly:
As reported here:
https://forum.shotcut.org/t/pulling-on-the-left-side-of-a-transition-clip-can-mess-up-and-erase-all-keyframes/35837

Also improve updating the keyframes panel if it is affected by transition trimming

### DIFF
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -1849,8 +1849,16 @@ void MultitrackModel::trimTransitionIn(int trackIndex, int clipIndex, int delta)
         playlist.resize_clip(clipIndex, info.frame_in, info.frame_out - delta);
 
         // Adjust filters.
+        playlist.clip_info(clipIndex, &info);
+        MLT.adjustClipFilters(*info.producer, info.frame_in, info.frame_out, 0, -delta);
+
+        // Notify filters that the in changed.
         playlist.clip_info(clipIndex + 2, &info);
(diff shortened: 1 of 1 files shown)


ddennedy:
This is doing something wrong with adjusting the filter in and out points. You do not need to use keyframes to see this. Just add a filter (untrimmed) to each clip involved in the transition, trim the transition a little back and forth, and the filter on each may be trimmed unexpectedly as a result:
image

image

bmatherly:
> trim the transition a little back and forth, and the filter on each may be trimmed unexpectedly as a result:

That was existing behavior, so I thought that was expected. I can look at it more later. When the transition "in" is trimmed, I can not figure out why the 2nd clip would be changed at all. I think the expectation is that trimming the left (in) side of a transition should have no affect on the clip on the right. Is that correct?

ddennedy:
> That was existing behavior

not what I saw when comparing with release. It only does that when the filter in or out is not the same as the clip’s in and out. When they are the same trimming needs to adjust filters to match.

> When the transition "in" is trimmed, I can not figure out why the 2nd clip would be changed at all.

very simple. the second clip needs to be extended to cover the new transition or to contract.

bmatherly:
> very simple. the second clip needs to be extended to cover the new transition or to contract.

Ok. I understand, but I am not finding a solution. Can you take a look at this one?

PR #1351, opened 2022-10-09
Add user configurable jump actions

bmatherly:
As suggested here:
https://forum.shotcut.org/t/hotkey-to-skip-ahead-at-least-1-minute-at-a-time/34685

This is an attempt to fulfill multiple user requests for user configurable seeking.
Also, some users had implemented a custom patch to the QML based timeline menu that added user configurable seeking which no longer works with the latest release.

I only added these as actions in the player menu. I do not want to add more controls to the player toolbar. And I do not think they belong in the Timeline toolbar as the user patch put them.

Open to discussion on the Shortcut selection
image

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -310,6 +310,9 @@ void MainWindow::setupAndConnectPlayerWidget()
     ui->menuPlayer->addAction(Actions["playerBackwardFiveSecondsAction"]);
     ui->menuPlayer->addAction(Actions["playerForwardTenSecondsAction"]);
     ui->menuPlayer->addAction(Actions["playerBackwardTenSecondsAction"]);
+    ui->menuPlayer->addAction(Actions["playerForwardJumpAction"]);
+    ui->menuPlayer->addAction(Actions["playerBackwardJumpAction"]);
+    ui->menuPlayer->addAction(Actions["playerSetJumpAction"]);
     ui->menuPlayer->addAction(Actions["playerSetInAction"]);
     ui->menuPlayer->addAction(Actions["playerSetOutAction"]);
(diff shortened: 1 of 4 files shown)


ddennedy:
This is a minor feature addition that I am not interested in to change in the near term as there are currently two development paths: stability (with fixes) and major changes. But I am leaving it open to consider for future.

bmatherly:
Another recent request for this:
https://forum.shotcut.org/t/is-there-a-way-to-backward-15-frames-or-0-5-second-in-the-timeline/36459

ddennedy:
> Open to discussion on the Shortcut selection

One thought and we already broke this rule and not taking full advantage of it on main menu, but Alt+letter is often a shortcut to a field, button, or menu when it has an underscore. For example, Alt+F open the File menu. You can see we forgot to do Player and Settings. How about Ctrl+J?

I am considering this for the next version because it looks like Qt 6 migration has some fundamental problems that I am unsure about.

ddennedy:
merged in ffe5cac with other commits

bmatherly:
>   You can see we forgot to do Player and Settings. 

Player and Settings ALT shortcuts do not work on Windows. But they work on Linux for me. I do not find an immediate reason for this. Everything looks correct to me.

> but Alt+letter is often a shortcut to a field, button, or menu when it has an underscore.

We may have created a problem by adding shortcuts for ALT+M and ALT+R. But we do not stop the user from creating their own ALT+letter shortcuts. As an experiment, I created a ALT+F shortcut for an action. When I trigger the shortcut, the file menu shortcut takes priority over the action I have it assigned to. I suppose users will pick up on that. But we should probably avoid ALT+letter in general in the future.

> How about Ctrl+J?

Yeah. looks good. Thanks for changing that.

PR #1352, opened 2022-10-09
Add action to cycle color of selected marker

bmatherly:
As suggested here:
https://forum.shotcut.org/t/shotcut-marker-qol-thoughts/30510/4

Also intended to fill similar requests to easily/automatically make markers have different colors:
https://forum.shotcut.org/t/could-successive-markers-be-given-different-colours/34703
https://forum.shotcut.org/t/shotcut-marker-qol-thoughts/30510

I suggest "ALT + C" as the shortcut. I also consider that maybe it should not have a shortcut and let users choose to set a shortcut if it is a frequent use feature for them. Open to suggestions.


ddennedy:
This is a minor feature addition that I am not interested in to change in the near term as there are currently two development paths: stability (with fixes) and major changes. But I am leaving it open to consider for future.

ddennedy:
> I suggest "ALT + C" as the shortcut. 

See comment in #1351. Since it is more marker-oriented than color, perhaps Ctrl+Alt+M.

PR #1353, opened 2022-10-09
Position the bubble help to the lower right of the mouse

bmatherly:
Based on the suggestion here:
https://forum.shotcut.org/t/hide-time-indicator-in-box-below-cursor/35977/17

### DIFF
--- a/src/qml/views/keyframes/keyframes.qml
+++ b/src/qml/views/keyframes/keyframes.qml
@@ -393,7 +393,7 @@ Rectangle {
                                                 // Show amount trimmed as a time in a "bubble" help.
                                                 var s = application.timecode(Math.abs(clip.originalX));
                                                 s = '%1%2 = %3'.arg((clip.originalX < 0) ? '-' : (clip.originalX > 0) ? '+' : '').arg(s.substring(3)).arg(application.timecode(n));
-                                                bubbleHelp.show(clip.x, trackRoot.y + trackRoot.height, s);
+                                                bubbleHelp.show(s);
                                             } else {
                                                 clip.originalX -= delta;
                                             }
@@ -406,7 +406,7 @@ Rectangle {
                                                 // Show amount trimmed as a time in a "bubble" help.
                                                 var s = application.timecode(Math.abs(clip.originalX));
                                                 s = '%1%2 = %3'.arg((clip.originalX < 0) ? '+' : (clip.originalX > 0) ? '-' : '').arg(s.substring(3)).arg(application.timecode(n));
-                                                bubbleHelp.show(clip.x + clip.width, trackRoot.y + trackRoot.height, s);
+                                                bubbleHelp.show(s);
                                             } else {
                                                 clip.originalX -= delta;
                                             }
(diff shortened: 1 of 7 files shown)


ddennedy:
I do not agree that it addresses the problem or request. Maybe it is better positioned but I do not know yet and because it changes so many contexts that need evaluation for somewhat arbitrary change that is maybe not worth the time. What I do notice is that v22.09's changes adversely affects the vertical position. That is what needs to be addressed.

bmatherly:
> What I do notice is that v22.09's changes adversely affects the vertical position. That is what needs to be addressed.

Maybe that is what I was noticing. I did not go back to previous versions to compare if this was a regression. But in 22.09, the bubble would be arbitrarily lower than the mouse pointer - sometimes by a whole clip. In the case of the timeline ruler, the bubble would be covering the "good" part of the audio waveform of clips on the track when the mouse is over the ruler. 

Now that you mention it, this vertical offset could have been caused by removing the toolbar from QML.

Regardless, this change puts the bubble help consistently and predictably near the mouse pointer. So I think it is generally an improvement - in addition to resolving the presumed regression. I did test it in all the contexts where "show" is called. But I may not have tested every possible variation of having adjacent clips, etc.

src/qml/views/keyframes/keyframes.qml
```diff
@@ -513,9 +513,11 @@ Rectangle {
...
         property alias text: bubbleHelpLabel.text
 
-        function show(x, y, text) {
-            bubbleHelp.x = x + tracksArea.x - tracksFlickable.contentX - bubbleHelpLabel.width;
-            bubbleHelp.y = Math.max(0, y + tracksArea.y - tracksFlickable.contentY - bubbleHelpLabel.height);
+        function show(text) {
+            var point = application.mousePos;
+            point = parent.mapFromGlobal(point.x, point.y);
+            bubbleHelp.x = point.x + 12;
+            bubbleHelp.y = point.y + 12;
```
ddennedy, line 520: These 12 values are dependent on the mouse cursor size, which I doubt there is a way to get in QML, but some people will run with a larger cursor and then the bubble help appears underneath a little, which is not a big deal but worth taking into consideration instead of being tight like this.

ddennedy:
```
            bubbleHelp.x = point.x + 20;
            bubbleHelp.y = Math.max(point.y - 20, 0);
```
works better for me to prevent obscuring waveforms and dealing with a larger cursor size.

bmatherly: I was thinking of it as a tooltip and tried to emulate that visual look/position. And in fact, I was even thinking about re-implementing the bubble help using [QToolTip](https://doc.qt.io/qt-6/qtooltip.html). But that is a bigger change, and I like the "bubble" look that we have.

Tool tips typically appear in the lower right - presumably because we read top-to-bottom/left-to-right:
image
Also, tool tips often have logic to move the box around to avoid being clipped off the screen:
image
I did not bother to write extra logic to avoid clipping the bubble help box

In this specific case, I think your suggested offsets do work  better in the timeline. So I made that change. But I am open minded about it and we can tweak them. I am also willing to re-implement the  bubble help as a QToolTip - but then we will have less control over the position. However, that would increase consistency with the rest of the Shotcut UI.

ddennedy: It was not intended to be a tooltip. I don’t want to reimplement many things in the GUI in the near- or mid-term.

PR #1354, opened 2022-10-11
Add an option to not reset normalization measurement

bmatherly:
This is useful when clips on a track have similar loudness. Setting this parameter will avoid clips being loud at the begining after the discontinuity causes a reset

As discussed here:
https://forum.shotcut.org/t/how-does-normalizing-work-across-clips/34669/4

Staging until after release.
Depends on https://github.com/mltframework/mlt/pull/830

### DIFF
--- a/src/qml/filters/audio_normalize_1p/ui.qml
+++ b/src/qml/filters/audio_normalize_1p/ui.qml
@@ -27,6 +27,7 @@ Item {
         maxgainSlider.value = filter.getDouble('max_gain');
         mingainSlider.value = filter.getDouble('min_gain');
         maxrateSlider.value = filter.getDouble('max_rate');
+        discResetCheckbox.checked = parseInt(filter.get('discontinuity_reset'));
     }
 
     width: 480
@@ -39,6 +40,7 @@ Item {
             filter.set('max_gain', 15);
             filter.set('min_gain', -15);
             filter.set('max_rate', 3);
+            filter.set('discontinuity_reset', 1);
             filter.savePreset(preset.parameters);
         }
         setControls();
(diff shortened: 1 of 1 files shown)


src/qml/filters/audio_normalize_1p/ui.qml
```diff
@@ -27,6 +27,7 @@ Item {
         maxgainSlider.value = filter.getDouble('max_gain');
         mingainSlider.value = filter.getDouble('min_gain');
         maxrateSlider.value = filter.getDouble('max_rate');
+        discResetCheckbox.checked = filter.get('discontinuity_reset');
```
ddennedy, line 30: You cannot set a boolean from this property value directly. In JavaScript a "0" is true:
```
> !!"0"
true 
```
You can use `parseInt(filter.get())`.

src/qml/filters/audio_normalize_1p/ui.qml
```diff
@@ -226,6 +228,25 @@ Item {
...
         }
 
+        Label {
+        }
+
+        CheckBox {
+            id: discResetCheckbox
+
+            text: qsTr('Reset on discontinuity')
+            onCheckedChanged: filter.set('discontinuity_reset', checked)
```
ddennedy, line 238: I believe using `onCheckedChanged` means that when you set `.checked` in `setControls()`, it is going to set the property. Will that be problematic at all? Since this inherits from AbstractButton, you can also use `onClicked` but you will need to change the undo onClicked.

ddennedy (changes requested):

src/qml/filters/audio_normalize_1p/ui.qml
bmatherly: Done

src/qml/filters/audio_normalize_1p/ui.qml
bmatherly: > I believe using onCheckedChanged means that when you set .checked in setControls(), it is going to set the property.

Yes
> Will that be problematic at all?

No. Because the property setting is ignored if it is already set to the same value
https://github.com/mltframework/shotcut/blob/master/src/qmltypes/qmlfilter.cpp#L165

PR #1355, opened 2022-10-14
Allow the user to override sample rate when transcoding

bmatherly:
We tend to call the file converter "edit friendly". In Shotcut, 48kHz is the most friendly audio sample rate.

See past issues which do not have a good solution:
https://forum.shotcut.org/t/splitting-audio-adds-pops-clicks/24903
https://forum.shotcut.org/t/audio-crackling-in-exported-media-frankenbited-audio/35876/9

Looking for discussion on this...
One risk/disadvantage of this is that the user can still set the sample rate to something other than 48000 in the export settings. In that case, they might get unexpected results. But in that case, we can tell them to change the export settings to 48000.

### DIFF
--- a/src/dialogs/transcodedialog.ui
+++ b/src/dialogs/transcodedialog.ui
@@ -144,11 +144,55 @@
     <widget class="QWidget" name="advancedWidget" native="true">
      <layout class="QVBoxLayout" name="verticalLayout_2">
       <item>
-       <layout class="QGridLayout" name="gridLayout" columnstretch="0,1">
+       <layout class="QGridLayout" name="gridLayout" columnstretch="0,0">
         <item row="4" column="1">
          <widget class="FrameRateWidget" name="fpsWidget" native="true">
           <property name="toolTip">
            <string>Override the frame rate to a specific value.</string>
+          </property>
+         </widget>
+        </item>
+        <item row="6" column="1">
+         <widget class="QCheckBox" name="convert709CheckBox">
+          <property name="toolTip">
+           <string>This is useful when the source video is HDR (High Dynamic Range), which requires tone-mapping to the old, standard range.</string>
+          </property>
+          <property name="text">
+           <string>Convert to BT.709 colorspace</string>
+          </property>
+         </widget>
+        </item>
+        <item row="1" column="1">
+         <widget class="QCheckBox" name="subclipCheckBox">
+          <property name="toolTip">
+           <string>This option converts only the trimmed portion of the source
+clip plus a little instead of the entire clip. When this option is
+used not all of the matching source clips are replaced, instead
+only the currently selected one.</string>
+          </property>
+          <property name="text">
+           <string>Use sub-clip</string>
+          </property>
+         </widget>
+        </item>
+        <item row="9" column="1">
+         <widget class="QCheckBox" name="advancedCheckBox">
+          <property name="toolTip">
+           <string>Enable this to keep the Advanced section open for the next time this dialog appears.</string>
+          </property>
+          <property name="text">
+           <string>Keep Advanced open</string>
+          </property>
+         </widget>
+        </item>
+        <item row="5" column="1">
+         <widget class="QComboBox" name="frcComboBox">
+          <property name="toolTip">
+           <string>Frame rate conversion method
+
+Duplicate: Duplicate frames.
+Blend: Blend frames.
+Motion Compensation: Interpolate new frames using motion compensation. This method is very slow and may result in artifacts.</string>
           </property>
          </widget>
         </item>
@@ -169,44 +213,13 @@
           </property>
          </widget>
         </item>
-        <item row="6" column="1">
-         <widget class="QCheckBox" name="convert709CheckBox">
-          <property name="toolTip">
-           <string>This is useful when the source video is HDR (High Dynamic Range), which requires tone-mapping to the old, standard range.</string>
+        <item row="7" column="0">
+         <widget class="QLabel" name="sampleRateLabel">
+          <property name="enabled">
+           <bool>true</bool>
           </property>
           <property name="text">
-           <string>Convert to BT.709 colorspace</string>
-          </property>
-         </widget>
-        </item>
-        <item row="5" column="0">
-         <widget class="QLabel" name="frcLabel">
-          <property name="text">
-           <string>Frame rate conversion</string>
-          </property>
-         </widget>
-        </item>
-        <item row="5" column="1">
-         <widget class="QComboBox" name="frcComboBox">
-          <property name="toolTip">
-           <string>Frame rate conversion method
-
-Duplicate: Duplicate frames.
-Blend: Blend frames.
-Motion Compensation: Interpolate new frames using motion compensation. This method is very slow and may result in artifacts.</string>
-          </property>
-         </widget>
-        </item>
-        <item row="1" column="1">
-         <widget class="QCheckBox" name="subclipCheckBox">
-          <property name="toolTip">
-           <string>This option converts only the trimmed portion of the source
-clip plus a little instead of the entire clip. When this option is
-used not all of the matching source clips are replaced, instead
-only the currently selected one.</string>
-          </property>
-          <property name="text">
-           <string>Use sub-clip</string>
+           <string>Sample rate</string>
           </property>
          </widget>
         </item>
@@ -227,14 +240,36 @@ only the currently selected one.</string>
           </property>
          </widget>
         </item>
-        <item row="7" column="1">
-         <widget class="QCheckBox" name="advancedCheckBox">
-          <property name="toolTip">
-           <string>Enable this to keep the Advanced section open for the next time this dialog appears.</string>
-          </property>
+        <item row="5" column="0">
+         <widget class="QLabel" name="frcLabel">
           <property name="text">
-           <string>Keep Advanced open</string>
+           <string>Frame rate conversion</string>
           </property>
+         </widget>
+        </item>
+        <item row="7" column="1">
+         <widget class="QComboBox" name="sampleRateComboBox">
+          <property name="enabled">
+           <bool>true</bool>
+          </property>
+          <property name="currentText">
+           <string>Same as original</string>
+          </property>
+          <item>
+           <property name="text">
+            <string>Same as original</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>44100</string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>48000</string>
+           </property>
+          </item>
          </widget>
         </item>
        </layout>
--- a/src/dialogs/transcodedialog.cpp
+++ b/src/dialogs/transcodedialog.cpp
@@ -106,6 +106,17 @@ void TranscodeDialog::set709Convert(bool enable)
     ui->convert709CheckBox->setChecked(enable);
 }
 
+QString TranscodeDialog::sampleRate() const
+{
+    QString sampleRate;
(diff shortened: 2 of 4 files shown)


ddennedy:
I don’t like this because you can input and export  higher than 48000.

bmatherly:
Another idea/solution I have:
* Add another setting to the project settings for sampling rate (similar to video mode and audio channels)
* Make this proposed transcode sample rate a user option in the transcode dialog (and default it to the project setting)
* Modify the SDL audio consumer to render using the project sample rate and then convert to 48000 before sending to SDL

ddennedy:
This is not edit friendly. These ideas are all bands aids that increase the complexity. The user did not say your suggestion solved the problem. Did you reproduce it?

ddennedy:
I have been reviewing some threads and bugs on the problem of audio artifacts to get some more perspective, and there are various workarounds people are using. I suspect there are multiple issues, some of which we addressed in 2020(?). I think you are claiming that regardless of tool if you two contiguous chunks of audio (uncompressed in memory), resample, and simply join them you almost always introduce an artifact. Do you think that there are also resample-related events that cause buffer under or overflows in some of the MLT services such as resample and mix?

ddennedy:
I do see that the resample filters are discarding for changes you made to improve a/v sync.

One thing that I want to avoid is asking people to convert certain things in their project for this problem, or convert everything, or they need to figure what to convert. That is headache and can lead to a lot of big files in the case of video or choosing to convert everything to be safe and simple.

There is also the problem of general discontinuity when joining disparate chunks of audio. That could possibly be addressed if there was some way intelligently join them, which might require a little extra samples before an in point or after an out point: tricky.

I am trying to think out-of-the-box or pie-in-sky here. What if the avformat producer could read all of the input's audio into memory and resamples it before using it? That would negate the need to resample and avoid a lot of decisions or big files. I made some measurements, and 10 minutes of stereo 16-bit 48 KHz is about 110MB. But 10 minutes of 5.1 24-bit 96KHz is nearly 1 GB. I think that uses too much memory even for the simpler case because some people bring in hours of game or live capture to cut down and/or work with many files. 

What if Shotcut managed some sort of full-time audio proxy? By full-time, I mean it applies whether proxy is turned on or not and also for export. It would create WAV proxies for everything, and at export time if you change the sample rate it builds proxies for that rate. Likewise, if sample rate is added to the project settings, then it automatically builds new proxies when changed.

Now, for the intelligent joining, what if there is a hidden filter we can add to tracks that can observe discontinuity points and smooth over them? The framework could set a property on a frame at the start of each new playlist item, for example. The filter can hold the previous frame's audio in its internal buffer to interpolate with the current frame. We can also add the FFmpeg [adeclick](https://ffmpeg.org/ffmpeg-filters.html#adeclick) filter to Shotcut for people to add to the Output track as a safety net or to deal with problematic sources.

bmatherly:
> resample each independently, and simply concatenate them you almost always introduce an artifact.

Correct. That is absolutely true due to repeated samples that are passed into the resampler to avoid underflow. 

> Do you think that there are also resample-related events that cause buffer under or overflows in some of the MLT services such as resample and mix?

I am pretty sure that I fixed all the under/overflow problems in 2020 when I made changes to duplicate samples.

> One thing that I want to avoid is asking people to convert certain things in their project for this problem, or convert everything, or they need to figure what to convert.

Ok. That is fair. It is a convincing argument to make this conversion a user option that defaults to "no change".
I still convert many of my own clips outside of Shotcut because of this sample rate conversion issue. So I would use it. But probably not everyone.

> There is also the problem of general discontinuity when joining disparate chunks of audio. 

A general solution for "ducking" audio between clips would also help to mask this sample rate conversion problem. I have been thinking about a way to add "clip ducking" to the mix transition so that it could perform a very fast fade out toward the end of a clip and a very short fade in at the beginning of the next clip. Alternately, we could make a button to "duck all clips on selected track" which would add fade in/out filters  to all clips on a track.  I am willing to look into this more.

> I am trying to think out-of-the-box or pie-in-sky here...

Your suggestions are outside of the box. But also complex. A way to implement this within the framework is to implement the audio sample rate conversion as a link rather than a filter. A link can read ahead samples from the next frame when it needs to pre-load the sample rate converter. I can easily implement a sample rate converter link. Then, we would need a way for the link to be added to all producers by the loader. Then, make sure the normalizer links are not shown in the UI the way that normalization filters are not shown today.

> Now, for the intelligent joining, what if there is a hidden filter we can add to tracks that can observe discontinuity points and smooth over them?

This is congruent with the clip ducking idea I suggested above. But why would it be hidden? I think the user will want to configure this filter.

> We can also add the FFmpeg [adeclick](https://ffmpeg.org/ffmpeg-filters.html#adeclick) filter to Shotcut for people to add to the Output track as a safety net or to deal with problematic sources.

This is a very good idea. This filter could be added to a track. I will look into this.

ddennedy:
> "ducking" audio between clips

Why ducking instead of interpolation as adeclick does? Ducking is more heavy-handed but works when you don't have a better tool. I think we need a smart and isolated declick since discerning people - the ones reporting the clicks - would probably prefer not to declick all of the audio for fear of a general slight degradation.

> why would it be hidden?

Because it would be added to all tracks automatically; I forgot to mention that. It does not need to be configurable. It simply looks at several samples before and after an edit point and does some interpolation. Something like adeclick needs configurable options because it is global, not informed, and lower level.

> I have been thinking about a way to add "clip ducking" to the mix transition 

You might have been thinking of something manual or an action that automates something manual here. But instead of the hidden filter I mentioned above, I think this informed declick could be added to the mix transition and require less effort to integrate. It simply needs to be informed about the start of clip - probably by a frame property set in `mlt_playlist.c`.

ddennedy:
> implement the audio sample rate conversion as a link rather than a filter

Please try that before the heavier lift of audio proxies. It should be pretty easy to change the loader to look for a link if it did not find the filter listed in loader.ini. 
As for the audio proxy idea, the avformat producer already has separate libavformat contexts for audio and video. I was thinking that for a video, when initializing the audio context, it checks for and uses the an `audio_proxy` property to get the file path before reusing `resource`. And Shotcut would be responsible for generating them and changing producers to use them. But yeah, it would be overall heavier and more error prone. Hopefully, a link when combined with ducking or smart declick is the solution.

bmatherly:
> Why ducking instead of interpolation as adeclick does? Ducking is more heavy-handed but works when you don't have a better tool.

Good comments. I agree. Clip ducking could be a workaround, but would be inferior to interpolation.

Also, we should reserve the word "ducking" for the more classical meaning of automatically reducing background audio in favor of foreground audio
https://www.pinnaclesys.com/en/tips/audio/audio-ducking/
This is a useful feature that has been requested in the past. Maybe another feature to add to the mix transition.

> I think this informed declick could be added to the mix transition and require less effort to integrate. It simply needs to be informed about the start of clip - probably by a frame property set in mlt_playlist.c.

I agree with this idea and I think it has the highest potential to make improvements with the lowest effort/risk

>> implement the audio sample rate conversion as a link rather than a filter

> Please try that before the heavier lift of audio proxies. 

I also agree with this and will consider it as a second priority.

I still prefer to edit with normalized clips to reduce conversions during editing and to get the most consistent results. I would  like to eventually extend the convert feature to be able to perform additional normalization:
* resolution & aspect ratio
* scan mode (partially possible with deinterlace option)
* colorspace (partially possible with 709 option)
* frame rate (already possible)
* audio sample rate (this patch)
* audio channels

But any normalization changes should be user options - and this patch does not include a user option. Would you be open to the idea of adding the missing normalization options I listed above? I might extend this patch to include the user option for sample rate, and also add resolution/aspect ratio.

The next step after adding normalizing options would be to implement batch conversion.

ddennedy:
> I still prefer to edit with normalized clips

Convert was not really intended for that. It is intended to handle engine limitations regardless of a project and its settings. Today, you can use Export (and Each Playlist Item in place of [Bath file conversion design](https://forum.shotcut.org/t/batch-file-conversion-design/36017)) to prepare media optimized for the project. However, that does not combine the two in one step, which is a valid desire.

But it should also be possible to use the hardware acceleration when choosing the H.264 option because you can do decode, scaling, and encoding all in the GPU memory extremely fast. And one should also be able to choose HEVC once you go to hardware acceleration because that is just as fast but produces smaller files. In fact, if you are going to make all those things configurable, people are going to ask for everything that is in Export today including the same set of presets, which seems reasonable enough. l I mean why have two very similar windows that are not the same, or two different sets of presets for something so similar? Don't also forget that certain files in a batch may need specific options or handling especially when it comes to things within Convert > Advanced today. But I guess you can say those should not be in the same batch.

All this is significant overlap now with a number of other mature tools like Handbrake and MediaConvert. (Handbrake nightly now does [HDR to SDR](https://www.reddit.com/r/handbrake/comments/tsk6bf/handbrake_is_incredible_hdr_to_sdr_tone_mapping/).) But let's not be naive about where this leads, and the work required. Or at least what I am saying is that you can add these options to Convert but only in a manner that is consistent with Export. (However, I also know that some people will be confused or forget about the difference between Export and. Convert since they look so similar especially when it comes to their Playlist integration.)

bmatherly:
> In fact, if you are going to make all those things configurable, people are going to ask for everything that is in Export today including the same set of presets, which seems reasonable enough. 

Sure. Users ask for everything. Your comments are valid, but maybe a little bit exaggerated.

The editing world has common workflows for media capture -> Ingest -> Editing -> Rendering. Ingest commonly includes a step to transcode to a mezzanine format. Premier Pro even has an option to perform the mezzanine transcoding in the background while the user is editing. That is how I think of the "Convert" feature in Shotcut - turn your source files into a common format that is good for editing. So I do not think we need to offer all options that are in export - only options that support the editing workflow.

I agree with some of your comments that inexperienced users may not understand these workflows - they just want to drop a junky file on the timeline and start editing. So they may be confused by two similar but different screens. Using careful naming and providing context would be key. But we already have this configuration and I am really only suggesting to adding two or more three user options in the "Advanced" section. So I do not think it will greatly increase the confusion.

ddennedy:
> I am really only suggesting to adding two or more three user options in the "Advanced" section

OK, go ahead.

bmatherly:
Update on this topic in general:

This PR now allows the user to choose from three options in the Advanced section of the transcode dialog:
* Use original
* 44100
* 48000

I would like this change for my own workflow since I prefer to normalize all my clips before I edit.

In addition...
I have looked into the ffmpeg adeclick filter. Unfortunately, that filter adds a delay - and so it does not work in the current avfilter imlementation because it does not always return the requested number of frames. I will have to write a link for this avfilter so that it can work in Shotcut.

I am currently working on a link for the swresample filter and will be committing that when it is ready. This will  be the most definitive solution for clicks caused due to repeated samples caused by the current resample filters.

src/dialogs/transcodedialog.ui
```diff
@@ -144,45 +144,44 @@
     <widget class="QWidget" name="advancedWidget" native="true">
      <layout class="QVBoxLayout" name="verticalLayout_2">
       <item>
-       <layout class="QGridLayout" name="gridLayout" columnstretch="0,1">
```
bmatherly, line 147: This is the mess I get when I use QT Designer to edit the ui rather than hand-editing the XML. All I did was add one label and one QComboBox. But the tools seems to move a bunch of things around.

ddennedy: Sometimes you need to break a layout and recreate it. In that case, you need to check the custom properties set on the layout beforehand in order to restore them. This is what was lost:
image

And yes, after changes the XML might look scary in a diff viewer.

ddennedy: I cleaned this up and also right-aligned the field labels per our UI conventions.

PR #1361, opened 2022-10-22
Fix The slider in the UI is not updated when a keyframe is deleted

bmatherly:
As reported here:
https://forum.shotcut.org/t/the-slider-in-the-ui-is-not-updated-when-a-keyframe-is-deleted/36150

QmlFilter has two signals with the same name:
 * changed()
 * changed(QString)

The QML Connections item can not bind to a signal when it is overloaded by another signal with the same name but different parameters. As a result, the onChanged() signal never fires in ui.qml.

This change merges the two similar signals into one. If the parameter name is not known, or does not apply, pass an empty string.

### DIFF
--- a/src/controllers/filtercontroller.cpp
+++ b/src/controllers/filtercontroller.cpp
@@ -179,7 +179,6 @@ void FilterController::setCurrentFilter(int attachedIndex, bool isNew)
         if (!m_mltService) return;
         filter = new QmlFilter(*m_mltService, meta);
         filter->setIsNew(isNew);
-        connect(filter, SIGNAL(changed()), SLOT(onQmlFilterChanged()));
         connect(filter, SIGNAL(changed(QString)), SLOT(onQmlFilterChanged(const QString &)));
     }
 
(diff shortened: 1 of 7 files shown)


src/controllers/filtercontroller.cpp
```diff
@@ -190,15 +189,15 @@ void FilterController::setCurrentFilter(int attachedIndex, bool isNew)
 void FilterController::onFadeInChanged()
 {
     if (m_currentFilter) {
-        emit m_currentFilter->changed();
+        emit m_currentFilter->changed("");
```
ddennedy, line 192: Please remove this and other calls with empty string, and instead add a default parameter value to the `changed` signal:
`void changed(QString name = QString());`

src/qmltypes/qmlfilter.cpp
```diff
@@ -691,7 +689,7 @@ void QmlFilter::preset(const QString &name)
         m_service.load(qUtf8Printable(fileName));
     }
 
-    emit changed();
+    emit changed("");
```
ddennedy, line 692: Another change to remove after changing the signal signature.

src/qmltypes/qmlfilter.h
```diff
@@ -145,7 +145,6 @@ public slots:
 signals:
     void presetsChanged();
     void analyzeFinished(bool isSuccess);
-    void changed(); /// Use to let UI and VUI QML signal updates to each other.
     void changed(QString name);
```
ddennedy, line 148: Consolidate to `void changed(QString name = QString());`

ddennedy (changes requested):

src/controllers/filtercontroller.cpp
bmatherly: done

src/qmltypes/qmlfilter.cpp
bmatherly: done

src/qmltypes/qmlfilter.h
bmatherly: done

PR #1366, opened 2022-10-27
Improve how ruler labels are shown

traprog:


bmatherly:
I tested this. I think it is good. I do not love how the time can be printed beyond the ruler. But this is probably better than cutting it of when it can be shown.

image

ddennedy:
Why? I am not keen to make personal little changes that might introduce some side effects or problem.

traprog:
I think this one is an obvious visual improvement that can benefit most users, not just me personally without apparent side effects.

Technically though, `Repeater`'s `model` property holds the number of repeated objects, and 
```
parent.width / (intervalSeconds * profile.fps * timeScale)
```
gets rounded up half way, and rounded down otherwise.
`Math.ceil` ensures that at least one objects is shown even if the ruler's width (`parent.width`) is really small.
If the ruler's width is the same as the interval, two objects are shown, and so forth.

`label.width` is taking the hint from `width` in `(x + width)`, which looks like it's intended to be there to hide a object only if the right side of the entire object reaches the `tracksFlickable.contentX`.

traprog:
> I do not love how the time can be printed beyond the ruler.

I can make the label to show only if it is within the ruler by adding:
```
&& (x + width + label.width) < rulerTop.width
```
to
```
44: visible: ((x + width + label.width) > tracksFlickable.contentX) && (x < tracksFlickable.contentX + tracksFlickable.width)  // left edge
```
in `Ruler.qml`.

ddennedy:
As for "why?" please do not assume I have read the forum, and feel free to link to an explanation:
https://forum.shotcut.org/t/improve-timeline-keyframe-ruler-label-display/36251

ddennedy:
This is a very minor finesse thing that I am not interested in to change in the near term as there are currently two development paths: stability and major changes. But I am leaving it open to consider for future.

PR #1367, opened 2022-10-27
Disable snapping if not on safe area rectangle lines

traprog:


bmatherly:
I think there could be a debate here. When you drag the corner, are you dragging a point, or are you dragging two lines (vertical and horizontal). I think you are dragging two lines and the current behavior meets my expectations. Without this current behavior, it would be hard to snap on purpose in some situations.

We already have an option in the grid menu to disable snapping. But we do not have the keyboard modifier to press ALT to disable snapping while dragging (like we have in the timeline). Rather than this change, I wonder if you would be willing to implement a keyboard modifier to disable snapping like we have in the timeline.

traprog:
Yeah, I think we are moving two lines when moving corners.

The counter argument is that all of the following snap to safe area lines, so it's not just a matter of a line snapping a line.
1) Size, Position, and Rotate's VUI's top, left, bottom, right handles -> Moving a line.
2) Size, Position, and Rotate's VUI's four corners -> Moving two lines.
3) Size, Position, and Rotate's VUI's center handle -> Moving a point.
4) Corner pin's VUI -> Moving a point.

There can be a use case where a user wants to crop top 80% and bottom 80% and doesn't want to crop anything on the left and right with the Crop: Rectangle filter. The current implementation allows this.

But is that the purpose of safe areas though? Is the ability to do this kind of thing an unwritten trick? I don't know what the optimal design is, so to speak. Snapping to invisible lines seemed obviously unintuitive, so this has been a suggestion on the forum. 

I will look into keyboard modifier regardless of if this is merged or not merged. No promise.

ddennedy:
No, thank you

PR #1368, opened 2022-10-29
Add Alt keyboard modifier to VUI to disable snapping to grids

traprog:
Implementation is the same as that of the timeline. That is to say, Alt is hard coded.

### DIFF
--- a/src/qml/filters/corners/vui.qml
+++ b/src/qml/filters/corners/vui.qml
@@ -251,8 +251,10 @@ Shotcut.VuiBase {
                         cursorShape: Qt.SizeFDiagCursor
                         drag.target: parent
                         onPositionChanged: {
-                            corner1Handle.x = snapX(corner1Handle.x + handleOffset) - handleOffset;
-                            corner1Handle.y = snapY(corner1Handle.y + handleOffset) - handleOffset;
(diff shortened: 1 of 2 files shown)


bmatherly:
LGTM. Personally, I prefer this over the checkbox in the grid submenu because:
* It is consistent with the timeline
* The user can engage/disengage it easily without stopping a drag in progress

src/qml/filters/corners/vui.qml
```diff
@@ -251,8 +251,10 @@ Shotcut.VuiBase {
                         cursorShape: Qt.SizeFDiagCursor
                         drag.target: parent
                         onPositionChanged: {
-                            corner1Handle.x = snapX(corner1Handle.x + handleOffset) - handleOffset;
-                            corner1Handle.y = snapY(corner1Handle.y + handleOffset) - handleOffset;
+                            if(!(mouse.modifiers & Qt.AltModifier)) {
```
ddennedy, line 254: please respect our code formatting: space between `if` and `(`

ddennedy (approved):
I do want to include this since it can be said that the shortcut to suspend snapping is not working here and can be considered a bug. Please do fix the minor code format, however.

ddennedy:
Thank you

PR #1387, opened 2022-12-15
Move the marker color cycle action to the timeline dock

bmatherly:
This allows the action to work even when the markers dock is not visible. Also, the action is applied to the marker under the cursor instead of the selected marker.

As observed here:
https://forum.shotcut.org/t/please-test-the-beta-for-version-22-12/36806/5

### DIFF
--- a/src/docks/markersdock.cpp
+++ b/src/docks/markersdock.cpp
@@ -135,7 +135,6 @@ MarkersDock::MarkersDock(QWidget *parent) :
     QIcon filterIcon = QIcon::fromTheme("marker", QIcon(":/icons/oxygen/32x32/actions/marker.png"));
     setWindowIcon(filterIcon);
     toggleViewAction()->setIcon(windowIcon());
-    setupActions();
 
     QScrollArea *scrollArea = new QScrollArea();
     scrollArea->setFrameShape(QFrame::NoFrame);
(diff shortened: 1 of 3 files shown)


ddennedy:
Small problem: in the Markers panel, when the list box has focus, cursor up/down, and choose Cycle Marker Color from the menu; it changes the color of the last marker row I clicked to give the list box focus. This then gets into the whole issue of current vs. selected in GUI lists. It looks like I figured out some way to synchronize those in Playlists. Not a big deal; just want to let you know.

bmatherly:
> This then gets into the whole issue of current vs. selected in GUI lists. 

This has nothing to do with current/selected because the action operates on neither. The action only operates on the marker under the timeline cursor. In your example, the reason that a previously selected maker changed is because you had not moved the timeline cursor off of it.

As another example, follow these steps:
1) Select a maker in the marker panel
2) Press the right arrow to step the player by one frame
3) Press CTL-ALT-M (or choose the cycle action from the menu)
Nothing will happen because there is no marker under the cursor.
This behavior is consistent with Delete (CTL-SHIFT-M) 

We could change the text in the actions to "Delete marker under cursor" and "Toggle marker color under cursor". But that sounds confusing.

Another option is that we could remove those two actions from the Marker menu. They would still be in the Timeline menu - which might make more sense since they are associated with the timeline cursor.

ddennedy:
Oh yeah, you’re right, we’ve long since introduced another context or target of an action. Aren’t UIs fun? I think the overall behavior with this change closest matches what people will want and expect.

PR #1398, opened 2023-01-14
Qt6

ddennedy:
Making this a pull request to make it easier to view diff on the web and track issues outside GitHub Issues system:

- [x] WebP via Qt (`qwebp.dll`) is not working on Windows.
- [x] Settings > Display Method > DirectX not working and no longer supported. Really, this sub-menu should be removed with migrations to Direct3D on Windows and Metal on Mac since that is the overall direction of Qt 6. That is a big project because `GLWidget` uses OpenGL to draw the video under the QuickWidget. This needs to be refactored into a base class with sub-classes with different tech implementations along with porting the shader code. The [SPIRV-Cross](https://github.com/KhronosGroup/SPIRV-Cross) toolkit can facilitate that, but first the existing GLSL shaders need to be converted into Vulkan GLSL: https://vulkan-tutorial.com/. That also makes it compatible with Vulkan, which is a good option to have on Linux. But it is not clear if the Linux version should be Vulkan only - uncertain about the state of it on older Linux systems. See also Qt's [Shader Tools](https://doc.qt.io/qt-6/qtshadertools-overview.html) for this.
- [x] macOS has something weird going on with the Filters panel. It shows wrong colors in Fusion theme.
- [x] on macOS some Quick dialogs appear on a new, separate window titled "Offscreen."
- [x] Glaxnimate/Linux's python not working.
- [x] Time Remap UI is blank on macOS.
- [x] On Windows seeing the video colors flicker to incorrect values during playback. These anomalies are not observed in the vector scope.
- [x] On Windows as was macOS, the colors in the Filters panel is weird. This happens when the chosen Shotcut color theme does not match with the OS color them dark or light. For example, you might have a light OS color theme and the default Shotcut theme Fusion Dark resulting in something like this:
image
Qt Quick Controls 6 is not honoring the custom app palette, and I tried various things to get it to work. What I did for macOS was to remove Themes from Settings and make it automatic light or dark Fusion based on the OS color. However, now the Shotcut color will change on many people, and they will complain.
- [x] On Windows, there are no native font and color dialogs from QML. The docs say:
> A native platform font dialog is currently available on the following platforms:
> - macOS
> - Linux (when running with the GTK+ platform theme)

So, it will affect some Linux desktops notably KDE and LXQt. That is weak. Instead a Quick implementation appears. Here is how it looks:
image
image

- [x] QML MessageBox is not a native dialog on any desktop OS and needs to be changed to Window with `flags: Qt.Dialog` or add custom `QMessageDialog`-backed.
- [x] If Filters panel is floated, after restart it fails to show the QML.
- [ ] There is a version compatibility problem on Linux. Older versions of Linux (Ubuntu 18.04) do not have a recent enough `libstdc++`. This is due to needing to install a much new version of gcc (10) on the ubuntu18.04 builder in order to compile Qt6 and Glaxnimate's modern C++. 

Caveats
- <strike>Using Qt-provided Qt binaries raises the minimum OS requirement to Ubuntu 20.04/glibc 2.31. The Qt build system is different for v6 being based on CMake and more modular. Well, I guess we can try: see their [configuration docs](https://github.com/qt/qtbase/blob/dev/cmake/configure-cmake-mapping.md).</strike>
- The minimum Windows version is now Windows 10. I do not yet know if that is a hard minimum or a "support" thing.
- Qt/Windows no longer provides the Windows Extras modules; so we lose the ability to show export progress on the taskbar icon.


bmatherly:
@ddennedy, would be good if you could take a look at this one:
https://github.com/mltframework/shotcut/pull/1398/commits/bd173de495d47f363adf7f7157fb1dc963a56cda

ddennedy:
Here is how the Fusion Light theme looks when Windows' color mode is Dark. Even the custom palette is not used for widgets:

image

ddennedy:
The light and dark palette issue also affects Linux running GNOME3 on my Pop!_OS 22.04 desktop. When the OS and application light vs. dark settings mismatch, it is broken. The problem on Linux is worse, however, due to the number of different desktop technologies: It is not clear if Qt is able to read all of their palettes and choose the correct thing.

insaneninja117:
This would officially break support for Windows 7 as Qt6+ uses API calls that are unrecognised by Windows 7. Unofficial Qt6 .dlls with Win 7 friendly API calls added back in exist and as long as ShotCut.exe itself doesn't start using Win10+ API calls then with the modified Qt dlls the program should remain unofficially usable by Win 7 holdouts such as myself (but probably only in portable mode if the installer would be updated alongside this in the future as well).

ddennedy:
Yes, I plan to break Windows 7 compatibility. This is now using Direct3D 11, which is not available for Windows 7. You can continue to use the latest version that supports Windows 7.

insaneninja117:
With regards to Direct3D 11 DX11 and DX11.1 are supported on Windows 7, DX11.2 and newer versions are not, so if you now use the CreateDXGIFactory2 command, that is another nail in the Win 7 coffin.
The unofficial QT fixes use the official source code as a starting point so while I would think that the "fixed" dlls are pretty trustworthy the question moves to for how long would these enthusiats keep it up.

Nevertheless, I understand your intention to move ShotCut to a more modern framework, I just wanted to clarify the hard minimum vs. support issue; it is a hard minimum.

PR #1408, opened 2023-02-25
Qt6 producer probe

bmatherly:
Requires https://github.com/mltframework/mlt/pull/879

I tested this with GPU mode and I did not experience any crashes.


ddennedy:
"wants to merge 130 commits into master"

This is based on the wrong branch. Just go ahead and commit this on the qt6 branch without a pull request after the MLT PR is merged.

PR #1411, opened 2023-03-06
Add new speed filters

bmatherly:
Add three new links - all providing variations of Speed: Forward, Forward/Reverse, and Slow Motion. Forward/Reverse is the superset that can do everything - but it also requires files with no b frames. Maybe the Slow Motion is not required since forward can do the same thing. But it limits the speed factor to 0x-1x which may be more intuitive for some people.

image

image


ddennedy:
Is Slow Motion including reverse? There is too many to choose from now. I wanted a simple way to tell people that if they do not want to convert from B frames or they want to keyframe speed (e.g. when looking at Properties > Speed) to use X. Now, it is more complicated to make recommendation (needs more info or explanation) and a roll-of-the-dice for casual users in general as to which will require conversion. I know we can decorate the label with some hint, but that will largely be meaningless to most. How about one new named "Forward Speed" or "Speed (Forward Only)?" If we do end up with multiple then names should follow our convention and start with "Speed:" and lose the parenthesis (ok for clarification but colon is for differentiation).

bmatherly:
Good point about the ":" vs "()". I will change that when we nail down the names.

The slow motion version is 0x - 1x. So no reverse (that would be -1x). I can remove the slow motion  version. I do not feel strongly about that one. Just offering it for discussion.

If we offer a speed filter that only goes forward, then people will absolutely ask for it to also do reverse. So we will need at least two filters. How about:
* Speed: Forward Only
  * Works with b-frames
* Speed: Forward & Reverse
  * Does not work with b-frames

I am also open to comments about the allowed speed range.
For "Forward Only" I chose 0-10x
For "Forward & Reverse" I chose -10x-10x

We could make it wider, but with such a wide range, the keyframe manipulation is not helpful for small changes. I also considered making 5x the limit. That is why I tried out the "slow motion" version to give the user the whole vertical keyframe space to manipulate the 0-1x range.

ddennedy:
OK, go ahead with "Speed: Forward Only" and "Speed: Forward & Reverse". As for the range, there is that keyframe header option "Zoom keyframe values." 10x seems fine for now.

src/qml/filters/speed/meta_forward.qml
```diff
@@ -0,0 +1,33 @@
...
+    isAudio: false
+    name: qsTr("Speed (Forward)")
+    keywords: qsTr('temporal speed ramp fast slow motion', 'search keywords for the Speed filter') + ' speed'
+    mlt_service: "timeremap"
+    qml: "ui_forward.qml"
+    isFavorite: false
+    allowMultiple: false
+
+    keyframes {
+        allowTrim: false
```
ddennedy, line 16: You need to bump the version in `link_timeremap.yml` and require it here using `minimumVersion`.

src/qml/filters/speed/meta_forward_reverse.qml
```diff
@@ -0,0 +1,34 @@
...
+    name: qsTr("Speed (Forward/Reverse)")
+    keywords: qsTr('temporal speed ramp fast slow motion reverse', 'search keywords for the Speed filter') + ' speed'
+    mlt_service: "timeremap"
+    qml: "ui_forward_reverse.qml"
+    isFavorite: false
+    allowMultiple: false
+    seekReverse: true
+
+    keyframes {
+        allowTrim: false
```
ddennedy, line 17: Add `minimumVersion` here too.

src/qml/filters/speed/meta_forward.qml
ddennedy: I already bumped the version in `link_timeremap.yml`.  JB pleaded for a MLT release for kdenlive. So, I did that. New version number is 2.

bmatherly:
> OK, go ahead with "Speed: Forward Only" and "Speed: Forward & Reverse". As for the range, there is that keyframe header option "Zoom keyframe values." 10x seems fine for now.

Done

src/qml/filters/speed/meta_forward.qml
bmatherly: Done

src/qml/filters/speed/meta_forward_reverse.qml
bmatherly: Done

PR #1415, opened 2023-03-21
MacOS: Create symlink to the qml folder after building

zadockmaloba:
Automatically create the symlink post build if it doesn't exist, since we only copy the qml folder to the Resource path during install.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -295,6 +295,13 @@ if(APPLE)
   target_sources(shotcut PRIVATE ${APP_ICON})
   set_source_files_properties(${APP_ICON} PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")
 
+  #Create a symlink to the qml folder after building
+  add_custom_command(
+      TARGET shotcut POST_BUILD
+      COMMAND ${CMAKE_COMMAND} -E create_symlink
+      ${CMAKE_SOURCE_DIR}/src/qml ${CMAKE_CURRENT_BINARY_DIR}/Shotcut.app/Contents/Resources/qml
+  )
+
   install(TARGETS shotcut BUNDLE DESTINATION ${CMAKE_INSTALL_PREFIX})
   install(DIRECTORY qml DESTINATION ${CMAKE_INSTALL_PREFIX}/Shotcut.app/Contents/Resources/shotcut/)
 endif()


ddennedy:
This needs to be a separate build target (not the shotcut executable target) so it does not apply during a non-development build. And it needs to be cross platform or at least made available on macOS only.

ddennedy:
Or have you verified that this does not interfere with the install target?

zadockmaloba:
> I noticed that the Linux build uses
> ` file(CREATE_LINK ${CMAKE_SOURCE_DIR}/src/qml ${CMAKE_BINARY_DIR}/share/shotcut/qml SYMBOLIC)`

file(CREATE_LINK...) causes CMake config to fail with the following:

Shotcut.app/Contents/Resources/qml':
  No such file or directory
  
This happens with a fresh build. As of now I'm not sure why this is happening, but will try building for Linux and compare.

zadockmaloba:
> > I noticed that the Linux build uses
> > ` file(CREATE_LINK ${CMAKE_SOURCE_DIR}/src/qml ${CMAKE_BINARY_DIR}/share/shotcut/qml SYMBOLIC)`
> 
> file(CREATE_LINK...) causes CMake config to fail with the following:
> 
> Shotcut.app/Contents/Resources/qml': No such file or directory
> 
> This happens with a fresh build. As of now I'm not sure why this is happening, but will try building for Linux and compare.

`file(CREATE_LINK ${CMAKE_SOURCE_DIR}/src/qml ${CMAKE_CURRENT_BINARY_DIR}/Shotcut.app/Contents/Resources/qml SYMBOLIC)`

ddennedy:
Maybe you need the `file(MAKE_DIRECTORY...` that is on the previous line.

src/CMakeLists.txt
```diff
@@ -295,6 +295,10 @@ if(APPLE)
   target_sources(shotcut PRIVATE ${APP_ICON})
   set_source_files_properties(${APP_ICON} PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")
 
+  #Create a symlink to the qml folder after building
+  file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Shotcut.app/Contents/Resources/qml)
```
ddennedy, line 299: MAKE_DIRECTORY should not have "/qml" at the end as that is the symlink in the next line.

ddennedy (changes requested):

src/CMakeLists.txt
zadockmaloba: Yes, my bad. Will fix it.

Issue #1434, opened 2023-05-01
23.04 TODO List


bmatherly:
#### Blockers
- [x] Motion Tracker region size crash: https://forum.shotcut.org/t/release-candidate-version-23-04-20/38515/21
- [x] Dan: Windows screen blanks or sleeps: #1435 
- [x] Brian: Move track is broken: https://github.com/mltframework/shotcut/issues/1429
- [x] Dan: Color range is broken: https://forum.shotcut.org/t/23-04-color-range-regression/38609

#### Non-blocker
- [x] #1439
- [x] Timeline Right Click: https://forum.shotcut.org/t/release-candidate-version-23-04-20/38515/28
- [x] #1440
- [x] Glaxnimate preview does not update automatically; you need to move the mouse over the canvas for it to refresh.,

ddennedy:
The WebM alpha export presets were fixed, but Ut Video alpha appears to be a bug in FFmpeg.

I have not been able to fix the video mode crash on the count generator with GPU effects turned on. That only occurs with the combination of GPU effects, counter, the playhead on the counter, and changing video mode. That is a very narrow set of conditions.

bmatherly:
> Ut Video alpha appears to be a bug in FFmpeg.

So what then? Revert back to an older FFmpeg? Or maybe just add it as a known bug to the release notes.

> That is a very narrow set of conditions.

Maybe it is a reliable way to recreate an otherwise infrequent crash. But I agree - not blocking for this release. I captured the crash in GDB and inspected the backtrace. A solution did not immediately occur to me. I have bookmarked that forum post and I will probably cycle back to it at some point - unless you beat me to it.

ddennedy:
Not going to revert FFmpeg for that one little bug when there are other options available.

PR #1436, opened 2023-05-04
Use "force_full_range" property to override the color range

bmatherly:
Parital fix for:
https://forum.shotcut.org/t/23-04-color-range-regression/38609

### DIFF
--- a/src/widgets/avformatproducerwidget.cpp
+++ b/src/widgets/avformatproducerwidget.cpp
@@ -1276,6 +1276,7 @@ void AvformatProducerWidget::on_rangeComboBox_activated(int index)
 {
     if (m_producer) {
         m_producer->set("color_range", index ? 2 : 1);
+        m_producer->set("force_full_range", index);
         recreateProducer();
     }
 }
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -362,6 +362,7 @@ void Util::applyCustomProperties(Mlt::Producer &destination, Mlt::Producer &sour
     p.clear("force_aspect_ratio");
     p.clear("video_delay");
     p.clear("color_range");
+    p.clear("force_full_range");
     p.clear("speed");
     p.clear("warp_speed");
     p.clear("warp_pitch");
@@ -376,7 +377,7 @@ void Util::applyCustomProperties(Mlt::Producer &destination, Mlt::Producer &sour
     if (!p.get_int(kIsProxyProperty))
         p.clear(kOriginalResourceProperty);
     destination.pass_list(source, "mlt_service, audio_index, video_index, force_progressive, force_tff,"
-                          "force_aspect_ratio, video_delay, color_range, warp_speed, warp_pitch, rotate,"
+                          "force_aspect_ratio, force_full_range, video_delay, color_range, warp_speed, warp_pitch, rotate,"
                           kAspectRatioNumerator ","
                           kAspectRatioDenominator ","
                           kCommentProperty ","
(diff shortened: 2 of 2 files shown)


src/widgets/avformatproducerwidget.cpp
```diff
@@ -1276,6 +1276,7 @@ void AvformatProducerWidget::on_rangeComboBox_activated(int index)
 {
     if (m_producer) {
         m_producer->set("color_range", index ? 2 : 1);
```
bmatherly, line 1278: I do not understand this line that sets "color_range". color_range is not a documented property for avformat_producer and it does not appear to use it anywhere. The "force_full_range" property is well documented, so it seems like that should be used.

ddennedy:
I have been working on this (I updated #1434 about that), and this is not the problem. The problem is actually in theory old if you had a source that is yuv422. The problem is that now the player is configuring the consumer for yuv420p now, and most sources are yuv420p. Now, avcolor_space filter things it has nothing to do, so no range conversion is performed. `color_space` is a documented AVOption, and avformat producer applies all MLT properties that is also an AVOption to the context objects in `apply_properties()`.

src/widgets/avformatproducerwidget.cpp
ddennedy: `force_full_range` should be deprecated

bmatherly:
Between this PR and https://github.com/mltframework/mlt/pull/912, I was able to get the expected behavior. But it sounds like you might have a handle on a more definitive fix. Thanks for looking into it.

I propose that we move this issue to "blocking" for the release.

src/widgets/avformatproducerwidget.cpp
bmatherly: That is surprising since it was just added last year:
https://github.com/mltframework/mlt/commit/0e0e526124f4392275bdeb2510e8fb086d23da8f
But I guess "color_range" is the new way.

ddennedy: Oh, yeah, that is confusing. Usage of "color_range" in Shotcut was added in 2019 while "force_full_range" was added in 2022.

bmatherly:
> Now, avcolor_space filter thinks it has nothing to do, so no range conversion is performed

That is also my conclusion. Maybe this shameful workaround could get us by:
https://github.com/mltframework/mlt/pull/912#pullrequestreview-1412223649

ddennedy:
I decided I prefer the semantics of using "color_range" since that is what we also use on the consumer, and it is an AVOption: https://mltframework.org/plugins/ProducerAvformat/#color_range

bmatherly:
Thanks Dan,
https://github.com/mltframework/mlt/commit/97dcafde3ebb6a196cc102d6ae5228216a91d60b is working for me in my tests. And I agree with the approach.

Some comments:

The documentation for color range lists string values (unknown, tv, pc, unspecified, mpeg, jpeg). But in the code we are comparing to a number.

`full_range = mlt_properties_get_int(properties, "color_range") == 2;`

I think our implementation should match the documentation. If you tell me which is preferred, I can propose a patch to make them consistent.

Would you like me to propose a patch to mark force_full_range as deprecated?

ddennedy:
View the output for `ffmpeg -h full` and search for `color_range`. It accepts both integers and strings. Our generated documentation says that, but it does not know how to show the integer options. But you are correct that my implementation is not working with the string values! I do not like the idea of comparing a property with several strings for every frame. So, I pushed a change to slightly refactor this into a property change listener instead of being inline within `producer_get_image()`. Then, "color_range" is handled as an AVOption and producer private state updated from the AVCodecContext. This way, it is only evaluated on property change, and the values - string or numeric - handled by FFmpeg. This property listener will probably become more generally useful such as as to always update a corresponding AVOption, but I want to keep the amount change right now small. I would have also liked to get rid of the redundant `producer_avformat.full_range` and just use `AVCodecContext.color_range`, but the same reason.

PR #1448, opened 2023-06-05
Add option to auto-scroll timeline by pages

ravemo:
Saw this feature in the roadmap and decided to implement it.
This option is located in Settings > Timeline. I have decided to add a 100-pixel window of overlap between each page because I think it would make it easier to edit things that happens to cross that transition. Let me know if you think it would be better to remove that overlap.

### DIFF
--- a/src/docks/timelinedock.cpp
+++ b/src/docks/timelinedock.cpp
@@ -928,6 +928,17 @@ void TimelineDock::setupActions()
     });
     Actions.add("timelineCenterPlayheadAction", action);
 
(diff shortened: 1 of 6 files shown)


ddennedy:
Hey, this is great. Thank you! Now, Center the Playhead and Auto-scroll by Pages behave mutually exclusive, but choosing one does not deselect the other. However, when I added this to the timeline I was thinking it would simply be the new behavior with no option. What do you think; should it be a setting? I will let you decide and make the appropriate change. Or, if I do not hear back from you within a week, I will remove the setting when I merge.

PR #1452, opened 2023-06-14
Add Track Auto Fade and Track Seam filters

bmatherly:
The filters are only visible for selection if a track is selected.

Seam:
image

Auto Fade:
image

### DIFF
--- a/src/controllers/filtercontroller.cpp
+++ b/src/controllers/filtercontroller.cpp
@@ -188,6 +188,7 @@ void FilterController::setProducer(Mlt::Producer *producer)
     if (producer && producer->is_valid()) {
         m_metadataModel.setIsClipProducer(!MLT.isTrackProducer(*producer));
         m_metadataModel.setIsChainProducer(producer->type() == mlt_service_chain_type);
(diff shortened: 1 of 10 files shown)


ddennedy:
FYI I am preparing a fixes-only release 23.06 at moment (without beta) that will not include this. From the screenshots, I can see that field labels are not supposed to be all capitalized - only the first word.
Regarding the indicators, I am concerned that people will think these should be checkable and report about it. Another way to do this is a Status field and toggle text visibility in its value column ala:
Status: Seam applied
Status: Fade in
Status: Fade out

P.S. When a field has a no control and only text/label then it does need a colon.
https://www.shotcut.com/notes/ui-conventions/

bmatherly:
Thanks for the comments.

I agree this feature should go through a beta period.

I prefer the simulated LED indicator for momentary status - similar to the Normalize: 1 pass filter reset indicator. I added the line separator to separate control from status like we do in Normalize and Compressor.

image

image

I am still open minded about this. But it moves it closer to being consistent with existing conventions.

PR #1456, opened 2023-07-10
Resource view

bmatherly:
This PR adds a resource view that can be triggered from the View menu. Also, when multiple files are opened (through File/Open or drag/drop), the resource view will pop up if any of the files need conversion.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -24,6 +24,7 @@ add_executable(shotcut WIN32 MACOSX_BUNDLE
   dialogs/listselectiondialog.ui
   dialogs/longuitask.cpp dialogs/longuitask.h
   dialogs/multifileexportdialog.cpp dialogs/multifileexportdialog.h
+  dialogs/resourcedialog.cpp dialogs/resourcedialog.h
   dialogs/saveimagedialog.cpp dialogs/saveimagedialog.h
   dialogs/slideshowgeneratordialog.cpp dialogs/slideshowgeneratordialog.h
   dialogs/systemsyncdialog.cpp dialogs/systemsyncdialog.h
@@ -73,6 +74,7 @@ add_executable(shotcut WIN32 MACOSX_BUNDLE
   models/motiontrackermodel.h models/motiontrackermodel.cpp
   models/multitrackmodel.cpp models/multitrackmodel.h
   models/playlistmodel.cpp models/playlistmodel.h
+  models/resourcemodel.cpp models/resourcemodel.h
   openotherdialog.cpp openotherdialog.h
   openotherdialog.ui
   player.cpp player.h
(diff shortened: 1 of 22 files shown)


ddennedy (changes requested):
One thing I notice is that when you drag-n-drop multiple to the Playlist, it bypasses this (unlike Timeline). However, it seems most peoples' workflow starts with drag-n-drop multiple to the Playlist (per the many tutorial videos I have seen). Same thing with the playlist import button (Add files to playlist). However, this is consistent with the current behavior, and one is prompted to convert when double-clicking a playlist item. But there is no prompt to convert when dragging playlist item(s) to Timeline, which is a regression. Now, one can drag HDR or VFR videos to Playlist and then from Playlist to Timeline (a common workflow) and they are never prompted to convert.

ddennedy:
Another couple little cosmetic things: the rows should have alternating background colors, and I think the columns should be resizable because the Name column could be very wide hiding other columns or you want to add more space between Video and Audio.

bmatherly:
> Now, one can drag HDR or VFR videos to Playlist and then from Playlist to Timeline (a common workflow) and they are never prompted to convert.

These are good insights. My first thought is that as soon as a file is pulled into Shotcut (either playlist, source or timeline), the prompt should appear. But after hearing your description, I am not so sure. Maybe it is OK if people drag HDR/VFR files into the playlist and do not receive a prompt?

ddennedy:
I am fine with the current release behavior (prompted in Source or Timeline) or when adding to the Playlist. But this change has neither when using the Playlist to import.

bmatherly:
Thanks for your review and comments.

> Now, one can drag HDR or VFR videos to Playlist and then from Playlist to Timeline (a common workflow) and they are never prompted to convert.

This is fixed now. Any file opened or imported by any means at all can prompt the resource dialog if it is VFR/HDR. Let me know if you find any other ways to sneak around it.

> the rows should have alternating background colors

Done

> the columns should be resizable

Done

I wonder what other requests people will have for this new dialog. People might ask to use the dialog to see what files are in the playlist, but not in the timeline. This is already a standing request for people that use the playlist as a "bin" and want to clean out clips that they have used. Or they might ask to click on a line in the dialog and have that select the clip in the timeline.

I thought about adding a "location" column in the table that would say "playlist" or "timeline" as appropriate. As it works now, if a file is in both the playlist and the timeline, it will only appear in the list once. But we could duplicate entries for playlist/timeline to show where they are.

I am open minded about adding more features to this dialog if it helps people to understand the resources that are in use.

ddennedy:
I like the current design (playlist + timeline | uniq), but I just realized it is not sorted and neither are the columns sortable. I do not see much value in column sorting, but I do see one for name. It does not need to be clickable heading with a descending option; simply sort the list. I will merge this now, and one of will get around to this if you agree.

Regarding other uses, I think we can eventually add a filter combo to show some different things. However, I think we should duplicate entries or a column with a vague location (Nth clip of track X). I am thinking it could select every clip - timeline or playlist - based on that file.

bmatherly:
> I like the current design (playlist + timeline | uniq), but I just realized it is not sorted and neither are the columns sortable. I do not see much value in column sorting, but I do see one for name. It does not need to be clickable heading with a descending option; simply sort the list. I will merge this now, and one of will get around to this if you agree.

I agree with this and it would not bother me if you make changes yourself based on your ideas/preference.

> Regarding other uses, I think we can eventually add a filter combo to show some different things. However, I think we should duplicate entries or a column with a vague location (Nth clip of track X). I am thinking it could select every clip - timeline or playlist - based on that file.

I also agree with this. There is one complication with duplicating entries. This dialog has a dual role: showing an overview of all resources, and allowing the user to convert files in batch. In the conversion scenario, a conversion will replace all instances of a resource with the converted file. If we duplicate entries, will the user understand that if they choose to convert one entry, all entries for that resource will be converted. I'm sure we can find a way to make this clear, but it would be something to watch out for.

ddennedy:
Sorry, I mean to write we should NOT duplicate entries!

PR #1461, opened 2023-07-25
GPS text ui: added cadence, temp, grade and units

dany123:
I've added the degrees symbol in some texts, it's part of extended ascii so I think it's fine in the file.

Btw, the long model list looked a bit more organised one entry per line (with a comment stating it's corresponding case value), not sure if you're willing to update the rules of the formatting tool for this but I think it would be a good change.


ddennedy (changes requested):
You should only include these new options based on the MLT filter version.

dany123:
> filter.isAtLeastVersion(3)

For some reason this isn't working, the call is always returning false, do I need to set the version somewhere myself?

I put these in the main setControls call to test out.
```
        console.log('v0+?' + filter.isAtLeastVersion(0))
        console.log('v1+?' + filter.isAtLeastVersion(1))
        console.log('v2+?' + filter.isAtLeastVersion(2))
        console.log('v3+?' + filter.isAtLeastVersion(3))
        console.log('v4+?' + filter.isAtLeastVersion(4))
```

and this is the output:
```
[Debug  ] <setControls> v0+?false
[Debug  ] <setControls> v1+?false
[Debug  ] <setControls> v2+?false
[Debug  ] <setControls> v3+?false
[Debug  ] <setControls> v4+?false
```

ddennedy:
I added your log line to the top of `setControls()` in `gpstext/ui.qml` and it works for me:
> [Debug  ] <setControls> v2+?true

There are other filters that use function.
The version is read from the MLT installed `$prefix/share/mlt-7/qt/filter_gpstext.yml`. Maybe Shotcut+MLT is unable to locate your `filter_gpstext.yml`. It is not super-critical for the core functionality of MLT and Shotcut, but it is consulted for special situations like this change request. Distro packages do not always have the latest, expected version of MLT for the Shotcut version.

dany123:
> The version is read from the MLT installed `$prefix/share/mlt-7/qt/filter_gpstext.yml`

This was my bad, thanks. 
I've still not managed to properly compile Shotcut (only mlt), so I'm just copy pasting stuff around sadly. 
Anyway, the check works now as it should.

dany123:
Quick question: I'll update the GPS graphic filter with grade support and a few minor fixes, do you have a preference if I should add them all in this PR or make a separate one?

ddennedy:
This is not going into this weekend’s release, and then I will be on vacation for 2 weeks. So I don’t have a preference, and you can choose to just use this branch.

PR #1462, opened 2023-07-27
Add audio declick (adeclick) filter.

bmatherly:
Depends on https://github.com/mltframework/mlt/pull/935

image

One disadvantage of adding this as a link is that it ends up under the "Time" filter category rather than the "Audio" category like someone might expect. This is because all links go in the time category. It can't really go in the audio category  because all links have to appear before the filters. I am open to ideas about how we can make that more intuitive.

image

### DIFF
--- a/src/qml/filters/audio_adeclick/ui.qml
+++ b/src/qml/filters/audio_adeclick/ui.qml
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2023 Meltytech, LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import Shotcut.Controls as Shotcut
+
+Item {
+    property bool blockControls: false
+
+    function setControls() {
+        if (blockControls)
+            return;
+        sliderWindow.value = filter.getDouble('av.window');
+        sliderThreshold.value = filter.getDouble('av.threshold');
+    }
+
+    width: 200
+    height: 130
+    Component.onCompleted: {
+        if (filter.isNew) {
+            filter.set('av.window', 50);
+            filter.set('av.overlap', 75);
+            filter.set('av.aorder', 2);
+            filter.set('av.threshold', 2);
+            filter.set('av.burst', 2);
+            filter.set('av.method', 'add');
+            filter.savePreset(preset.parameters);
+        }
+        setControls();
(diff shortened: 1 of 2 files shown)


richardpl:
Do note that with "method=add" it will introduce invalid sinusoids in output due to bad non-COLA window generated in code. This have been fixed in master version of FFmpeg. Also burst option is also important, so I wonder why it is not exported too.

bmatherly:
> Also burst option is also important, so I wonder why it is not exported too.

I excluded it because I did not find an easy way to describe it. The description in the documentation is very technical. I'm not really sure I understand it myself. And we want the settings to appeal to non-technical people. If I add it, invariably, some user will ask on the forum what it is and what to set it to. Maybe you can help me come up with a simple description that non-technical users can relate to. Or maybe we need a tool tip that says something like: "low values are good for such-and-such, and high values are better for..."

richardpl:
In simple terms it will connect isolated detected impulsive noises into bigger ones and interpolate all of them, useful for cases when actual impulsive noise spread multiple samples - so if its not simple one sample click but more complex ones spreading multiple samples. This option is required mostly because of how autocorrelation for detection and autoregression modeling of new interpolated samples works. If one uses 0 for burst then it become necessary for filtering out all impulsive noise to perform multiple serial instances of same filter.

src/qml/filters/audio_adeclick/ui.qml
```diff
@@ -0,0 +1,148 @@
...
+    width: 200
+    height: 130
+    Component.onCompleted: {
+        if (filter.isNew) {
+            filter.set('av.window', 50);
+            filter.set('av.overlap', 75);
+            filter.set('av.aorder', 2);
+            filter.set('av.threshold', 2);
+            filter.set('av.burst', 2);
+            filter.set('av.method', 'add');
```
ddennedy, line 41: Per Paul's comment and our version of FFmpeg, I think the method needs to be changed. Did you test with save?

ddennedy (changes requested):

src/qml/filters/audio_adeclick/ui.qml
bmatherly: Save sounds better with my test clip. So I changed it.

bmatherly:
> Do note that with "method=add" it will introduce invalid sinusoids in output due to bad non-COLA window generated in code. This have been fixed in master version of FFmpeg. Also burst option is also important, so I wonder why it is not exported too.

Thanks for the comments. I changed the method to "save" and I added the "Burst Fusion" option. Users do like options.

@ddennedy Do you have any comment about the fact that this filter appears under the "Time" filter category? I think people would naturally look under the "Audio" category. But it appears in Time because it is a link. If someone asks, I thing we can explain that it affects  timing because it has to look ahead at future samples.

ddennedy (approved):

bmatherly:
> Update: I tried a change to facilitate that, but then adding the filter fails:
> 
> [Warning] AttachedFiltersModel::addFilter Failed to load filter "timeremap"
> 
> AttachedFiltersModel::add() depends on accurate QML metadata, and I think it would be ugly to make it lookup MLT metadata here. I think we have to accept this as a quirk bug.

Thanks for trying it.  That is also where I started. We could probably make it right with enough elbow grease. But I don't know if the juice is worth the squeeze.

With the current link design, any filter can be a link by using the [link filter wraper](https://github.com/mltframework/mlt/blob/master/src/framework/mlt_link.h#L76). Perhaps in the future, we will make all filters be links (under the hood). To the user, this would be hidden and the user interface would be the same from their perspective. But it would avoid quirks like this.

PR #1467, opened 2023-08-23
fix export/turn-off proxy if file starts w/ data stream

ddennedy:
Here are the Shotcut changes. This is a little tricky because I want to maintain backwards compatibility with both MLT latest releases that does not include astream/vstream and existing project files that are using audio_index and video_index. So, we cannot convert to the new property names wholesale even if a new/unreleased MLT version is detected. So, if you search the code, there is still a lot of places where "audio_index" and "video_index" are read, and it depends on the MLT change keeping the relative vs. absolute stream values in sync.

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -1517,6 +1517,7 @@ void DetachAudioCommand::redo()
     if (audioClip.is_valid() && videoClip.is_valid()) {
 
         // Disable audio on the video clip.
+        videoClip.set("astream", -1);
         videoClip.set("audio_index", -1);
         // Remove audio filters from the video clip.
         for (int i = 0; i < videoClip.filter_count(); i++) {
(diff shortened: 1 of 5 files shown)


bmatherly:
I get a crash when I try to set the video index to "None" using the letterman file 20090227_233739-H.264.m2ts_1080i_interlaced_letterman_sample.ts

```
[Debug  ] <MLT> [producer avformat] "/home/brian/Videos/broadcast/20090227_233739-H.264.m2ts_1080i_interlaced_letterman_sample.qt" checking VFR: pkt.duration 1001
--Type <RET> for more, q to quit, c to continue without paging--

Thread 50 "shotcut" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff2bfff700 (LWP 70675)]
0x00007fffbe9f3fc9 in decode_audio (self=0x55555c898600, ignore=0x7fff2bffe734, pkt=0x7fff2bffe6c0, samples=1602, timecode=0.96763333333333335, fps=29.970029970029969)
    at /home/brian/shotcut/src/mlt/src/modules/avformat/producer_avformat.c:3154
3154	            pts -= av_rescale_q(self->first_pts,

```
I can do more debugging tomorrow.

bmatherly:
> I get a crash when I try to set the video index to "None"

Fixed by https://github.com/mltframework/mlt/pull/941/commits/8f2b043a3febbb5cd063fe31b6e2f773c281df48

bmatherly (approved):
Looks good to me. No further comments from me.

PR #1473, opened 2023-09-30
Add support for new easing and smooth keyframe types

bmatherly:
Add "Smooth Natural" and "Smooth Tight" keyframe types. Rename "Smooth" to "Smooth Loose". All "Smooth Loose" values are backwards compatible with "Smooth".

Re-implement the keyframe curve line in the keyframes dialog to use the same equation as MLT so the user can see the difference in the curve type in the UI.

image

Depends on https://github.com/mltframework/mlt/pull/957


ddennedy:
I only want one smooth type in Shotcut. We just went through similar with audio clicking and before that the time and speed filters. We discuss an issue and you provide 3 or 4 solutions. Do you know about the dilemma of choice?

ddennedy:
I propose to keep the vast majority of this new code to facilitate editing of all types but only show one of the new ones to use in the menu. Which one is the most suitable for the time filters?

bmatherly:
"Smooth Natural" is the best suited for the time filters (and all filters, in my opinion).

Here is a way that we can keep some backwards compatibility:
* In "new" Shotcut, when the user adds a keyframe and sets the type to "Smooth", use the smooth_natural MLT type.
* In new Shotcut, show the "Smooth" type when either smooth_loose or smooth_natural is detected in the MLT properties
* If the user opens an old MLT in new Shotcut, the smooth_loose keyframes will not  be converted to smooth_natural - they will be unchanged
* If the user opens an old MLT in new shotcut, and then again in old shotcut, any keyframes that were previously smooth_loose will be unchanged. But any smooth keyframes that were added in new shotcut will be "default" detected as linear in the old shotcut.

bmatherly:
> I propose to keep the vast majority of this new code to facilitate editing of all types but only show one of the new ones to use in the menu. 

Done

ddennedy:
How difficult would it be to add MLT version checks so it can be released independently or help distributions that have not yet upgraded MLT?

bmatherly:
> How difficult would it be to add MLT version checks so it can be released independently or help distributions that have not yet upgraded MLT?

Good idea. I will look into that while I am adding the easing types. I will keep this PR as a draft until I have also added the easing  types to it.

bmatherly:
> How difficult would it be to add MLT version checks so it can be released independently or help distributions that have not yet upgraded MLT?

This is done

I have added support for the easing types. Currently, they are presented in the existing right-click menu with icons to illustrate the shape of the curve. I am open minded about making a dialog or some other selection method. But I thought I would start with this and receive usability comments.

image

One topic worth discussing is that some of the easing methods will purposefully calculate a value that is outside of the range between the two points. These easing types include: back, elastic and bounce. The result is that it is possible to set an easing type that will cause a value to go outside of the allowed range. For example, here is the brightess filter that allows values from 0% to 200%. But if you apply back easing to a keyframe with a high or low enough value, it will result in values that go outside of that range.

image
The slider in the filters panel correctly limits the value to the allowed range. But maybe some other filters will result in unexpected results. The good thing is that the keyframe panel shows the line cutoff - so the user has some indication that it is going out of bounds.

ddennedy:
The sub-menu looks good, but let's try a few cosmetic tweaks. The top-level of the context menu can have icons to connect with the graph display: square for Hold, diamond for Linear, circle for Smooth, and a new shape for eases. There are some conventions for ease in-out modifiers of keyframe symbols. We do not need to try represent every variant in the graph view, but it would be nice to indicate when easing is applied. See the main picture on [this page](https://thesbit.com/types-of-keyframes-in-after-effects/) for what After Effects uses:



Looks like with our system we do not to have hold variants of these. It works by nature of the previous or next keyframe. So, just use the Ease In, Ease Out, and Easy Ease (In/Out) shapes for our needs.

For the clamping issue, we will have to review all of the parameters to see what happens unfortunately, and I will help with that.

bmatherly:
I like what you are saying. We already know the shapes for hold, linear, smooth. Here is my proposal for the 3 easing types:

image

The keyframe on the timeline would have one of the 6 shapes shown in the menu. All the  variants of each of the three easing types would show the same icon.

> For the clamping issue, we will have to review all of the parameters to see what happens unfortunately

Back and Elastic are the only ones that can go past the limits. If any filters to not behave with that, we could add a metadata parameter to disallow those easing types (like we used to have with "allowSmooth").

ddennedy:
Perfect. 

> we could add a metadata parameter to disallow those easing types

Go ahead and do that and verify it with one. Then, I will merge it, and we can make fixes and changes for parameters in followup commits. Maybe there will be a mixture of clamping and disallowed easings. 
P.S. If you branch mltframework/shotcut then I can contribute but not while it is a branch in your fork.

bmatherly:
> Go ahead and do that and verify it with one.

Done

> If you branch mltframework/shotcut then I can contribute but not while it is a branch in your fork.

I will keep that in mind for the future. Or, we could do that now if you want to collaborate before it is merged. But I think we would have to create a new PR if we change the  branch.

I will work on changing the keyframe shape in the keyframe dock. When that is done, I will either commit it here, or in master depending on when this is merged.

bmatherly:
I have added new shapes for the keyframe types. I wonder if we should make them a little bigger.

image

I will spend some time testing filters to see how they respond to overshoot.

bmatherly:
Here is an example of overshoot using the timeremap filter with an elastic ease-in keyframe. You can clearly see that the interpolation causes the value to go negative. This is an invalid value (the UI does not let the user set it below zero).
image
Observations:
* The filter UI shows the negative input time calculation
* The line on the keyframe timeline is cut off
* The time remap filter obviously can't provide a negative frame. So it handles the situation by showing the very first frame (the timecode is burned into the video).

I think this is a good way to handle it. We can allow the elastic easing type. If it causes invalid values, the user can clearly see that it is wrong and the filter handles it gracefully.

ddennedy:
I finally had a chance to play with this by sliding text in from the left (similar to the preset with that name). I wanted it to slow down as it arrived to position 0,0. Well, it did not work at first, and it took me awhile to figure out how to make it work. Basically, the concepts are implemented wrong from a UX perspective. Ease-in is supposed to control the timing immediately before the keyframe on which it is applied: easing into the keyframe. Likewise, ease-out is easing out-of,  away-from, or immediately after the keyframe on which it is applied. However, this implementation basically sometimes affects the next keyframe and swaps in and out. I think you are viewing it from the perspective of the line segment following the current keyframe where "in" is the beginning of that segment, and "out" is the end of that segment. That works for the simple hold, linear, smooth cases: "I am going to hold now and then go linear." But to make eases work like other tools (e.g. Premiere and After Effects), for an ease-in on a keyframe I need to use an ease-out on the previous keyframe. And to ease-out on a keyframe I need to use an ease-in.

bmatherly:
Thanks for testing. I understand and agree with your points. But I am not really sure what to do. I did not invent the current keyframe paradigm. I worked within the existing framework. There is nothing I can do to make the keyframe type on a keyframe affect the interpolated samples that precede it - at least not without making major changes to the MLT keyframe behaviors.

> I think you are viewing it from the perspective of the line segment following the current keyframe where "in" is the beginning of that segment, and "out" is the end of that segment.

That's not "how I am viewing it". That's just "how it is already designed" in MLT. Do we want to change that? This is already apparent in the current release version: Add a single keyframe to a filter. Change the keyframe type to "linear". Observe that absolutely nothing changes. Why not? Because the interpolation is only applied to the values between two keyframes. So the last keyframe in any sequence can't have a "type".

Here is a proposal for how we could represent things differently in Shotcut without changing MLT:
1) Change the keyframe indicator to be a neutral shape for all types (maybe a cross)
2) Change the keyframe indicator to remove the "type" and only allow moving and removing keyframes
3) Add a new "interpolation" indicator that is rendered at the midpoint of the line segment between keyframes.
4) The interpolation indicator will use the existing and new shapes to represent the interpolation types (linear, smooth, easing, etc)
4) Implement the "keyframe type" context menu to be triggered when the interpolation indicator is right clicked
5) When the "interpolation type" is changed, Shotcut will change the "keyframe_type" value on the keyframe that precedes the line segment that the user is modifying

Effectively, this proposal would be to make the interpolation method apply to the line segment between keyframes, and not the keyframes themselves. 

Let me know your thoughts on this. If we do not like it, I can try to brainstorm other ideas.

ddennedy:
I was afraid that and not surprised, but I had hope that would figure out something that is not major. I have an idea on how to fix ease-in and -out but it fails to address Ease In/Out unless that can be somehow split apart. But basically, the idea is to swap "in" with "out". And then in MLT, to ease-in on keyframe N, do what we today for ease-out on keyframe N-1. In case, N is 0, it can be a no-op since the values preceding it are not different. So, forgetting about in/out for the moment, does that seem feasible? I see `interpolate_item()` is passed previous and next items.

ddennedy:
Or, put another way, when interpolating item N, see if there is an ease-in on item N + 1. If so, do that instead. On item N + 1, if it has an ease-in, do a linear interpolation if there is no ease in/out.

bmatherly:
I am not sure if I understand your suggestion. Let me use an illustration to see if I understand...

Consider a sequence with keyframes k1-k3 - and line segments s1 & s2 between them:
k1 - s1 - k2 - s2 - k3

The user right clicks on the keyframes and sets the following types:
k1 = smooth
k2 = ease-in
k3 = smooth

In this scenario:
* s1: MLT would notice the ease-in on k2 and perform ease-in.
* s2: MLT would notice the ease-in on k2 and perform linear

I think the user would find this unexpected because neither of the smooth types would be performed. And why would they want linear interpolation on s2?

If in/out is too complicated, we can just remove it. in/out can always be performed by adding a keyframe between two keyframes and putting an ease-in before and an ease-out after it.

ddennedy:
Ease in and out are typically used with linear interpolation, which is why Adobe tools have hold-based variations of eases. One uses linear keyframes and eases to simplify things instead of smooth or bezier keyframes to make custom curves. In your example, the smooth k3 is for the segment following it. And the smooth on k1 is overridden by the ease-in. Basically, smooth plus ease-in is not supported, and that is less bizarre than this PR's UX. If you really wanted that something like you can add another smooth keyframe between k1 and k2. (As for the hold variants, one can do a hold-ease-out by making k1 hold and k2 ease-out. There would be no direct way to do ease-in-hold; rather, k2 would be ease-in and k3 would be hold based on the same value.)

ddennedy:
Hey, I started implemented my idea in local branches sans the curve display in Shotcut. So far it is working good. I also have some idea how to do Ease In/Out now. I will share them soon so you can see and feel what I mean.

ddennedy:
See

https://github.com/mltframework/mlt/commit/9b0232174b5fca34a00c621601e0e1119b3d3f28

and

https://github.com/mltframework/shotcut/commit/71ac2e7c0d51f9094da35e0449eac6602b7b598d

(no curves UI update for Shotcut yet)

bmatherly:
> Hey, I started implemented my idea in local branches sans the curve display in Shotcut. So far it is working good. I also have some idea how to do Ease In/Out now. I will share them soon so you can see and feel what I mean.

Thanks Dan. I need to take some time to understand your patches.

In the meantime, can you explain more why you see the in and out swapped?
I graphed all the easing curves here: https://github.com/mltframework/mlt/pull/957#issuecomment-1749817584
And they match curve illustrations here: https://easings.net/
Also, the easing calculations are based on a reputable source: http://robertpenner.com/easing/

ddennedy:
I already explained how ease in does not ease into a keyframe and ease out does not ease out of a keyframe. This is currently backwards from the way all other tools work. You should watch a YouTube tutorial about ease keyframes. It does not matter if they look correct to you from a coding perspective. I won’t release this in its current form.

ddennedy:
After expanding my review of literature and how to videos of other tools I might have to take back what I said. The problem is terminology and relativity: in and out relative to what? As one reviews this, you will find that sometimes the words are used inconsistently by a presenter because it can be confusing. It seems traditionally (and the way easing is done/explained in your articles and CSS for the Web) that easing describes the start and end of an animation where the minimal animation is 2 states or keyframes: begin and end. In that case, in and out are relative to the animation or span between states: ease into the change and east out of the change. It seems Adobe redefined these for its "Easy" ease keyframe types. Maybe that is why they call them "Easy" or perhaps simply because they are predefined. Final Cut and Davinci Resolve do not have builtin ease in and out and thus require simulation which gets into the gray areas when tutorials describe them. In Adobe tools, in and out is relative to the keyframe as I described above. So which should we use? I worry because like I said in my first reply, at first I struggled to make it work. And I suspect most people think about it like I did and the way Adobe did it.
Maybe we should be consistent with Glaxnimate. It is sensible to be consistent with our own bundled tool. Sigh, it follows your approach:
https://www.youtube.com/watch?v=Jbq4aq4Unb0

It looks like we will have a lot of support and education ahead of us around this messy subject.

bmatherly:
Thanks for your thoughtful response. And thanks for challenging me to think more critically about this feature. I agree that it is not ready to merge yet.

In Adobe, at least, it seems that the "ease" menu entries are really just convenience shortcuts to modify control points in the underlying beizer curve. Users can always pop open the curve editor and edit the curves directly to get the same effect. The curve editor, by the way, edits the line segment between two keyframes. I think that the keyframe "icon" is really just a status display of the beizer control points immediately preceding and following that particular keyframe.

My renewed thinking is that when a user manipulates a key frame, they should be able to modify the curve before and after the keyframe - like you said: "relative to what?". So maybe it would be "Ease in from previous" and "Ease out to next". I think I could modify the keyframe indicators to hint at the curve type before and after the keyframe.

Also, I think we will not expose in/out. If someone wants in/out, they will have to add an intermediate keyframe and do "in from previous" before and "out to next" after.

bmatherly:
I think this is ready for review again. I believe I have resolved the review comments.

Keyframe indicators are not double-sided to indicate the interpolation between keyframes
The context menu now allows editing previous and next

image

image

PR #1480, opened 2023-10-20
Port from qAsConst to std::as_const

BSDKaffee:
qAsConst() has been marked deprecated in Qt 6.6.


ddennedy:
I need to test this with the g++ 10 that I use in our Linux builds for compatibility reasons.

PR #1483, opened 2023-11-01
Fix transition properties are not restored when undo trim out

bmatherly:
I would love a quick visual sanity check on the approach for this one. Maybe there is a better way.

Steps to reproduce:
1) Put two clips on the timeline - drag one over the other to make a transition
2) Set a property on the transition (e.g. Barn Door Horizontal)
3) Trim the left or right side of the transition until it is removed
4) Edit->undo
5) Observe that the property is not restored (e.g. dissolve instead of Barn Door Horizontal)

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -1137,12 +1137,13 @@ bool AddTransitionByTrimInCommand::mergeWith(const QUndoCommand *other)
 }
 
 RemoveTransitionByTrimInCommand::RemoveTransitionByTrimInCommand(MultitrackModel &model,
-                                                                 int trackIndex, int clipIndex, int delta, bool redo, QUndoCommand *parent)
+                                                                 int trackIndex, int clipIndex, int delta, QString xml, bool redo, QUndoCommand *parent)
     : TrimCommand(parent)
     , m_model(model)
     , m_trackIndex(qBound(0, trackIndex, qMax(model.rowCount() - 1, 0)))
     , m_clipIndex(clipIndex)
     , m_delta(delta)
+    , m_xml(xml)
     , m_redo(redo)
 {
     setText(QObject::tr("Remove transition"));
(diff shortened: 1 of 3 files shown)


src/commands/timelinecommands.cpp
```diff
@@ -1202,6 +1223,25 @@ void RemoveTransitionByTrimOutCommand::undo()
     if (m_clipIndex > 0) {
         LOG_DEBUG() << "trackIndex" << m_trackIndex << "clipIndex" << m_clipIndex << "delta" << m_delta;
         m_model.addTransitionByTrimIn(m_trackIndex, m_clipIndex, m_delta);
+        // Copy properties from old transition to new transition
+        std::unique_ptr<Mlt::ClipInfo> clipInfo = m_model.getClipInfo(m_trackIndex, m_clipIndex);
```
ddennedy, line 1227: Simplify to `auto clipInfo =`...

src/commands/timelinecommands.cpp
```diff
@@ -1202,6 +1223,25 @@ void RemoveTransitionByTrimOutCommand::undo()
...
+        std::unique_ptr<Mlt::Service> oldService(new Mlt::Producer(MLT.profile(), "xml-string",
```
ddennedy, line 1228: Simplify to `Mlt::Producer oldService(MLT.profile(), "xml-string",`...
Why must it use pointers and `Service`?

src/commands/timelinecommands.cpp
```diff
@@ -1202,6 +1223,25 @@ void RemoveTransitionByTrimOutCommand::undo()
...
+                                                                   m_xml.toUtf8().constData()));
+        while (oldService && oldService->is_valid()) {
+            if (oldService->type() == mlt_service_transition_type) {
+                Mlt::Transition oldTransition(*oldService);
+                std::unique_ptr<Mlt::Service> newService(new Mlt::Producer(clipInfo->producer));
```
ddennedy, line 1233: Simplify to `Mlt::Producer newService(clipInfo->producer);`

src/commands/timelinecommands.cpp
```diff
@@ -1202,6 +1223,25 @@ void RemoveTransitionByTrimOutCommand::undo()
...
+                while (newService && newService->is_valid()) {
+                    if (newService->type() == mlt_service_transition_type &&
+                            QString(oldService->get("mlt_service")) == QString(newService->get("mlt_service"))) {
+                        newService->inherit(*oldService.get());
+                        break;
+                    }
+                    newService.reset(newService->producer());
```
ddennedy, line 1240: Given the above simplification this becomes `newService = newService.producer();`

src/commands/timelinecommands.cpp
bmatherly: I copied working code from somewhere else. I don't mind changing it. But it gets complicated because "get_producer()" actually returns a Mlt::Service pointer that needs to be freed by the caller. I guess I can do this:

```
Mlt::Service *tmpNewService = newService.producer();
newService = Mlt::Producer(*tmpNewService);
delete tmpNewService;
```

I'll put a patch together.

src/commands/timelinecommands.cpp
bmatherly: Done

src/commands/timelinecommands.cpp
bmatherly: It has to be service because of a bug in Producer that I just pushed to MLT. But this patch should stay with Service so that it works with released versions of MLT.

src/commands/timelinecommands.cpp
bmatherly: Done except still service

src/commands/timelinecommands.cpp
bmatherly: Changed to this to avoid memory leak:

```
Mlt::Service *tmpNewService = newService.producer();
newService = Mlt::Producer(*tmpNewService);
delete tmpNewService;
```

PR #1487, opened 2023-11-15
Fix translation error in shotcut_zh_CN.ts

carezkh:


ddennedy:
You must submit translation updates through transifex.com

PR #1488, opened 2023-11-28
add undo/redo for filters and keyframes

bmatherly:
Based on this previous patch https://github.com/mltframework/shotcut/tree/undo_filters2

I had to change some tactics to accommodate changes since the original branch was created. I will keep testing. But I think this is ready for review.


bmatherly:
I have completed my testing on this. I think it would be great to merge it into master to avoid future merge conflicts.

PR #1492, opened 2023-12-06
Add timeline looping on markers

bmatherly:
As requested here:
https://forum.shotcut.org/t/loop-region-for-playback/32123/16

Add an option to the context menu to loop/unloop around a range marker.
image

When range marker looping is enabled, the timeline is highlighted around that region and the cursor can not seek outside of that region.
image

Depends on https://github.com/mltframework/mlt/pull/964

### DIFF
--- a/src/docks/timelinedock.cpp
+++ b/src/docks/timelinedock.cpp
@@ -60,7 +60,10 @@ TimelineDock::TimelineDock(QWidget *parent) :
     m_ignoreNextPositionChange(false),
     m_trimDelta(0),
     m_transitionDelta(0),
-    m_blockSetSelection(false)
+    m_blockSetSelection(false),
+    m_loopMarker(-1),
+    m_loopStart(-1),
+    m_loopEnd(-1)
 {
     LOG_DEBUG() << "begin";
     m_selection.selectedTrack = -1;
@@ -133,6 +136,7 @@ TimelineDock::TimelineDock(QWidget *parent) :
     markerMenu->addAction(Actions["timelineDeleteMarkerAction"]);
     markerMenu->addAction(Actions["timelineMarkSelectedClipAction"]);
     markerMenu->addAction(Actions["timelineCycleMarkerColorAction"]);
+    markerMenu->addAction(Actions["timelineToggleLoopMarkerAction"]);
     m_mainMenu->addMenu(markerMenu);
     Actions.loadFromMenu(m_mainMenu);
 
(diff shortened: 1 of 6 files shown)


ddennedy:
You can’t seek outside? Hmm that’s weird. Most looping controls like this only affect playback.

bmatherly:
> You can’t seek outside? Hmm that’s weird.

Yes, we can. And I am willing to do that. But there is a lag due to frames being pre-rendered and presented by the consumer worker threads before they are provided to the application. Putting the loop logic in the seek function guarantees frame accuracy. So I am posting this proposal for exactly this conversation :) I don't feel strongly either way.

> Most looping controls like this only affect playback.

That is why I set the internal properties to be private (underscore prefix). The seek function in MLT is implemented in the producer. I always wondered why the consumer does not have a seek. But I see this function as an advanced form of seek - which is a player control action.

Another advantage of implementing the loop in the producer is that the MLT controller in Shotcut doesn't have to be aware of which producer is in the player and apply/unapply the looping depending on whether it is a timeline producer, playlist producer or clip producer.

bmatherly:
> You can’t seek outside? Hmm that’s weird.

I completely misunderstood your comment. What would be best to happen if the user tries to seek outside the loop range? It would display that one frame, but if they press play it would jump to the loop range? I am open mined about this. This PR is just a starting point.

ddennedy:
> What would be best to happen if the user tries to seek outside the loop range?

You can review what a popular media player or audio editor whose UI that you respect does. 

> if they press play it would jump to the loop range?

I think that is decent behavior. What Audacity does is if the playhead is before the region it plays from the playhead and starts looping when it reaches the end of the region. If the playhead is after the region it plays from the playhead to the end. I like that too. 

Yet another option I think is closest to [what I envisioned for the road map](https://forum.shotcut.org/t/looping-playback-design/36014): add new looping playback modes and not a marker context menu item. For example, Shift+Space would loop the current or next closest region in the timeline or in-out in the Source player (Loop Region). Ctrl+Space could loop over the current timeline clip or in-out in the Source (Loop Clip). Ctrl+Shift+Space could loop over the entire Source or Project (Loop All). These could be in a drop-down menu on the play button. Then, whenever you seek, playback stops as it currently does, and there is no more looping playback until you choose that action.

bmatherly:
Canceled in favor of this https://github.com/mltframework/shotcut/pull/1500

PR #1494, opened 2023-12-18
Add multi-split to timeline clips

bmatherly:
Add a new action (SHIFT + S) to split all clips under the playhead. As requested here:
https://forum.shotcut.org/t/split-across-multiple-tracks/293
https://forum.shotcut.org/t/automating-keyboard-shortcuts-with-autohotkey/12608/10
https://forum.shotcut.org/t/easier-workflow-solution/1456/7
https://forum.shotcut.org/t/propagation-on-tracks/39448/4
https://forum.shotcut.org/t/editing-basics-ripple-multiripple/20488/7

Also, add support for multiple-selection to the existing split command. Now, Shotcut will first look for one or more selected clips under the playhead and split those. If not selected clips on under the playhead, then Shotcut will revert to the current behavior to select a clip based on track selection.

Also added menu entry to the timeline menu
image

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -808,31 +808,37 @@ bool TrimClipOutCommand::mergeWith(const QUndoCommand *other)
     return true;
 }
 
-SplitCommand::SplitCommand(MultitrackModel &model, int trackIndex,
-                           int clipIndex, int position, QUndoCommand *parent)
+SplitCommand::SplitCommand(MultitrackModel &model, const std::vector<int> &trackIndex,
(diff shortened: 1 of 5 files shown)


ddennedy:
While reviewing this I was a little surprised to find the Split commands in the Selection submenu. We may have had a discussion about this, but I forget. I was looking first in Edit. The Split commands are drowning in a pool of menu items that all start with capital S and "Select" making them harder to spot. Do you agree to move them to the Edit submenu?

bmatherly:
> Do you agree to move them to the Edit submenu?

I had exactly the same thought. But I didn't want to mix the change in with this review. Done in f62db92c138b295778099117e6e1f156a392ca9a

PR #1501, opened 2024-01-01
Add clip nudging to the timeline

bmatherly:
A nudge moves the clip forward or backward one frame

Looking for some discussion on this idea. Dragging on the timeline does not work well in tiny increments. I know there are some workaround ways to move clips in small increments using snapping. But I do not find those to be intuitive.

I propose these two new nudge commands (forward and backward) that will move the selected clips by one frame. In this proof-of-concept, I added two commands to the hamburger and clip context menus.

image

I chose shortcuts "N" and "SHIFT+N" for forward and  backward respectively.

One disadvantage to this is that when a move command occurs, the current selection is cleared. This makes it difficult to easily move a clip 3 frames (for example) by typing N+N+N - the user has to reselect the clip between each nudge. I think this might have been done to work around a problem with moving multiple clips across tracks. I am not sure. Maybe I can find a way to improve that.

Maybe users will request a way to specify how many frames should be in a nudge. If we can make quick repetition of nudges work, I do not think that would be worth the effort.

### DIFF
--- a/src/docks/timelinedock.cpp
+++ b/src/docks/timelinedock.cpp
@@ -637,9 +637,24 @@ void TimelineDock::setupActions()
     connect(action, &QAction::triggered, this, [&]() {
         auto selectedClips = selection();
         if (!selectedClips.isEmpty()) {
+            if (selection().size() > 1)
+                setSelection({selection().first()});
             int trackIndex = selection().first().y();
             int clipIndex = selection().first().x();
+            bool valid = clipIndex > 0 && !isTransition(trackIndex, clipIndex)
+                         && !isBlank(trackIndex, clipIndex);
+            if (valid && !Settings.timelineRipple()) {
+                valid = isBlank(trackIndex, clipIndex - 1) && ((clipIndex == clipCount(trackIndex) - 1)
+                                                               || isBlank(trackIndex, clipIndex + 1));
+            } else if (valid) {
+                valid = clipIndex < clipCount(trackIndex) - 1;
+            }
+            if (!valid) {
+                emit showStatusMessage(tr("Nudge Forward is not available"));
+                return;
+            }
             auto clipInfo = m_model.getClipInfo(trackIndex, clipIndex);
+            TimelineSelectionBlocker selectBlocker(*this);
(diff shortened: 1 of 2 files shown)


ddennedy:
Nudging is tricky, partly because multiple in succession is expected. You can add a `TimelineSelectionBlocker` inside each new action lambda to prevent clearing the selection, but the problem is that `clipIndex` changes if a gap is opened or transition added.  It generally works better with ripple on because if a gap opens to the left, the blank is selected, and nudging to the right works to somehow grow the blank and effectively continue to nudge the originally selected clip right, but not ripple-nudging a blank left. However, if multiple clips are selected other weirdness can occur. You should add that `TimelineSelectionBlocker` and do more testing to see what I mean. I think the selection clearing or not sticking to the originally selected clip might be a blocker because otherwise it will cause a lot of bug reports or complaints.
Maybe some restrictions can be added such as:
- Break up a multiple selection and select only the first selected clip.
- If Ripple is off, you can only nudge a clip that has blanks on both sides (treat end-of track as blank too).
- If Ripple is on, you can only nudge a clip backward if there is a blank on the left.
I think these remedy the problems I found if they are not too confusing. Basically, we block nudge from adding a single frame transition (or the clip index from changing). When nudge is blocked, display a status message like "Nudge Forward is not available".

For keyboard shortcuts, in the past I was thinking of using comma for backward and period for forward. At least on English keyboards these correspond to the < > keys (but those shortcuts are for marker seeking). But N is sensible too, I think.

ddennedy:
I started implementing these rules in my working copy if you want to wait and evaluate.

bmatherly:
I think those rules make sense. For me, personally, nudging is most important for a single selected clip. Maybe to keep it simple, nudging should only be enabled for a single selected clip. And nudging should NEVER cause a transition to be created. So it could fail out if there is not blank space following the clip. It might be easier to support to have simple limitations on the feature rather than trying to explain a complex combination of constraints to users.

ddennedy:
OK, give these changes a try. Happy New Year

bmatherly:
Thanks Dan. This works great for me. Happy new year to you as well!

PR #1503, opened 2024-01-09
Add clip grouping

bmatherly:
This implements the proof of concept for clip grouping.

Add a new action to the context menu to group/ungroup with shortcut "CTRL+G"
image

A clip can only be a member of one group.

When selecting in the timeline, I have added a new modifier: hold ALT key to ignore all groups. So for example, click on clip to select the clip and all clips in the group. Hold ALT and click on the clip to only select the clip.

The group parameter is saved in the XML as a property on the playlist clip:

```
    <entry producer="producer0" in="00:00:00.000" out="00:00:03.967">
      <property name="shotcut:group">0</property>
    </entry>
```

Not yet implemented:
* Make the filter and properties panel smarter to not show anything when a group is selected
* Automatically create a group when detaching audio

### DIFF
--- a/src/commands/timelinecommands.cpp
+++ b/src/commands/timelinecommands.cpp
@@ -38,6 +38,32 @@ Mlt::Producer *deserializeProducer(QString &xml)
     return new Mlt::Producer(MLT.profile(), "xml-string", xml.toUtf8().constData());
 }
 
+int getUniqueGroupNumber(MultitrackModel &model)
+{
+    QSet<int> groups;
+    for (int trackIndex = 0; trackIndex < model.trackList().size(); trackIndex++) {
+        int i = model.trackList().at(trackIndex).mlt_index;
+        QScopedPointer<Mlt::Producer> track(model.tractor()->track(i));
+        if (track) {
+            Mlt::Playlist playlist(*track);
+            for (int clipIndex = 0; clipIndex < playlist.count(); clipIndex++) {
+                QScopedPointer<Mlt::ClipInfo> info(playlist.clip_info(clipIndex));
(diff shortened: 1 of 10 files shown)


ddennedy:
It is working good for me. I changed "UnGroup" to "Ungroup". I played around with changing the text of the action to either "Group" or "Ungroup" but then we'd need to consolidate strings as it needs to be set in a few places. Switch back to the simplicity of "Group/Ungroup"

ddennedy:
One problem I found is that after moving a group of clips they are no longer a group. I think moving will be the most commonly used operation with groups.

bmatherly:
> I found is that after moving a group of clips they are no longer a group

I think I have that fixed and also undo remove.

I am also noticing some unexpected behavior with the clip selection after an operation is performed. For example, click on a group, move the selection of clips. After the move, nothing is selected. I would prefer they stay selected. That is kind of minor. 

I have found other situations that make the selection inconsistent. For example: detach audio from a clip, click a clip to select the new group of two clips, click undo. Now the clips are back together and the original clip is selected. Now right click the clip - detach is not available. Why not? Because the selection changed signal was never sent due to the selection blocker on all undo operations. I am trying to decide if these selection quirks are worth digging into. Probably not for this PR.

ddennedy:
I think we could work on some of the quirks during beta or even the next release if the changes seem risky.

bmatherly:
Regarding this idea:
> Make the filter and properties panel smarter to not show anything when a group is selected

We had discussed maybe showing a list in the properties panel. But I realize that if multiple clips are selected, they might not even by the same type (avformat, color, etc).

So I propose this simpler suggestion to blank them out if multiple clips are selected. I also tweaked the message in the filters panel to be more appropriate for 0 or multiple clips selected.
image

PR #1518, opened 2024-02-18
Fix crash when switching proxy on/off

bmatherly:
The close() signal from the multitrack model eventuallys causes th MLT producer to close. So do not call it when loading.


ddennedy:
I cherry picked db91b88 into master, switched to this branch, and rebased it on master.

bmatherly:
Good plan. I think the problem with the progress bar is that some signals are sneaking out and calling UI code outside of the UI thread. For my large test project, the progress bar is less important with this optimization change which cuts the loading time to about 1/3. So I think it is OK to leave the progress bar out.

PR #1520, opened 2024-02-23
Add configurable time format

bmatherly:
This pull request is for review and staging until after the next release.

Add a new menu entry for time selection
image

This entry affects everything except for the rulers in the player and timeline
Here is a screenshot with some examples:
image

### Notes from the Design Discussion:

 **default to time clock format:**

Not done - currently defaults to SMPTE DF - which was the previous display. I can change it, but I am concerned that some people will be surprised when they upgrade Shotcut and the time display changes.

 **context menu on the time display:**

Not implemented. I made a Settings menu entry. I could also add a context menu on the time entry text field in the play if that is helpful. But people should be aware that it affects the entire application.

 **try to affect everywhere time is displayed:**

Done - with the exception of the Rulers. I can also change the rulers. But I notice that there is some "fudge factor" logic that we would have to overcome. I think that is probably to get around rounding errors.

**input parsing as well based on MLT’s interpretation of string to time:**

Done

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -206,10 +206,10 @@
      <property name="title">
       <string>Time Format</string>
      </property>
-     <addaction name="actionTimeFrames"/>
      <addaction name="actionTimeClock"/>
      <addaction name="actionTimeDF"/>
      <addaction name="actionTimeNDF"/>
+     <addaction name="actionTimeFrames"/>
     </widget>
     <widget class="QMenu" name="menuProxy">
      <property name="title">
@@ -283,11 +283,11 @@
     <addaction name="menuBackup"/>
     <addaction name="actionUser_Interface"/>
     <addaction name="menuPreviewScaling"/>
-    <addaction name="menuTimeFormat"/>
     <addaction name="menuProxy"/>
     <addaction name="menuPlayerSettings"/>
     <addaction name="menuPlaylist"/>
     <addaction name="menuTimeline"/>
+    <addaction name="menuTimeFormat"/>
     <addaction name="menuLanguage"/>
     <addaction name="menuTheme"/>
     <addaction name="menuDrawingMethod"/>
@@ -1183,7 +1183,7 @@
     <bool>true</bool>
    </property>
    <property name="text">
-    <string>Drop Frame</string>
+    <string>Timecode (drop Frame)</string>
    </property>
   </action>
   <action name="actionTimeFrames">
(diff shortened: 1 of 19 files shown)


bmatherly:
@ddennedy  do you have a minute to review this? I would like to merge it. But I would like your comments since I deviated from the original design discussion somewhat.

ddennedy:
No I’m traveling

ddennedy:
> Add a new menu entry for time selection

I prefer this over only the context menu. However, I think "Drop Frame" sounds too scary on its own and both drop and non-drop should end with "Timecode."

> I can also change the rulers. 

Well, the rulers are special as they use timecode without frames. I think it is good to leave them as-is.

> currently defaults to SMPTE DF

I agree to not change the default. But mine defaulted to Non-Drop Frame.

src/mainwindow.ui
```diff
@@ -202,6 +202,15 @@
      <addaction name="actionPreview540"/>
      <addaction name="actionPreview720"/>
     </widget>
+    <widget class="QMenu" name="menuTimeFormat">
+     <property name="title">
+      <string>Time Format</string>
+     </property>
+     <addaction name="actionTimeFrames"/>
+     <addaction name="actionTimeClock"/>
```
ddennedy, line 210: Please move "Time Format" to just below "Timeline" in the Settings menu. And move "Frames" to the bottom of the sub-menu.

src/settings.cpp
```diff
@@ -1278,3 +1278,14 @@ void ShotcutSettings::setBackupPeriod(int seconds)
 {
     settings.setValue("backupPeriod", seconds);
 }
+
+mlt_time_format ShotcutSettings::timeFormat() const
+{
+    return (mlt_time_format)settings.value("timeFormat", mlt_time_smpte_ndf).toInt();
```
ddennedy, line 1284: Change default to `mlt_time_smpte_df`

ddennedy (changes requested):

bmatherly:
> However, I think "Drop Frame" sounds too scary on its own and both drop and non-drop should end with "Timecode."

Done. I put a different spin on it. Let me know what you think.

> I agree to not change the default. But mine defaulted to Non-Drop Frame.

Fixed

image

src/settings.cpp
bmatherly: Done

src/mainwindow.ui
bmatherly: Done

PR #1525, opened 2024-03-24
Separate filter undo commands for parameters and keyframes

bmatherly:
This is currently a work in progress for review and discussion:

Start a new undo command for individual parameters and keyframes in filters. This method works for all keyframs in the keyframe panel.

Tracking of individual parameters is only implemented for the color grading filter for demonstration purpose. We would have to go to every filter qml file and add the start/end function calls to make it work for all filters. I am willing to do this, but want to make sure this method is agreeable, first.

### Parameters in the filter panel
* History text: "Change Color Grading filter: Shadows (Lift)"
* Merging: only merge if it is the same filter and the same parameter (the history text must match)

### Keyframe add and remove
* History text: "Change Brightness filter: add keyframe"
* Merging: never merge

### Modify (move) keyframe in the keyframes panel
* History text: "Change Color Grading filter: modify keyframe"
* Merging: only merge if it is the same filter, the same parameter, and the same keyframe

Here is a demonstration:

https://github.com/mltframework/shotcut/assets/821968/ea6f0ca7-67e9-4157-858f-2f5f81a4f4fd

### DIFF
--- a/src/qmltypes/qmlfilter.cpp
+++ b/src/qmltypes/qmlfilter.cpp
@@ -180,7 +180,7 @@ void QmlFilter::set(QString name, QString value, int position)
         if (qstrcmp(m_service.get(qUtf8Printable(name)), qUtf8Printable(value)))  {
             m_service.set_string(qUtf8Printable(name), qUtf8Printable(value)) ;
             emit changed(name);
-            updateChangeCommand(name);
+            updateUndoCommand(name);
         }
     } else {
         // Only set an animation keyframe if it does not already exist with the same value.
@@ -189,7 +189,7 @@ void QmlFilter::set(QString name, QString value, int position)
                 || value != m_service.anim_get(qUtf8Printable(name), position, duration())) {
             m_service.anim_set(qUtf8Printable(name), qUtf8Printable(value), position, duration());
             emit changed(name);
-            updateChangeCommand(name);
+            updateUndoCommand(name);
         }
     }
 }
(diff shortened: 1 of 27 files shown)


src/qmltypes/qmlfilter.h
```diff
@@ -139,13 +136,21 @@ class QmlFilter : public QObject
...
     Q_INVOKABLE void crop(const QRectF &rect);
-    void startUndoTracking(FilterController *controller, int row);
+
     Q_INVOKABLE void copyParameters();
     Q_INVOKABLE void pasteParameters(const QStringList &propertyNames);
 
+    // Functions for undo/redo
+    void startUndoTracking();
+    Q_INVOKABLE void startChangeParameterCommand(const QString &desc = QString());
+    void startChangeAddKeyframeCommand();
```
bmatherly, line 146: For these "start" functions, I was thinking about calling them something else. Some ideas I had:
* StartUndoKeyframeMacro()
* StartUndoKeyframeSequence()
* StartUndoKeyframeCommand()

I am open to suggestions.

We would want to make sure the "end" command name matches the start name.

src/qmltypes/qmlfilter.cpp
```diff
@@ -770,27 +763,121 @@ int QmlFilter::keyframeIndex(Mlt::Animation &animation, int position)
...
+    }
+}
+
+void QmlFilter::startChangeParameterCommand(const QString &desc)
+{
+    if (!m_previousState.count()) {
+//        LOG_DEBUG() << "Undo tracking has not started yet";
+        return;
+    }
+    m_changeInProgress++;
```
bmatherly, line 784: I found that the filter changes can end up nesting start commands. For example, modifyig a parameter might trigger a callback that will modify other parameters. So the changeInProgress flag is really a counter to keep track of when the nested calls have been unrolled.

PR #1535, opened 2024-05-15
use DJI LRF (MP4) as proxy if available

ManiacTwister:
DJI cameras create low resolution mp4 files using the same name as the full video file just with the extension replaced to LRF. They can be used as proxy, similar to gopros LRV files.


ddennedy:
merged in squashed merge commit where I fixed the astyle failure 121133f6

PR #1538, opened 2024-05-22
remove redundant function definitions

dyliuti:
【QString removeFileScheme(QUrl &url);】is defined and implemented in Util


ddennedy:
thank you

PR #1540, opened 2024-05-24
 remove member variable; move func execution time

dyliuti:


dyliuti:
fix bug: select the non-first filter -> undoredo or [mininue mainwodow, show mainwindow] or switch to filterdock -> reselect the first filter
fix bug: switch to filterdock -> flicker
Test: close filterdock and open it again  -> ok

PR #1541, opened 2024-05-24
fix bug: flicker and reselect the first fitler

dyliuti:
online bug: slect no-first filter at filterdocker -> undoredo or [miniusize mainwindow, show mainwindow] or switch back to filterdock -> reselect the firset filter
online bug: switch back to filterdock -> fliterdocker flicker
online bug: click the top or bottom of filteritem and then auto select first item
test: close filterdock and open again -> ok


ddennedy:
Changes to FilterDock are not accepted. The commit history is not clean (has an old commit and a merge).

PR #1545, opened 2024-06-05
Fix memory leak

bmatherly:
producer->profile returns a new object that is never freed

### DIFF
--- a/src/dialogs/filedatedialog.cpp
+++ b/src/dialogs/filedatedialog.cpp
@@ -22,6 +22,8 @@
 #include "MltProducer.h"
 #include "proxymanager.h"
 
+#include <MltProfile.h>
+
 #include <QComboBox>
 #include <QDateTimeEdit>
 #include <QDebug>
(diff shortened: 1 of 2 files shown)


src/dialogs/filedatedialog.cpp
```diff
@@ -113,7 +115,8 @@ void FileDateDialog::populateDateOptions(Mlt::Producer *producer)
     }
 
     // Add metadata dates
-    Mlt::Producer tmpProducer( *(producer->profile()), "avformat", resource.toUtf8().constData() );
+    Mlt::Profile profile(mlt_profile_clone(producer->get_profile()));
+    Mlt::Producer tmpProducer( profile, "avformat", resource.toUtf8().constData() );
```
ddennedy, line 119: Why not use `MLT.profile()` here like we do so many other places?
There is not supposed to be spaces inside the parenthesis.

ddennedy:
`TextProducerWidget` also has a place where it should be using `MLT.profile()`:
```
--- src/widgets/textproducerwidget.cpp
+++ src/widgets/textproducerwidget.cpp
@@ -146,8 +146,7 @@ void TextProducerWidget::loadPreset(Mlt::Properties &p)
         filter.reset(MLT.getFilter(kRichFilterName, m_producer.data()));
         if (filter && filter->is_valid())
             m_producer->detach(*filter);
-        Mlt::Profile profile = m_producer->get_profile();
-        filter.reset(createFilter(profile, m_producer.data()));
+        filter.reset(createFilter(MLT.profile(), m_producer.data()));
         m_producer->attach(*filter);
         emit producerChanged(m_producer.data());
     }
```
This is the only other problematic usage of `Mlt::Service::profile()` I found.

src/dialogs/filedatedialog.cpp
bmatherly: Not a compelling reason - but so much depends on mltcontroller.h. If I change one small thing in mltcontroller.h, the whole application has to recompile. I think we should prefer MLT.profile() over Mlt::Service::profile() in all cases.

bmatherly: > There is not supposed to be spaces inside the parenthesis.

done

bmatherly:
> TextProducerWidget also has a place where it should be using MLT.profile()

Done

PR #1552, opened 2024-07-02
Add new subtitle panel

bmatherly:
Start in the Subtitles Panel by adding a subtitle "Track":
image
For each set of subtitles, I called them "tracks". But maybe that is confusing. In MLT I called them a "Feed". I am open to suggestions about this.

Name the track and specify the language:
image

Add subtitles with the buttons or the keyboard shortcuts:
image

The keyboard shortcuts need help. I chose a series of keys that are close to each other on the keyboard.

When subtitles are present, they will appear in the new subtitle bar in the Timeline:
image

To add subtitle overlay burn-in, add a "Subtitle Burn In" filter to the output. This filter is only available for the output - not clips or tracks.
image

Multiple burn-in filters can be added simultaneously - and they can each be burning in a different track.
image

But only one subtitle track can be active at once for editing:
image


src/models/subtitles.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef SUBTITLES_H
+#define SUBTITLES_H
+
+#include <string>
+#include <vector>
+
+namespace Subtitles {
+
+struct SubtitleItem {
+    int64_t start;
```
ddennedy, line 27: int64_t is an undefined type here and needs `#include <cstdint>`

ddennedy:
Some thoughts:
- Most people will not care about subtitle name and language; just let me create some timed text. Instead of initializing **Add Subtitle Item** disabled, it can be enabled and when you click it without no subtitle, automatically make one using the same defaults as the New Subtitle Track dialog. After all, that is how I have been using it thus far.
- Wouldn't it be cool if the burn-in filter's keyframes repeated at the start of every new subtitle item? There is a trend in social media to have a bouncy new text for every word like so

https://www.youtube.com/shorts/afw47UHLDdQ

bmatherly:
> Most people will not care about subtitle name and language; just let me create some timed text. Instead of initializing Add Subtitle Item disabled, it can be enabled and when you click it without no subtitle, automatically make one using the same defaults as the New Subtitle Track dialog. After all, that is how I have been using it thus far.

That is a really good idea. I will work on that.

> Wouldn't it be cool if the burn-in filter's keyframes repeated at the start of every new subtitle item? There is a trend in social media to have a bouncy new text for every word like so

I think that wouldn't be too hard. All we would need is a feature to convert the subtitles to text graphics like in the video. This would be instead of using the subtitle overlay filter on the output track. The feature would add a new track and make a color producer with text filter for every subtitle item. Then, the user could use the existing copy filter feature to apply an SPR filter to all the text graphics. I'll look into this.

ddennedy:
> All we would need is a feature to convert the subtitles to text graphics like in the video.

Maybe, but I was thinking there could be a button in the filter, Repeat Keyframes, that would copy the first keyframes within the duration of the first item, and paste them at the starting time of every subsequent item. This is something you would need to run again manually every time you change the filter or subtitle. It would be important that the last keyframes is a hold type to prevent interpolation between cycles, and the code probably needs to ensure it.

bmatherly:
> I was thinking there could be a button in the filter

Thanks. I understand your suggestion better. I guess there are two possible methods:

1) Duplicate keyframes in the Subtitle filter
* Click a button and Shotcut duplicates the first keyframes and adds hold keyframes as appropriate
* This has fewer clicks than the other method
* This has less opportunity for customization
2) Convert to graphics (The Premier Pro method)
* Click a button to add a track and generate text clips for each subtitle item
* User can use the "Apply copied filters" feature to apply an effect to all subtitle items
* Additionally, the user could apply customizations to each text item. For example, maybe they want to make the text red in one. Or maybe they want to move the position of each subtitle to be by the speaker's head
* This would be more clicks than the other method

I am not sure which method I like the best. I suppose we could implement them both.

bmatherly:
I implemented "Generate Text From Subtitles". But more work is needed because we do not have keyframes for font size, and we do not have zoom option for text geometry. If you add a SPR filter to animate the zoom, it zooms from the center of the frame, not the center of the text.

ddennedy:
OK yeah, bouncy text won’t be great if you turn off fixed font size because then the text size depends in number of characters. But other things like slide or fade in and out will work.

PR #1556, opened 2024-07-19
Update shotcut_pt_BR.ts

ghost:


ddennedy:
You must do this through transifex.com where people collaborate in a friendly web app. Otherwise, the next time I pull from there it Will overwrite your changes.

Issue #1555, opened 2024-07-19
Some filters appear only after switching the Display Method


VanNostrand:
Here is my current setup:
I am using Gentoo Linux, my current shotcut version is 24.06.25. It took me a while to find that issue, I can't say how long I have it, maybe a year? I currently use Wayland 1.23.0, Plasma 6.1.3, Qt 6.7.2, mesa 24.1.4, but also had it with Plasma 5, previous Qt versions, previous shotcut versions.

Here is the issue:
One filter I use quite often is "Text: Rich". But when I start shotcut it does not appear in the list of filters (and I think there are others too, but that one I can reproduce 100%). I have to change the Display Method either from Mesa to OpenGL or the other way round (it does not matter what shotcut is currently using, it only matters if the method is switched and shotcut restarted) and then it appears!
It does not affect all filters; the filter "Text: Simple" for example is always there.

If I close shotcut, start it again, then the filter is gone and I have to repeat this procedure to get it back.

Please ask if you need more information.

bmatherly:
As a test, have you tried the Linux portable bundle from the Shotcut website? It might help to narrow down the cause if we know the problem is in that package.
https://shotcut.org/download/

VanNostrand:
I tried it and get the same behaviour. When I load my mlt file, select a track, add a filter and search for "text" there is no RichText filter. When I change DisplayMethod and shotcut restarts, the filter is there.

bmatherly:
After you start Shotcut, and the filter is not in the list, can you save the application log and upload it to this issue? From that log, we can see if the Rich Text filter is being loaded. The main reason it would not load is if it does not find the underlying QT library that it uses.

VanNostrand:
```
[Info   ] <main> Starting Shotcut version 24.06.25 
[Info   ] <main> Linux version "2.15" 
[Info   ] <main> number of logical cores = 16 
[Info   ] <main> locale = QLocale(German, Latin, Germany) 
[Info   ] <main> install dir = "/usr/bin" 
[Info   ] <ShotcutSettings::log> language "de_DE" 
[Info   ] <ShotcutSettings::log> deinterlacer "onefield" 
[Info   ] <ShotcutSettings::log> external monitor "" 
[Info   ] <ShotcutSettings::log> GPU processing false 
[Info   ] <ShotcutSettings::log> interpolation "nearest" 
[Info   ] <ShotcutSettings::log> video mode "atsc_720p_60" 
[...]

PR #1558, opened 2024-07-21
fix German translation for Rich Text

VanNostrand:
In context of fonts "fett" means "bold" and should not be used for translation here because Rich Text is a specific name for a text format. See issue #1555 for more context: it seemed that the filter was only available after a restart (which in this case switched language).


ddennedy:
Thanks for your help. You must do this through transifex.com where people collaborate in a friendly web app. Otherwise, the next time I pull from there it Will overwrite your changes.

bmatherly:
https://explore.transifex.com/ddennedy/shotcut/

PR #1560, opened 2024-08-02
Move `isMemoryLow` to `MainWindow`

mtlynch:
`MainWindow` is the only consumer of the `isMemoryLow` function, so the function does not need to be part of the `Util` class's interface. 

The `Util` class is currently kind of a hodgepodge with low cohesion among the class functions, so it may be helpful to reduce the surface area of `Util`.


mtlynch:
Meta: I discovered Shotcut recently and love the product! I'm just getting my feet wet with the code and thought this would be a small improvement, but I'm not sure if these types of PRs are welcome.

mtlynch:
Actually, on second thought, this change is debatable, so I'll come back when I have something less controversial.

bmatherly:
> Meta: I discovered Shotcut recently and love the product! I'm just getting my feet wet with the code and thought this would be a small improvement, but I'm not sure if these types of PRs are welcome.

We are happy to have some help!

If you are looking for contribution ideas, have a look at our [roadmap ](https://shotcut.org/roadmap/)and [suggestions category on the forum](https://forum.shotcut.org/c/suggestion/7)

PR #1562, opened 2024-08-04
Add a script for installing build dependencies

mtlynch:
The script is a rewrite of the code in the build-shotcut Docker image: https://github.com/mltframework/mlt-scripts/blob/091d82238a796f0a4f3afd2e5adfce6fc738be4e/docker/shotcut-build/Dockerfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,13 @@ GPLv3. See [COPYING](COPYING).
 
 The fastest way to build and try Shotcut development version is through [Qt Creator](https://www.qt.io/download#qt-creator).
 
-### From command line
+### Install build dependencies
 
-First, check dependencies are satisfied and various paths are correctly set to find different libraries and include files (Qt, MLT, frei0r and so forth).
+To install the dependencies required for compiling Shotcut from source, run the following script:
+
+```bash
+sudo ./scripts/install-linux-dependencies.sh
+```
 
 #### Configure
 
--- a/scripts/install-linux-dependencies.sh
+++ b/scripts/install-linux-dependencies.sh
@@ -1,0 +1,101 @@
+#!/bin/bash
+
+# Install dependencies for building shotcut on Linux.
+
+# Exit on first failure.
+set -e
+
+# Exit on unset variable.
+set -u
+
+# Echo commands before executing them, by default to stderr.
+set -x
+
+apt-get install -yqq \
+    apt-transport-https \
+    autoconf \
+    automake \
+    autopoint \
+    bison \
+    bzip2 \
+    ca-certificates \
+    curl \
+    flex \
+    gettext \
+    git \
+    gnupg \
+    gperf \
+    intltool \
+    intltool \
+    ladspa-sdk \
+    libarchive-dev \
+    libcurl4-openssl-dev \
+    libdouble-conversion-dev \
+    libdv-dev \
+    libegl1-mesa-dev \
+    libeigen3-dev \
+    libepoxy-dev \
+    libexif-dev \
+    libffi-dev \
+    libfftw3-dev \
+    libgtk2.0-dev \
+    libjack-dev \
+    liblist-moreutils-perl \
+    libltdl-dev \
+    libmp3lame-dev \
+    libmp3lame-dev \
+    libpotrace-dev \
+    libsamplerate-dev \
+    libsdl2-dev \
+    libsox-dev \
+    libssl-dev \
+    libtheora-dev \
+    libtool \
+    libva-dev \
+    libvorbis-dev \
+    libwebp-dev \
+    libxcb-icccm4 \
+    libxcb-image0 \
+    libxcb-keysyms1 \
+    libxcb-render-util0 \
+    libxcb-xinerama0 \
+    libxkbcommon-x11-0 \
+    libxml-parser-perl \
+    libxml2-dev \
+    libxslt1-dev \
+    make \
+    nasm \
+    ninja-build \
+    openssl \
+    p7zip \
+    patch \
+    perl \
+    pkg-config \
+    python-is-python3 \
+    python3-pip \
+    python3.9-dev \
+    ruby \
+    scons \
+    sed \
+    software-properties-common \
+    swig \
+    unzip \
+    va-driver-all \
+    wget \
+    xutils-dev \
+    xz-utils \
+    yasm \
+    zip
+
+curl https://apt.kitware.com/keys/kitware-archive-latest.asc \
+  | gpg --dearmor - >/etc/apt/trusted.gpg.d/kitware.gpg
+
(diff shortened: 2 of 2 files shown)


mtlynch:
To add more context, I'm new to the codebase, and I'm having trouble just getting the code to compile. The CI build depends on the Docker image that's defined in a separate repo, so I'm hoping to pull the code locally so it's obvious to contributors exactly what dependencies are required. 

If desired, I can create a parallel PR into mtlframework/mlt-scripts that refactors the Dockerfile to use this script, but ideally, we'd define that Docker image in this directory so that the reader doesn't have to jump between repos to understand the build requirements.

ddennedy:
This only works on certain versions of Debian-based distributions. Since this is only an aid (not limited to strict usage context), it is invitation to grow in all sorts of directions. And who will maintain it?
The dependencies are simple and expressed in CMakeLists.txt: recent MLT, Qt, and fftw.

Regarding the general reorg, I am not very keen on people telling me to change my processes that are working well for me.

mtlynch:
>This only works on certain versions of Debian-based distributions. Since this is only an aid (not limited to strict usage context), it is invitation to grow in all sorts of directions. And who will maintain it?

@ddennedy - This is proposing to move code that already exists, so I'm not seeing the change in maintenance cost. If the `build-shotcut` [docker image](https://github.com/mltframework/mlt-scripts/blob/091d82238a796f0a4f3afd2e5adfce6fc738be4e/docker/shotcut-build/Dockerfile) pulls in this script, the total code to maintain stays the same, but the script makes it easier for developers on Linux to install the dependencies.

>The dependencies are simple and expressed in CMakeLists.txt: recent MLT, Qt, and fftw.

As a data point, I don't find this simple. The documentation lists the dependencies as "Qt, MLT, frei0r and so forth," which is not clear. And these are not trivial dependencies to install, either. You can't just install these from a package manager. The reader needs to go to each of them and learn their particular install process.

>Regarding the general reorg, I am not very keen on people telling me to change my processes that are working well for me.

I like using Shotcut and wanted to contribute something back to the project. I've spent several hours just trying to understand how to install the dependencies, so I was trying to contribute back an artifact of the progress I've made in understanding it. 

I find it pretty strange and disappointing that you're taking offense to me contributing a PR to an open-source project that claims to welcome contributions. If someone submits a PR, they're not "telling you to change \[your\] process." It's just a suggestion from someone trying to work with you. If you don't want that type of contribution, that's fine, but perhaps you could document this in the README to save everyone some time. See [litestream](https://github.com/benbjohnson/litestream?tab=readme-ov-file#contribution-policy) as an example.

ddennedy:
To work on Shotcut code, you do not need the exact setup that CI uses. Maybe if you need to workout a very specific bug only found in that combination of versions of components. 
That script from the Dockerfile is specifically for Ubuntu 20.04, which is an old version specifically to make a more the portable binary bundle from which the AppImage and Snap are made. Its goal is to build on something as old as possible for glibc version compatibility, but how old also depends on the C++ version requirements of the more aggressive Qt and Glaxnimate  projects, which requires a recent enough compiler. So, it is a very specific script ("limited to strict usage context" mentioned above). As I said, the script and especially referring to it in README would invite extension to other versions and distributions thus taking it out of this "strict usage context." Then, if the Dockerfile is refactored to use it, I now have to make it work with something less clear and specific.

> "Qt, MLT, frei0r and so forth," which is not clear

That is intentionally ambiguous because it changes a lot over time and depends on the environment and needs. For example, you don't need MLT from Git unless you are also working on MLT or the Shotcut code you are working on somehow makes it necessary. You do not seem to appreciate the headache of software fragmentation, and it is not limited to Linux anymore either.  I am not going to maintain a list of the entire tree of dependencies (and versions) for many software distributions, and I know from experience neither will someone else.

The minimum versions of direct dependencies from the current `CMakeLists.txt`
- `find_package(Qt6 6.4 REQUIRED`
- `pkg_check_modules(mlt++ REQUIRED IMPORTED_TARGET mlt++-7>=7.22.0)`

> You can't just install these from a package manager. 

Here are a few (all current enough MLT): 
- https://packages.ubuntu.com/noble/libmlt++-dev
- https://packages.debian.org/trixie/libmlt++-dev
- https://packages.fedoraproject.org/pkgs/mlt/mlt-devel/
- https://packages.msys2.org/base/mingw-w64-mlt (Windows)
- https://formulae.brew.sh/formula/mlt (macOS)
- https://ports.macports.org/port/mlt/ (macOS)

And they all have current enough Qt, frei0r, and fftw.

How to be a good new team player: be useful and develop trust. Submitting minor refactors is not very useful and comes across as nit-picking. You can develop trust by demonstrating knowledge of the product by answering users' questions and improving docs. Fixing bugs that users face also demonstrates knowledge while being very useful.

bmatherly:
This conversation reminds me of the build scripts posted here:
https://mltframework.org/docs/buildscripts/
When setting up a new build environment, I used to go to that page and copy/paste the "apt-get install..." command for KDENLIVE. But those have gotten too far out of date. So now I just copy/paste the apt-get command from the Dockerfile. Then, I just run the build-shotcut.sh script and fix any lingering missing dependencies. It usually take a couple of iterations.

I had previously described by build process here:
https://forum.shotcut.org/t/what-is-the-official-way-of-building-in-2020/23396/4
This is still the method I use.

mtlynch:
@ddennedy - The way you run your project is very different than the way that I like to work, so I'm just not a match for this project. But it's your project and you're free to run it how you wish.

@bmatherly - I appreciate your responses and found them very welcoming and encouraging.

If you (either of you) are trying to attract new developers, I recommend eliminating obstacles for newcomers. It's very difficult to contribute to a project as a newcomer if the build instructions are vague or point to other projects that also have unclear build instructions (e.g., the MLT install instructions say to build from source, whereas Dan is saying that it's actually available from package managers).

I understand that you expect new contributors to fix bugs and add requested features, but that's much harder to do if even active contributors don't know the exact steps to create a dev environment that compiles the project.

bmatherly:
> If you (either of you) are trying to attract new developers, I recommend eliminating obstacles for newcomers.

I guess I don't find the barrier to entry to be too high. Have you tried running the [build script](https://github.com/mltframework/shotcut/blob/master/scripts/build-shotcut.sh) in your Ubuntu/Debian environment? The last time I did it from "scratch" on a fresh Ubuntu install was about a year ago. I think it took four hours total. Most of that time was spent with me not paying attention while the script grinds away. Then, when the script fails, I look at the error message, figure out what package I need to install, and then restart the script. After three or four iterations, I eventually see "Success!" and I am ready to edit code.

If you choose to go down this path, feel free to reach out to me personally if you get stuck. Maybe you could document your experience for other people to follow.

PR #1564, opened 2024-08-05
Update installer script (add Italian language)

bovirus:
@ddennedy 

Add Italian language


ddennedy:
This too used to be translated, but I turned it off. There is a bug in innosetup command line compiler used in our automated build such that it does not encode text properly and many languages appear with garbage characters. This required that I manually run the compiler for release, which I do not like as it is manual and error prone.

PR #1567, opened 2024-08-11
Add gps_power entry to GPS text filter

dany123:
For https://github.com/mltframework/mlt/pull/1006


ddennedy:
What about version check?

dany123:
The check is there at line 800 for the string composition. The switch case doesn't need a check because if the entry doesn't appear there won't be something to click to trigger case 21.

ddennedy:
OK, I overlooked that. It is fine after MLT merge.

PR #1591, opened 2024-10-03
Update shotcut_en.ts

abhishekcs0:


ddennedy:
That string must be changed in the C++ source code. This file is generated from the source code. I will do it for you. Thanks for the suggestion.

PR #1593, opened 2024-10-09
Transcribe subtitles

bmatherly:
Starting this draft review for collaboration and discussion. This pull request is currently working with a few TODOs:

- [x] Compile for Windows (Brian)
- [x] Compile for Mac (Dan)
- [x] Add a way to select a model?
- [x] Add a way to select the whisper.cpp main example if it is not found? Or maybe the feature should just be hidden.

### DIFF
--- a/scripts/build-shotcut.sh
+++ b/scripts/build-shotcut.sh
@@ -1105,8 +1105,14 @@ function set_globals {
   # whisper.cpp
   CONFIG[30]="cmake -B build -G Ninja -D CMAKE_INSTALL_PREFIX=$FINAL_INSTALL_DIR $CMAKE_DEBUG_FLAG -D BUILD_SHARED_LIBS=ON -D GGML_NATIVE=OFF -D WHISPER_BUILD_SERVER=OFF -D WHISPER_BUILD_TESTS=OFF"
   [ "$TARGET_OS" = "Darwin" ] && CONFIG[30]="${CONFIG[30]} -D CMAKE_OSX_ARCHITECTURES='arm64;x86_64'"
-  CFLAGS_[30]=$CFLAGS
-  LDFLAGS_[30]=$LDFLAGS
+  if test "$TARGET_OS" = "Darwin" ; then
+    CONFIG[30]="${CONFIG[30]} -D CMAKE_OSX_ARCHITECTURES='arm64;x86_64'"
+    CONFIG[30]="${CONFIG[30]} -D OpenMP_C_FLAGS=-I/opt/local/include/libomp -D OpenMP_CXX_FLAGS=-I/opt/local/include/libomp -D OpenMP_C_LIB_NAMES=libomp -D OpenMP_CXX_LIB_NAMES=libomp -D OpenMP_libomp_LIBRARY=omp"
+    LDFLAGS_[30]="$LDFLAGS -L /opt/local/lib/libomp"
+  else
+    CFLAGS_[30]=$CFLAGS
+    LDFLAGS_[30]=$LDFLAGS
+  fi
   BUILD[30]="ninja -C build -j $MAKEJ"
   INSTALL[30]="install_whispercpp"
 }
@@ -1863,10 +1869,10 @@ function deploy_mac
 
   log Copying supplementary executables
   cmd mkdir -p MacOS 2>/dev/null
-  cmd cp -a "$FINAL_INSTALL_DIR"/bin/{melt,ffmpeg,ffplay,ffprobe,glaxnimate,gopro2gpx} MacOS
+  cmd cp -a "$FINAL_INSTALL_DIR"/bin/{melt,ffmpeg,ffplay,ffprobe,glaxnimate,gopro2gpx,whisper.cpp-main} MacOS
   cmd mkdir -p Frameworks 2>/dev/null
   cmd cp -p ../../lib/libCuteLogger.dylib Frameworks
-  for exe in MacOS/Shotcut MacOS/melt MacOS/ffmpeg MacOS/ffplay MacOS/ffprobe MacOS/glaxnimate; do
+  for exe in MacOS/Shotcut MacOS/melt MacOS/ffmpeg MacOS/ffplay MacOS/ffprobe MacOS/glaxnimate MacOS/whisper.cpp-main; do
     fixlibs "$exe"
     log fixing rpath of executable "$exe"
     cmd install_name_tool -delete_rpath "$FINAL_INSTALL_DIR/lib" "$exe" 2> /dev/null
@@ -1942,6 +1948,10 @@ function deploy_mac
   cmd cp -a "$FINAL_INSTALL_DIR"/share/glaxnimate Resources
   cmd install -d lib
   cmd cp -pLR /opt/local/Library/Frameworks/Python.framework/Versions/${PYTHON_VERSION_DARWIN}/lib/python${PYTHON_VERSION_DARWIN} lib
+
+  # Whisper.cpp models
+  log Copying Whisper.cpp models
+  cmd cp -a "$FINAL_INSTALL_DIR"/share/shotcut/whisper_models Resources/shotcut
 
   log Fixing rpath in libraries
   cmd find . -name '*.dylib' -exec sh -c "install_name_tool -delete_rpath \"/opt/local/lib/libomp\" {} 2> /dev/null" \;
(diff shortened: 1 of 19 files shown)


scripts/build-shotcut.sh
```diff
@@ -1200,6 +1220,13 @@ EOF
   cmd ninja install
 }
 
+function install_whisper {
+  cmd ninja -C build install
+  cmd install -p -c build/bin/main $FINAL_INSTALL_DIR/bin/whisper_main
```
bmatherly, line 1225: This is a point of discussion. Whisper provides this "main" example program. But whisper does not provide an install target for it. I chose to name the file "whisper_main" in our package because we can't have an application called "main". But also, for distro packagers, they will never have this application installed because Whisper.cpp does not provide an install target for it. I am not sure if "whisper_main" is a good name. And I am not sure how to support distro packagers.

scripts/build-shotcut.sh
```diff
@@ -1200,6 +1220,13 @@ EOF
...
+  cmd mkdir -p $FINAL_INSTALL_DIR/share/shotcut/whisper_models
+  cmd install -p -c models/ggml-base-q5_1.bin $FINAL_INSTALL_DIR/share/shotcut/whisper_models
```
bmatherly, line 1227: Another point of discussion: I chose to install the openai model in share/shotcut. Is that the right place? I think so because whisper.cpp does not have a "share" install target. Nor does Whisper.cpp provide any recommendation for where to install the models.

scripts/build-shotcut.sh
ddennedy: I want to refer to "whisper.cpp" over "whisper" in almost everything that is official to be clear about which project we are using. When most distros preface an executable now with the project name (e.g. `mlt-melt` in Fedora) they use hyphen over underscore. So, we can call this `whispercpp-main` to anticipate what a distro packager would likely do for this.

scripts/build-shotcut.sh
ddennedy: This is OK.

ddennedy:
> Add a way to select a model?

I seem to recall we discussed we would only support initially a certain file name in the app data directory. How about appdata + `models/whisper/ggml`? I want to establish an organization where there might be motion tracking models, for example, in a different sub-folder. And there will be a similar hierarchy for appdata + `plugins/` where user-downloaded LADSPA, VST2, frei0r, and maybe eventually OpenFX plugins can be installed. For whisper, I want people to be able to install multiple models, and they can copy or link the current one to the generic target name. I am not sure `ggml` is the best name for that, but it is the format. I suppose other suitable names are `whisper`,`main`, `current`, `custom`, or `shotcut`. You pick.

> Add a way to select the whisper.cpp main example if it is not found? 

I did this for glaxnimate. If our executable is not found, open a file-open dialog for the user to choose one and save it in settings. See the call to `QFileDialog::getOpenFileName` in `glaxnimateproducerwidget.cpp`. (I had to do some command line option tests for glaxnimate that is more complex than needed for whisper.cpp-main.)

src/docks/subtitlesdock.cpp
```diff
@@ -360,6 +415,12 @@ void SubtitlesDock::setupActions()
     connect(action, &QAction::triggered, this, &SubtitlesDock::generateTextOnTimeline);
     Actions.add("subtitleGenerateTextAction", action, windowTitle());
 
+    action = new QAction(tr("Transcribe Audio..."), this);
```
ddennedy, line 418: What do you think about calling it "Speech-to-Text..." or "Create from Audio..." as that might be more obvious.  "Transcribe" seems rather high-brow.

bmatherly: I agree. And "Speech-to-Text" also maps to the label in the dialog to "Select tracks with speech".

scripts/build-shotcut.sh
bmatherly: Brilliant feedback. I agree.

bmatherly:
> I did this for glaxnimate. If our executable is not found, open a file-open dialog for the user to choose one and save it in settings.

I see you implemented that. Thanks. We will also need that for the model. If the package was not provided by us, and does not have whisper.cpp-main where we expect it, then surely the model will not be where we expect it, either.

I will work on implementing your suggestions in the next day or two.

bmatherly:
Just to clarify this comment:

> How about appdata + models/whisper/ggml?

I don't think we can put it in the local app data directory because it needs to come with the installer (or zip file). So it would go in the application installation directory instead.

ddennedy:
> I don't think we can put it in the local app data 

This is only for a user supplied override rather than have a way to choose a model in the UI. Not just for when not using one of our builds. In that latter case, I think you could open a file dialog to choose a model faster choosing the executable. But that admittedly can be a little confusing. Do you want to add a way to choose a custom model in the UI in general? If so, it would cover both situations, but then it probably needs a way to see what it is and reset it. Did you see my commit that implemented app data dir? It too can be the universal method and less work?

ddennedy:
Oh, and app data is not the place we install the model in our builds. That still works when not overridden.

bmatherly:
> Do you want to add a way to choose a custom model in the UI in general?

I think we need to. Otherwise, for distro packages, users would have to know the "secret" that they have to drop a model with a specific name in a specific location in the app data directory. We can write documentation, but nobody will read it and we will still receive support requests.

I propose to add a new expandable drawer to the existing dialog box where the user can choose the path for whisper.cpp-main and the model. For our package builds, the drawer will be closed by default and can only be opened by clicking an "advanced" button. For distro packages, the drawer will open by default because Shotcut did not find a model and exe - prompting the user to set them both. The OK  button will not work until both are set.

ddennedy:
Ok you work on that and I’ll take care of the other builds including windows.

scripts/build-shotcut.sh
bmatherly: Done

src/docks/subtitlesdock.cpp
bmatherly: Done

bmatherly:
> Ok you work on that and I’ll take care of the other builds including windows.

Done

Some other optimizations that we can add to this PR or wait until after:
* I have not enabled support for any GPUs in the build yet. There are switches that can be turned on, but i do not know what happens if they are not present.
* The default number of threads is 4. I think we should change that to use the number of threads in the system.

ddennedy:
> I have not enabled support for any GPUs in the build yet.

it comes for free on macOS, but the CUDA SDK license is incompatible for us.

ddennedy:
Re CUDA and GPL I found that despite this some projects do include it in builds such as Blender and this
https://github.com/LeelaChessZero/lc0/issues/184

Also, Ubuntu includes CUDA packages but not msys2 we use for Windows. The Windows CUDA SDK is some downloaded executable that I think can be automated. There is a good risk that the CUDA libs in the 20.04 version of Ubuntu we use for our builds is too old. That should stay on 20.04 as long as possible to maintain compatibility. So, it may be possible but too much trouble for now. If it is possible, there are other things we can include with CUDA such as some FFmpeg filters that can enable proxy transcodes entirely on the GPU.

ddennedy:
FYI, on my M1 MacMini with whisper.cpp's usage of GPU Metal, all 14:48 minutes of Sintel only took 26 seconds. I see whisper.cpp can be built with OpenVINO, which supports OpenCL and will give broader coverage. Something to look into. OpenVINO could also provide us with audio noise reduction, which is on the road map.

bmatherly:
> FYI, on my M1 MacMini with whisper.cpp's usage of GPU Metal, all 14:48 minutes of Sintel only took 26 seconds. I see whisper.cpp can be built with OpenVINO, which supports OpenCL and will give broader coverage. Something to look into. OpenVINO could also provide us with audio noise reduction, which is on the road map.

Sounds promising! Did you happen to do any other benchmarking on other systems? I am interested to know what most users can expect because I mostly focused my benchmarking on my "potato" machine.

ddennedy:
Basically, now that I implemented the thread count arg, on most modern machines with at least 8 logical processors, it is quite fast too. I would say it will feel normal as opposed to oddly slow to most people--based on the speed of other Shotcut jobs including export. No benchmarks tho.

ddennedy:
OpenVINO also has background subtraction
https://docs.openvino.ai/2024/omz_demos_background_subtraction_demo_python.html

PR #1601, opened 2024-11-03
refactor: replace non-empty QString constructors with QStringLiteral()

Integral-Tech:
- Qt provides a macro `QStringLiteral()`, which makes constructing QString objects from string literals more efficient.


ddennedy (changes requested):
"13 files are formatted incorrectly" per our `astyle` 3.1 check rules in `.astylerc`

Integral-Tech:
@ddennedy `astyle` check fixed :)

src/commands/timelinecommands.cpp
```diff
@@ -611,7 +611,7 @@ void MoveClipCommand::redo()
                     // pulling clips on same track
                     (playlist.is_blank_at(newStart) && targetIndex == clipIndex - 1)) {
                 bool done = true;
-                if (targetIndex >= clipIndex && m_model.isTransition(playlist, clipIndex - 1)) {
+                if (targetIndex == clipIndex && m_model.isTransition(playlist, clipIndex - 1)) {
```
ddennedy, line 614: This is wrong.

ddennedy (changes requested):
There is no longer any `QStringLiteral` in this diff! It looks like that got whacked when reformatting.

Integral-Tech:
> There is no longer any `QStringLiteral` in this diff! It looks like that was damaged when reformatting.

Fixed :)

ddennedy (approved):

PR #1602, opened 2024-11-04
Add support for XML clips in chains.

bmatherly:
This allows speed effects to XML clips.
It also adds a properties panel for XML clips.

Also fixes this issue:
https://forum.shotcut.org/t/issues-when-using-mlt-as-a-clip-in-a-timeline/44164

If the XML clip has a filter on the output, the filter will not appear in the project - it is "hidden" inside the chain.

Depends on
https://github.com/mltframework/mlt/pull/1047

This is what the properties panel looks like (similar to the Output properties):
image

The way this is designed, the embedded XML will always be rendered at the XML clip profile (by the consumer producer). Then it will be converted to the project profile if necessary.

Discussion: Should we show a warning in the properties panel if the XML clip profile does not match the project profile?

Discussion: Do we want to add a button or method to lauch Shotcut from the properties panel so the clip can be edited in another instance of Shotcut? Maybe that would be confusing to have two instances at the same time.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -164,6 +164,7 @@ add_executable(shotcut WIN32 MACOSX_BUNDLE
   widgets/lissajouswidget.ui
   widgets/lumamixtransition.cpp widgets/lumamixtransition.h
   widgets/lumamixtransition.ui
+  widgets/mltclipproducerwidget.cpp widgets/mltclipproducerwidget.h
   widgets/networkproducerwidget.cpp widgets/networkproducerwidget.h
   widgets/networkproducerwidget.ui
   widgets/newprojectfolder.cpp widgets/newprojectfolder.h
--- a/src/controllers/filtercontroller.cpp
+++ b/src/controllers/filtercontroller.cpp
@@ -215,7 +215,8 @@ void FilterController::setProducer(Mlt::Producer *producer)
         m_metadataModel.updateFilterMask(!MLT.isTrackProducer(*producer),
                                          producer->type() == mlt_service_chain_type,
                                          producer->type() == mlt_service_playlist_type,
-                                         producer->type() == mlt_service_tractor_type);
+                                         producer->type() == mlt_service_tractor_type,
+                                         producer->get("mlt_service") != QString("xml-clip"));
     } else {
         setCurrentFilter(QmlFilter::DeselectCurrentFilter);
     }
--- a/src/docks/playlistdock.cpp
+++ b/src/docks/playlistdock.cpp
@@ -992,22 +992,12 @@ void PlaylistDock::addFiles(int row, const QList<QUrl> &urls)
         }
         Mlt::Producer p;
         if (path.endsWith(".mlt") || path.endsWith(".xml")) {
-            if (Settings.playerGPU() && MLT.profile().is_explicit()) {
-                Mlt::Profile testProfile;
-                Mlt::Producer producer(testProfile, path.toUtf8().constData());
-                if (testProfile.width() != MLT.profile().width()
-                        || testProfile.height() != MLT.profile().height()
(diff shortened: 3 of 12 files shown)


bmatherly:
Of interest:
https://forum.shotcut.org/t/adding-to-timeline-a-mlt-as-a-clip-with-different-fps-crashes-shotcut/46418

An idea from that post is that I could add the duration to the properties panel.

ddennedy:
> An idea from that post is that I could add the duration to the properties panel.

Some things I noticed while debugging the forum report and testing these branches:
- The audio plays with heavy crackling artifacts if you nest a project with audio and a different frame rate (outer project 60 fps, inner project 29.97).
- dragging .mlt file to playlist, double-click to open from playlist, Properties does not show the new producer widget
- same goes when drag .mlt to Timeline
- File > Open MLT XML as Clip (before adding to Timeline) is giving me empty Properties (I tried a couple projects)

image

I am using the MLT branch as well.

ddennedy:
FYI, the audio issue on mismatching frame rate exists regardless this change, but if this change can accommodate speed I wondered if it could improve this. It is worse when the nested project fps is less than the outer's.

bmatherly:
> FYI, the audio issue on mismatching frame rate exists regardless this change, but if this change can accommodate speed I wondered if it could improve this. It is worse when the nested project fps is less than the outer's.

I will look into this. My intention is to fix these kinds of problems.

> It permits the addition of Speed: Forward & Reverse and Time Remap, which can go in reverse, and the project might perform very poorly because it masks whether its videos seek well (has no B frames).

This is a very good point. Part of my motivation was to allow speed effects on a nested clip. But now I think that is not a good idea and probably not that important since a user can simply encode the embedded project and apply speed to the encoded file. I think I will remove the chain wrapper (but keep the producer_consumer wrapper).

Thanks again for your testing. I will spend some more time on this. I would like it to be bullet proof before we show it to a wider audience.

bmatherly:
> Discussion: Should we show a warning in the properties panel if the XML clip profile does not match the project profile?

This is done

> The audio plays with heavy crackling artifacts if you nest a project with audio and a different frame rate (outer project 60 fps, inner project 29.97).

Fixed
    
> dragging .mlt file to playlist, double-click to open from playlist, Properties does not show the new producer widget

Fixed

> same goes when drag .mlt to Timeline

Fixed

> File > Open MLT XML as Clip (before adding to Timeline) is giving me empty Properties (I tried a couple projects)

Fixed

> It permits the addition of Speed: Forward & Reverse and Time Remap, which can go in reverse, and the project might perform very poorly because it masks whether its videos seek well (has no B frames).

I did not prohibit speed effects (but I can easily if we want to). There are some use cases that I would like to support. For example, adding a timer on a clip and then adding that as a subproject to show a timer moving in fast motion with a fast forward video. I am open minded about this.

The major difference with this latest version is that the xml-clip producer is always opened in its native profile (frame rate) and a timeremap link is added to the chain if the frame rate does not match the project frame rate. The timeremap link performs frame rate conversion.

The only thing I am slightly disatisfied with is that I notice interlaced files have lower performance due to two deinterlace links being stacked (one inside the xml producer and one in the encapsulating chain). Deinterlace links cause extra processing even when the project is progressive because they pull previous/next frames in case a filter requests deinterlace. I haven't thought of a good solution for this, and I am not sure of the priority.

ddennedy:
> I did not prohibit speed effects (but I can easily if we want to). 

What about only allowing Speed: Forward Only? Basically, check `seekReverse` against the producer's `mlt_service`. It would be ideal if we could filter the filter menu based on this, but I doubt we can do that yet.

bmatherly:
> It would be ideal if we could filter the filter menu based on this, but I doubt we can do that yet.

I did it.

PR #1604, opened 2024-11-06
refactor: remove QString::number() in arg() arguments

Integral-Tech:
- `QString::arg()` has overload for numeric data types, hence it is unnecessary to use `QString::number()` in `QString::arg()` arguments.

### DIFF
--- a/src/widgets/scopes/audiowaveformscopewidget.cpp
+++ b/src/widgets/scopes/audiowaveformscopewidget.cpp
@@ -259,15 +259,15 @@ void AudioWaveformScopeWidget::mouseMoveEvent(QMouseEvent *event)
 
     qreal position = (qreal)event->pos().x() / (qreal)width();
     int sample = (qreal)samples * position;
-    QString text = tr("Sample: %1\n").arg(QString::number(sample + 1));
+    QString text = tr("Sample: %1\n").arg(sample + 1);
 
     for (int c = 0; c < channels; c++) {
         const int16_t *q = audio + (channels * sample) + c;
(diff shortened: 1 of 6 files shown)


src/widgets/scopes/audiowaveformscopewidget.cpp
```diff
@@ -259,15 +259,15 @@ void AudioWaveformScopeWidget::mouseMoveEvent(QMouseEvent *event)
...
 
     for (int c = 0; c < channels; c++) {
         const int16_t *q = audio + (channels * sample) + c;
         qreal scaledValue = (qreal) * q / MAX_AMPLITUDE;
         qreal dbValue = 20 * log(fabs(scaledValue));
         if (dbValue < 0.01 && dbValue > -0.01) dbValue = 0.0;
-        text += tr("Ch: %1: %2 (%3 dBFS)").arg(QString::number(c + 1), QString::number(scaledValue, 'f', 2),
-                                               QString::number(dbValue, 'f', 2));
+        text += tr("Ch: %1: %2 (%3 dBFS)").arg(c + 1).arg(scaledValue, 0, 'f', 2)
+                .arg(dbValue, 0, 'f', 1);
```
ddennedy, line 270: This has the incorrect precision of 1.

ddennedy (changes requested):
Looks good except that one mistake.

src/widgets/scopes/audiowaveformscopewidget.cpp
Integral-Tech: Oh, sorry :(

Integral-Tech:
@ddennedy Fixed :)

PR #1611, opened 2024-12-03
Implement Gradient Map filter

YakoYakoYokuYoku:
This simply implements support for a gradient map filter, which was added to MLT in mltframework/mlt#1046. This has a UI similar to the Color Ramp node from Blender, although more basic as of now and more features could be added if desired. Its icon is a demo of mapping an image of the [Old Town Hall of Esslingen am Neckar](https://free-images.com/display/town_hall_town_hall.html) to a blue to green to red gradient. So the equivalent of setting up the `gradientmap` filter with `melt`, e.g. ...

```
stop.1='#ff0000ff 0.0' stop.2='#ff00ff00 0.5' stop.3='#ffff0000 1.0'
```

... would be to have three sliders, blue to the left, green in the middle, red to the right.


src/qml/filters/gradientmap/ui_gradientmap.qml
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    width: 350
+    height: 180
+    Component.onCompleted: {
+        if (filter.isNew) {
+            filter.set('stop.1', '#ff000000 0.0');
+            filter.set('stop.2', '#ffffffff 1.0');
+            filter.savePreset(defaultParameters);
+        }
```
ddennedy, line 39: This section is to define stock presets, but you have not included the Preset control. You should include it as well as a couple of them such as the ones from your MLT pull request.

src/qml/filters/gradientmap/ui_gradientmap.qml
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2017-2022 Meltytech, LLC
```
ddennedy, line 2: Change the copyright year to 2024.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
```diff
@@ -0,0 +1,307 @@
+/*
+ * Copyright (c) 2019-2022 Meltytech, LLC
```
ddennedy, line 2: Change the copyright year to 2024.

src/qml/filters/gradientmap/ui_gradientmap.qml
```diff
@@ -0,0 +1,76 @@
...
+        if (filter.isNew) {
+            filter.set('stop.1', '#ff000000 0.0');
+            filter.set('stop.2', '#ffffffff 1.0');
+            filter.savePreset(defaultParameters);
+        }
+        setControls();
+    }
+
+    ColumnLayout {
+        // columns: 5
```
ddennedy, line 44: The layout seems weird. The "Gradient Stops" label usually appears left of its controls, but I think here you are trying to say everything below here are the gradient stops like a heading. Actually, for this UI since there are so few parameters maybe that label is not needed. It should probably be switched to a `GridLayout` with the addition of Presets.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
```diff
@@ -0,0 +1,307 @@
...
+            border.width: 1
+            radius: 4
+            gradient: Gradient {
+                id: gradientView
+                orientation: Gradient.Horizontal
+            }
+        }
+    }
+
+    Shotcut.DoubleSpinBox {
```
ddennedy, line 262: This should have a label such as "Selected Stop". Also, it is cleaner if the UI is 1-based instead of 0-based.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
```diff
@@ -0,0 +1,307 @@
...
+        onValueChanged: {
+            for (var idx = 0; idx < gradientMap.stops.length; idx++) {
+                var stop = gradientMap.stops[idx];
+            }
+            var stop = gradientMap.stops[value];
+            stopPosition.value = stop.position;
+        }
+    }
+
+    Shotcut.DoubleSpinBox {
```
ddennedy, line 279: This should have a label such as "Position" and the field should have a suffix " %" and converted from the MLT property value to the control value with `*  100` and converted from the control value to MLT property value with `/ 100`. We do not like bare numeric parameters in our UI. They should have a label and strive to have some unit of measure (% here).

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
```diff
@@ -0,0 +1,307 @@
...
+        value: typeof stops[stopSpinner.value] != 'undefined' ? stops[stopSpinner.value].position : 0
+        decimals: 3
+        from: 0
+        to: 1
+        stepSize: 0.01
+        onValueModified: _setStopColor(stopSpinner.value, stops[stopSpinner.value].color, value)
+    }
+
+    Rectangle {
+        id: stopColor
```
ddennedy, line 292: This should have a label such as "Color".
 Finally, this can be in one row with the above 2 controls like
`Selected Stop   __1__   Position   __50 %_   Color [ ]`
Does that make sense?

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
```diff
@@ -0,0 +1,307 @@
...
+            }
+            if (typeof next === 'undefined') {
+                _insertStop(index, {"color": prev.color, "position": 1.0});
+                return;
+            }
+            _insertStop(index, {"color": next.color, "position": (prev.position + next.position) / 2.0});
+        }
+    }
+
+    Button {
```
ddennedy, line 152: The + and - buttons are too wide. Try adding to them `implicitWidth: height`

Alternatively, you could make this more consistent with our other gradient control by removing these buttons and placing a spinner with the number of stops after the control on the same row. See filter "Audio Spectrum Visualization"
image
(Except Gradient Map filter could use "stops" instead of "colors" as the suffix.)

ddennedy (changes requested):

src/qml/filters/gradientmap/ui_gradientmap.qml
YakoYakoYokuYoku: Yes, users would also like to save quite a few of them.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
YakoYakoYokuYoku: I'm kinda against the label to be honest, probably a hover tip would be better.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
YakoYakoYokuYoku: So we are going like Krita, as they do percentages too, no problem with that whatsoever.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
YakoYakoYokuYoku: I'd rather put the color into its row because I find it more intuitive for it to be like that. I could take a look at the other two, though.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
YakoYakoYokuYoku: The add and remove buttons at least gives me the chance to select where to insert a stop and which one to delete, controlling a number of stops only makes sense for me if they are equidistant.

src/qml/modules/Shotcut/Controls/qmldir
```diff
@@ -4,6 +4,7 @@ ColorPicker 1.0 ColorPicker.qml
 DoubleSpinBox 1.0 DoubleSpinBox.qml
 Gauge 1.0 Gauge.qml
 GradientControl 1.0 GradientControl.qml
+GradientMapControl 1.0 GradientMapControl.qml
```
ddennedy, line 7: `GradientMapControl` should not be separate from the `ui.qml` because it will never be reused. These are only for reusable components.

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
ddennedy: Bare controls with no label and numeric fields suffix have no meaning! It is like a puzzle game. Do you have a way to view the UI you created? Basically, you can extract the latest Shotcut portable build, replace the `libmltplus.so` that contains your new `gradientmap` filter and edit these QML files in `Shotcut.app/share/shotcut/qml/...`
Here is what it looks like now that is completely unacceptable for my standards with Shotcut
image
The numeric fields are not layed out correctly or in any 
Why does the selected stop's color button need to be the full width of the dock?

Here, I fixed it for you.
image

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
ddennedy: OK

src/qml/modules/Shotcut/Controls/GradientMapControl.qml
YakoYakoYokuYoku: Ah, I've took out the width value of both spinboxes and they've done that weird squish overlap, that was an oversight from my part.

> Why does the selected stop's color button need to be the full width of the dock?

I thought that it was more intuitive for users that way, I'm more familiarized with Blender in that respect. As per what it stands now from your retouches it looks good.

PR #1617, opened 2024-12-30
Add curve option for audio fade in/out filters

bmatherly:
As suggested here:
https://forum.shotcut.org/t/add-perceptually-linear-option-to-audio-fade-in-out/43680

In that suggestion, the poster requests a "perceptually linear" curve. The current keyframe type is linear, but it is operating on dB. So I would argue that it is already perceptually linear. Nonetheless, it is not that hard to expose the existing keyframe interpolation types. So this PR provides a drop-down box in the filter panels for fade in/out to allow the user to select the curve type. 

image

image

Here is the original (linear) fade waveform:
image

Here is an example using Ease Out Exponential:
image

Here is an example using Ease In/Out Circular:
image

Note 1: it is not possible to provide a linear VALUE option since that would require the inverse of "20 log" which we do not offer.

Note 2: I did not try to render the fade line in the timeline to represent the curve type

Note 3: I did not provide options for the "bouncy" interpolation types

Discussion topic 1: would it be worthwhile to also offer this for the video fade filters?

Discussion topic 2: is "Curve" the best label for this? We don't really expose that it is using keyframes under the hood. So I did not want to use "Keyframe Type". Other ideas?

Discussion topic 3: Should I try to add some kind of graphical representation of the interpolation type (like we do in the keyframe type menu)? I could change the combobox do a menu with icons.

### DIFF
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -1497,6 +1497,7 @@ void MultitrackModel::fadeIn(int trackIndex, int clipIndex, int duration)
                 filter.reset(getFilter("fadeInVolume", info->producer));
 
                 if (duration > 0) {
+                    mlt_keyframe_type type = (mlt_keyframe_type)Settings.audioInCurve();
                     // Add audio filter if needed.
                     if (!filter) {
                         Mlt::Filter f(MLT.profile(), "volume");
@@ -1504,12 +1505,19 @@ void MultitrackModel::fadeIn(int trackIndex, int clipIndex, int duration)
                         info->producer->attach(f);
                         filter.reset(new Mlt::Filter(f));
                         filter->set_in_and_out(info->frame_in, info->frame_out);
+                    } else {
+                        Mlt::Animation animation(filter->get_animation("level"));
+                        if (animation.is_valid()) {
+                            type = animation.key_get_type(0);
+                        }
                     }
 
                     // Adjust audio filter.
                     filter->clear("level");
                     filter->anim_set("level", -60, 0);
                     filter->anim_set("level", 0, duration - 1);
+                    Mlt::Animation animation(filter->get_animation("level"));
+                    animation.key_set_type(0, type);
                     filter->set(kShotcutAnimInProperty, duration);
                     isChanged = true;
                 } else if (filter) {
@@ -1614,6 +1622,7 @@ void MultitrackModel::fadeOut(int trackIndex, int clipIndex, int duration)
                 filter.reset(getFilter("fadeOutVolume", info->producer));
 
                 if (duration > 0) {
+                    mlt_keyframe_type type = (mlt_keyframe_type)Settings.audioOutCurve();
                     // Add audio filter if needed.
                     if (!filter) {
                         Mlt::Filter f(MLT.profile(), "volume");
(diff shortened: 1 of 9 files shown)


ddennedy:
Thank you for attempting this. I think the request is to use the same function that is used for the scales on the peak meter and volume slider in Shotcut. Search the code for "IEC_Scale" This IEC refers to standard 61672. I thought this was supposed to be like log base 10, and I do not know why that code does not use it directly. In any case, I think we need that curve here more so than all of these options. Or, at least, it should be default or second recommendation.

bmatherly:
The IEC scale function is here

Shotcut:
https://github.com/mltframework/shotcut/blob/master/src/widgets/iecscale.h

MLT:
https://github.com/mltframework/mlt/blob/2b9f25072130002790ba8eacf30391718c94e9da/src/modules/normalize/filter_audiolevel.c#L34

But I do not think that is it.  That function comes from an IEC standard for visualizing audio levels in a meter.

In my (limited) experience, most audio tools default to the log function and operate the fade in dB because that most closely matches the human hearing system:
https://en.wikipedia.org/wiki/Fade_(audio_engineering)#Logarithmic

That is also what we currently do in our audio fades. You can see here, the volume filter accepts the user value in DB and performs inverse log to convert back to amplitude:
https://github.com/mltframework/mlt/blob/2b9f25072130002790ba8eacf30391718c94e9da/src/modules/normalize/filter_volume.c#L40

I think the user making that suggestion incorrectly assumed that we perform a linear fade on the amplitude. Maybe I should just go into that post and offer that information for clarification.

But I also notice that many audio tools offer custom fade "curves" that allow the user to customize the fade. A popular fade curve is the "S curve" which is kind of like an ease in/out.

ddennedy:
> operate the fade in dB because that most closely matches the human hearing system

Yes, but not linearly in dB. Look at any mixer slider, and it is in dB scale with non-linear increments. It seems be log10 or IEC 61672. So, from user experience to signal manipulation it seems to be log (user control/meter) over log (db).

bmatherly:
> Yes, but not linearly in dB. Look at any mixer slider, and it is in dB scale with non-linear increments. 

Those are some helpful observations.

> from user experience to signal manipulation it seems to be log (user control/meter) over log (db).

As an amateur audio engineer, I never presumed that the db scale on the slider or meter was intended to represent a linear sounding fade scale. I always thought that scale existed to give me more resolution in the areas of the sound envelope that I need the most resolution. I am much more likely to make minute gain adjustments between 0 and -10db than I am between -40db and -50db. So I don't want the slider/meter space to  be wasted on those ranges. Also, when I fade a signal out by moving the slider from 0 to -80db, I am certain that I never move my finger at a linear velocity. 

Your questions have prompted me to do more reading about other tools and the options they offer. One thing I observe is that each tool seems to have their own "secret sauce" when it comes to fading. Some offer many options while others offer few options. The one thing that I see consistently is that everyone seems to agree that a logarithmic fade is considered "natural". But even then, there is secret-sauce to decide how to get the signal to zero since a log would go all the way to negative infinity.

I found this article pretty helpful and it seems to represent a good summary of many other articles and how-tos that I have read:
https://www.sageaudio.com/articles/how-to-create-fades-during-mastering

Here are the key takeaways for me:
* Fading in/out is an artistic decision and different fades may be preferred by the artist for different situations
* Therefore, people want to chose from a variety of fade curves
* Logarithmic (the curve we currently use) seems to be accepted as the most "natural" sounding fade.

Due to the variety and inconsistency in the industry, I suggest that we adapt this PR to NOT use the keyframe interpolation naming. Instead, we can choose more abstract names that may be familiar to other tools. 

Here is a suggestion:
**"Natural" - Our current method (log) - uses "linear" keyframe**
image

**"S-Curve" - slow start, fast middle, slow end - uses "in/out quartic" keyframe**
image

**"Fast/Slow" - starts fast and ends slow - uses "out circular" keyframe**
image

**"Slow/Fast" - starts slow and ends fast - uses "in quadratic" keyframe**
image

I do not think we should offer a "linear" option because of the confusion that could come from that.

ddennedy:
Good analysis. I agree to your suggestions. I wonder if people will expect the visual fades in the timeline to reflect the “curve.” I am ok with that name as well.

bmatherly:
> I wonder if people will expect the visual fades in the timeline to reflect the “curve.”

They probably will. But we will remind them that the fade indicator in the timeline represents both the audio and video fades. I do not think it will be required to render the curve shape in the timeline. But I am willing to do that if it is really important.

> I am ok with that name as well.

I notice that both the Sage article (link above) and the Audacity documentation refers to them as "Fade Types". So I was thinking about changing the label to "Type". I don't feel strongly about it.

PR #1618, opened 2025-01-06
Fix typo in ui.qml

mrleemon:


bmatherly:
Thanks!

PR #1620, opened 2025-01-11
Add "Copy Current Filter" action

bmatherly:
As suggested here:
https://forum.shotcut.org/t/copy-single-filters/13909

This PR adds a new button next to the existing "Copy Filters" Button. This new button is called "Copy Selected Filter".
image

In the forum discussion, the conversation converged on the idea of adding multi-select to the filter list so that the user can choose what filters to copy. I am in support of this. However, I only implemented single filter copy in this pull request. The reasons that I did not implement multi-select are:
1) The QML ListView Item does not offer multi-select. So I added a lot of extra logic to disable the built-in selection and to implement my own multi-select implementation on top of it. It was mostly working.
2) After I implemented multiselect for the list view, I found that there are a lot of dynamic interactions between the QML filter list, filter controller, and filter dock. I kept adding more and more code to keep all three classes happy with the multi-select.

Eventually, I decided that the extra complexity and risk of regression is too great and I changed to this simpler single selection implementation.


ddennedy:
I am not a fan of the icon and adding a confusing alternative copy action next to the other on the tool bar. Perhaps the existing copy should have a drop down arrow to choose from alternatives:
- Enabled (checked)
- All
- Selected

You must click the drop down arrow to access the alternatives (click icon chooses Enabled) unless you use the actions window or assign a shortcut. What do you think of that proposal?

bmatherly:
I agree with your comments. Unfortunately, the QML button Item does not have all the same features as a QButton and does not offer the dual function of an action and a submenu.

I submitted a variation that always shows the submenu with the three options. 
image

I think that maybe it looks a little strange that they all have the same icon. Maybe I should just remove the icon.

ddennedy:
Please remove the icons from the submenu. Do we need the “filters suffix?” It should appear  in the “Filters >” group in the Actions window.  The singular vs. plural might raise issues from translators, and the plural form of the qsTr method needs to supplied with a count parameter. Otherwise, I’m fine that it always opens the menu.

bmatherly:
Both good comments. This is done. You were right - the Shortcuts menu still makes sense because it says "Filters -> Copy All", etc.

PR #1622, opened 2025-01-16
Patch: introduce proper plural form strings and fix typo

rschiang:
This patch consolidates some source string issues found during the translation process.

- fisheye: `ortographic` → `orthographic`, as in [orthographic projection](https://en.wikipedia.org/wiki/Orthographic_projection)
- fisheye: `seperate` → `separate`
- subtitles: fix `QObject::tr` calls with proper plural form support (with the `n` parameter)

Hope this helps! 🚀


ddennedy:
Thank you, but this must wait until after the next release because we are late in beta when we do not change strings that can break existing translations (not enough time for translators to update).

rschiang:
Thank you for the swift reply! Waiting until the next release cycle seems fair.

Please merge them whenever you feel right!

PR #1623, opened 2025-01-18
Fix convert time when deinterlace or frame rate conversion

bmatherly:
As reported here:
https://forum.shotcut.org/t/convert-to-edit-friendly-with-deinterlace-to-double-frame-rate-progress-eta-in-jobs-panel-is-wrong/45551

Uses the FFMpeg  time status instead of frames.

### DIFF
--- a/src/jobs/ffmpegjob.cpp
+++ b/src/jobs/ffmpegjob.cpp
@@ -28,11 +28,12 @@
 #include <QRegularExpression>
 #include <Logger.h>
 
+#include <cstdio>
+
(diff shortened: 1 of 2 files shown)


ddennedy:
I do not want to review every change you have. Merge after the release.

bmatherly:
That is fine. Normally I would not make a pull request for this. But I am staging things here until after the release so I do not forget about it. Now I realize that you are getting a notification even if I do not request you in the review.

ddennedy:
The beta period accommodates low risk fixes, even for old bugs, that don’t change strings. This change qualifies because even though it could go wrong under a certain situation, it is simply progress feedback. Feel free to merge it.

PR #1624, opened 2025-01-20
Show project profile in title bar

bmatherly:
As suggested here:
https://forum.shotcut.org/t/settings-video-mode/12790/2

Looking for discussion on this idea.

For me personally, I have had trouble sometimes when it is not perfectly clear what profile is being used. For timeline projects, I  can look at the properties for the "keystone" clip in the timeline. But for playlist and clip projects, I am not sure the  best way to see it. If I do not use Automatic mode, I can look in the settings. But in automatic mode, the best I have found is to go to the Advanced section of the export panel and see what export settings are default.

I would like the profile to be more explicitly visible. At first, I was not excited about the idea of putting it in the title bar. But after trying it, I think it might be OK. I am interested in suggestions for other places it could go. What I like about the title bar is that it is always visible, and every obvious when it changes.

image

I am also interested in discussion about what to show. In this proposed example, I chose to show the resolution, frame rate, and audio channels. But should we add aspect ratio or color space?

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -252,6 +252,9 @@ void MainWindow::connectUISignals()
     connect(this, &MainWindow::producerOpened, this, &MainWindow::onProducerOpened);
     connect(ui->mainToolBar, SIGNAL(visibilityChanged(bool)), SLOT(onToolbarVisibilityChanged(bool)));
     ui->actionSave->setEnabled(false);
+    connect(this, &MainWindow::audioChannelsChanged, this, &MainWindow::updateWindowTitle);
+    connect(this, &MainWindow::producerOpened, this, &MainWindow::updateWindowTitle);
+    connect(this, &MainWindow::profileChanged, this, &MainWindow::updateWindowTitle);
 }
 
 void MainWindow::setupAndConnectUndoStack()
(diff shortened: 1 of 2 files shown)


ddennedy:
A different request asked for the full path of the project in the title bar. Try to include that and see how usable it is.

> for playlist and clip projects, I am not sure the best way to see it. 

I am little surprised you did not know to use Settings > Video Mode > Custom > Add... I can understand that is not very obvious to most users, but I suspect most people use Export.

bmatherly:
> A different request asked for the full path of the project in the title bar. Try to include that and see how usable it is.

Here is what it looks like with a simple path:
image

Here is what it looks like with a path that is too long for my HD monitor. Maybe there should be a limit of "x" characters...
image

> I am little surprised you did not know to use Settings > Video Mode > Custom > Add

That requires navigating three levels of menu, closing a warning dialog, and then finally closing the Add dialog. Both the add dialog and export panes are not status of the profile, but only a way to infer what the profile probably is.

bmatherly:
Here is an example that limits the length of the project path+name to be 1/4 of the width of the window - leaving plenty of room for the profile and the application name. I think I like this. I will update the PR with this change.
image

PR #1625, opened 2025-01-21
Add vertical offset parameter to the No Sync filter

bmatherly:
As requested here:
https://forum.shotcut.org/t/filter-no-sync-horizontal-offset/45942

Depends on this PR:
https://github.com/dyne/frei0r/pull/203

Note: I swapped horizontal and vertical nomenclature in the UI. The vertical offset moves the picture up and down. The horizontal offset moves the picture left and right. 

@ddennedy, code review not really required. But I would like to make sure we are both on the same page for horiztonal/vertical labels.

image


src/qml/filters/nosync/meta.qml
```diff
@@ -13,14 +13,21 @@ Metadata {
     keyframes {
         allowAnimateIn: true
         allowAnimateOut: true
-        simpleProperties: ['0']
+        simpleProperties: ['0', '1']
         parameters: [
             Parameter {
-                name: qsTr('Offset')
+                name: qsTr('V Offset')
```
ddennedy, line 19: Change this to "Vertical" (no offset).

src/qml/filters/nosync/meta.qml
```diff
@@ -13,14 +13,21 @@ Metadata {
...
                 property: '0'
                 isCurve: true
                 minimum: 0
                 maximum: 1
+            },
+            Parameter {
+                name: qsTr('H Offset')
```
ddennedy, line 26: Change this to "Horizontal"

src/qml/filters/nosync/ui.qml
```diff
@@ -80,7 +88,36 @@ Shotcut.KeyframableFilter {
         }
 
         Label {
-            text: qsTr('Offset')
+            text: qsTr('Vertical Offset')
```
ddennedy, line 91: Change this to "Vertical"

src/qml/filters/nosync/ui.qml
```diff
@@ -80,7 +88,36 @@ Shotcut.KeyframableFilter {
...
+            id: verticalKeyframesButton
+
+            onToggled: {
+                enableControls(true);
+                toggleKeyframes(checked, vertical, verticalSlider.value / verticalSlider.maximumValue);
+            }
+        }
+
+        Label {
+            text: qsTr('Horizontal Offset')
```
ddennedy, line 120: Change this to "Horizontal"

ddennedy (changes requested):

src/qml/filters/nosync/meta.qml
bmatherly: done

src/qml/filters/nosync/meta.qml
bmatherly: done

src/qml/filters/nosync/ui.qml
bmatherly: done

src/qml/filters/nosync/ui.qml
bmatherly: done

bmatherly:
This builds on Linux for me. I have not tried to build it in Windows or Mac.

ddennedy:
If we have a build problem with Frei0r then alI of this PR will be reverted unless you expect me to l tease it apart because I typically do a squash merge.

bmatherly:
> If we have a build problem with Frei0r then alI of this PR will be reverted unless you expect me to l tease it apart because I typically do a squash merge.

I think squash merge and full revert would be best. Or, we could try to test the build before we merge. I see that the PR only triggers the code style check. Is there a way to make it kick full builds from the branch?

Otherwise, I can checkout the branch on Windows and test the build. But I do not have a Mac to test on.

Or maybe it is easer to merge and then deal with a broken build. I would follow your preference.

ddennedy:
Remove the unrelated Frei0r commit and wait for me to get around to it.

ddennedy:
Sorry I overlooked the Frei0r PR in the description because I was only following updates on my phone. Let’s try it. I think it was only the Windows SDK  build that failed previously, and I can revisit that if needed.

ddennedy (approved):

PR #1635, opened 2025-02-21
Clang format

bmatherly:
As discussed on the forum.

This is more change that I expected. But I don't see anything offensive.


ddennedy (changes requested):
You also need to change `.github/workflows/check-code-format.yml`

bmatherly:
> You also need to change .github/workflows/check-code-format.yml

Good idea. Done. And i tested the workflow with passing and failing formatting (look at the commit history).
This might be a good one to squash due to all the commit noise.

PR #1636, opened 2025-02-23
Add a shortcut to hide the VUI

bmatherly:
As suggested here:
https://forum.shotcut.org/t/need-a-button-to-disable-the-frame-on-some-filters/46913

Posting this review for feedback on the name of the action and the shortcut.

image


ddennedy:
I think it is wordy. It is in the Player menu; so, does it need to include "Player." To confirm, it does not affect the grid overlay, correct?

bmatherly:
> I think it is wordy. It is in the Player menu; so, does it need to include "Player."

Good idea. Done.

image

PR #1645, opened 2025-03-16
Fix crash when selecting track after align to reference

bmatherly:
The align to reference track runs the XML consumer on the reference track. This replaces the previous consumer (which is the multitrack). This change restores the multitrack back on the track service.

As reported here:
https://forum.shotcut.org/t/align-to-reference-track-sometimes-fails-and-makes-my-project-crash/47843
https://forum.shotcut.org/t/crash-after-align-to-reference-track/47684

Depends on this PR: https://github.com/mltframework/mlt/pull/1074

Other ideas I had to fix this:
* Make sure the Shotcut code never uses the consumer() function on a service
* Invent a way to do a deep copy on any service and then run the XML consumer on that copy
* Find a way to modify the XML consumer so that it doesn't get saved on the service. This would require some kind of fundamental change to the way consumers connect to services. Maybe it could be an option for the connect function.

### DIFF
--- a/src/mltcontroller.cpp
+++ b/src/mltcontroller.cpp
@@ -542,6 +542,8 @@ QString Controller::XML(Service *service, bool withProfile, bool withMetadata)
     Service s(service                                  ? service->get_service()
               : (m_producer && m_producer->is_valid()) ? m_producer->get_service()
                                                        : nullptr);
+    // Save the consumer on this service so it can be restored.
+    mlt_service saveConsumer = mlt_service_consumer(s.get_service());
     if (!s.is_valid())
         return "";
     int ignore = s.get_int("ignore_points");
@@ -557,6 +559,8 @@ QString Controller::XML(Service *service, bool withProfile, bool withMetadata)
     c.start();
     if (ignore)
         s.set("ignore_points", ignore);
+    // Restore the consumer that was previously on this service
+    mlt_service_set_consumer(s.get_service(), saveConsumer);
     return QString::fromUtf8(c.get(kMltXmlPropertyName));
 }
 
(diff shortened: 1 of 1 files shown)


src/mltcontroller.cpp
```diff
@@ -557,6 +559,8 @@ QString Controller::XML(Service *service, bool withProfile, bool withMetadata)
     c.start();
```
bmatherly, line 559: The connect function immediately above this comment is where the consumer on the service is being replaced with the XML consumer. Before this call, the consumer is the multitrack (for a track). After this function closes, the XML consumer is deleted. But the track service still holds a reference. Not sure about how other services use the consumer internally. It must not be a problem for clips since we use this function regularly on clips.

ddennedy:
You do not want to add MLT c++ method for this new function?

bmatherly:
> You do not want to add MLT c++ method for this new function?

I did not try that. But I am willing to if you think it is a good idea. I expect that the mlt_service will be NULL for many producers. But I think I can still make the c++ wrappers work with that.

Part of my reason for this PR is to make sure this concept is a good idea at all. I don't completely understand why we keep a reference to a consumer on a service. Is that just so we can "work backwards" through the MLT structures if we need to?

FYI: the line that eventually crashes is here where it calls m_track.consumer()
https://github.com/mltframework/shotcut/blob/master/src/widgets/trackpropertieswidget.cpp#L96
There are two other places in Shotcut were we call Mlt::Service::consumer().
It appears that the MLT framework uses the function in MltFilteredConsumer and MltFilteredProducer.

ddennedy:
OK, I was thinking a script bindings might have a use for it, but I would not want it to return an invalid new object esp. considering it could be commonplace. We can leave it like you have it for now. Sadly, we must release a new MLT version and require it in Shotcut for this, but we have to cross that bridge eventually.

ddennedy:
I will bump the minimum MLT version in `CMakeLists.txt` when MLT is released just prior to next Shotcut release.

PR #1648, opened 2025-03-29
Update shotcut_hu.ts

JohnFowler58:
Localization of new strings and some corrections.


ddennedy:
Thank you for trying, but we only accept translation updates through transifex.com. Otherwise, the next time I pull from there your changes are overwritten.

JohnFowler58:
Well, I localize programs in my free time, as a hobby. I won't be paying transifex.com

ddennedy:
Transifex is free for open source projects and contributors. Do not be so easily tricked! You were warned when opening the pull request that we do not accept translation updates this way.

PR #1662, opened 2025-04-05
Add an option to re-run analysis for all 2-pass filters

bmatherly:
As suggested here:
https://forum.shotcut.org/t/force-stabilization-analysis-for-all-clips/47938

Posting for comment here. Not sure how I feel about the name and putting it in the Timeline->Edit menu. Open to suggestions.

image

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -2393,8 +2393,9 @@ void MainWindow::setupActions()
         if (parser.filters().size() > 0) {
             QMessageBox dialog(QMessageBox::Question,
                                windowTitle(),
-                               tr("This will start %1 analysis jobs. Continue?")
-                                   .arg(parser.filters().size()),
+                               tr("This will start %n analysis job(s). Continue?",
+                                  nullptr,
+                                  parser.filters().size()),
                                QMessageBox::No | QMessageBox::Yes,
                                this);
             dialog.setDefaultButton(QMessageBox::Yes);
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -38,6 +38,7 @@
 #include "docks/encodedock.h"
 #include "docks/filesdock.h"
 #include "docks/filtersdock.h"
+#include "docks/findanalysisfilterparser.h"
 #include "docks/jobsdock.h"
 #include "docks/keyframesdock.h"
 #include "docks/markersdock.h"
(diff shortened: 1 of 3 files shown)


ddennedy:
The name does not suggest all clips with the filter as the user requested. And that Edit menu is mainly sensitive to playhead or selection. I would expect it to rerun analysis on whatever is selected. Does the user need to first select all to affect all clips with the filter? What about the clips in the playlist with the filter?

bmatherly:
The implementation in this PR operates on all timeline filters (regardless of selection). It does not operate on Playlist or source.

I agree with your comments about the name and menu. But I did not see a better menu to put it in.

I can make it also operate on Playlist if we want. But then the menu will need to be at a higher level (not in the Timeline menu). One option would be to add it as a button in the "Resources" dialog (under View->Resources).

image

ddennedy:
> The implementation in this PR operates on all timeline filters (regardless of selection).

Imagine a large project with many of these. It would enqueue a large number of jobs. For example, someone did a normalize 2 pass and then heavily edited a long clip to remove dead spots. The new action needs to be very clear about "all." Also, some people might not want to include 2 pass normalize since that is probably not needed. I do not have good ideas at the moment, but neither do I think it is a very useful thing to include. A lot of desktop software now opens these very long menus, and I need to read through them all to look for the one I common thing I need or guess which is closest to my less common need. I do not know why you chose this.

ddennedy:
One idea is to just do something like this automatically when the video mode is changed, but don't reset the 2 pass audio norrmalize as that is not necessary (motion tracker should be included). This action already has a warning dialog. It would not mention this, but it is already longer than most people want to read.

bmatherly:
Thanks for continuing to brainstorm on this with me.

My primary use case for this is that the analysis becomes "stale" after some editing. For example, adding an audio filter before the normalize filter will cause it to have different results. Or splitting clips. I think it is common for a user to add a clip, add a two pass filter (stabilize or normalize), run the analysis. But then, some time later, add more filters or make cuts and leave the old analysis results stale. For this reason, I would like to include the normalize filter. The nice thing about the normalize filter is that it completes very quickly (compared to stabilize).

One way to avoid a surprise for the user would be to count the number of 2-pass filters and pop a confirmation dialog to say "This action will create 7 jobs in the job queue. Continue?".

You make a good comment about motion tracking. If the user changes the project video resolution, it won't help to just "reanalyze" the motions - because the starting square is specified by pixel location/size.  I'm going way out of bounds here, but maybe filters should be marked "stale" and somehow we would show a todo list for the user to revisit any filter that uses absolution pixel coordinates after they change the project resolution. I think that changing video mode wasn't a target use case for me because changing the video mode is kind of an "all bets are off" situation. The user is really forced to watch the entire project and look for problems caused by changing the video resolution.

ddennedy:
Those are all good points. The Resources dialog is not appropriate as that is media file oriented, but this acts on filters attached to clip objects in the project. Currently, the project-oriented functions are in the File menu: save, backup, versions, export EDL, etc. So, if it is broadened to include playlist objects as well, it can go in there. It feels weird but the best fit. If kept as timeline-only it could go into an Other submenu, but we do not yet have it, and a submenu with one item is inappropriate. If you want to add an Other, you could move Timeline > View > Properties into there. I think there could (should?) be a dialog that shows the amount of things, but also it could offer checkboxes to limit it to certain kinds of filters, but I do not have a strong feeling about that.

bmatherly:
I moved it under the file menu and added the playlist to the scope.
image

I also added a prompt dialog with the number of jobs.
image

ddennedy:
There are conflicts to resolve. 
i haven’t tried this yet, but what happens if someone runs this and then immediately tries to make an export? Will it think it needs to re-analyze a bunch of things?

src/mainwindow.cpp
```diff
@@ -2385,6 +2386,43 @@ void MainWindow::setupActions()
...
+        if (m_timelineDock->model()->tractor() && m_timelineDock->model()->tractor()->is_valid()) {
+            parser.start(*m_timelineDock->model()->tractor());
+        }
+        if (m_playlistDock->model()->playlist() && m_playlistDock->model()->playlist()->is_valid()) {
+            parser.start(*m_playlistDock->model()->playlist());
+        }
+        if (parser.filters().size() > 0) {
+            QMessageBox dialog(QMessageBox::Question,
+                               windowTitle(),
+                               tr("This will start %1 analysis jobs. Continue?")
```
ddennedy, line 2396: This needs to use the plural form for `tr`
`tr("This will start %1 analysis jobs. Continue?", nullptr, parser.filters().size())`

ddennedy (changes requested):

src/mainwindow.cpp
ddennedy: Making another reply in case you did not see it... I changed "jobs" to "job(s)". As I recall, I need to manually update `translations/shotcut_en.ts` to handle the plural forms for English. And I search for "(s)"  to help find them.

bmatherly:
> There are conflicts to resolve.

I rebased, resolved conflicts, squashed, and force pushed.

> what happens if someone runs this and then immediately tries to make an export?

For filters that had been previously re-analyzed, nothing will happen because the re-analysis jobs will be in the queue  before the export job.

For filters that were NEVER analyzed at all, there will be doubled jobs. One job from the "reanalyze" operation that hasn't completed yet. And then another job added by the export command (assuming the user clicked "yes" when they were prompted). 

I think that if the user does those operations in immediate succession, the prompt provided by the export operation will give them a clue that they should wait for the analysis jobs to complete.

>This needs to use the plural form for tr

Done

src/mainwindow.cpp
bmatherly: Done

bmatherly: > I changed "jobs" to "job(s)"

I'm glad you mentioned that. Done.

PR #1682, opened 2025-05-10
Update task icons

Firecrafter28:
Made the task icons (task-complete, task-attempt, task-reject) more modern


ddennedy:
Thanks for offering to help, but these icons come from a collection, and I do not want to change them at this time.

bmatherly:
I agree that for the oxygen theme we should keep the icon set consistent. The style has sort of a "Window 7" shimmer feel. And these don't match that.

However, for the light and dark themes, I have always been bothered that we used the oxygen icons for those. The "Default Icon" set that we use for other icons in Light and Dark themes are basic and lack the gradients and sheen of the Oxygen icons. I actually think these would be better for the Light and Dark themes.

PR #1687, opened 2025-05-26
Add an "extension" system.

bmatherly:
Not ready for merge. Posting for initial review and comments.

This adds a new "extension" system that can be used to download additional features. This commit uses the extension system to allow the user to download additional Whisper.cpp models.

The "Speech to Text" dialog now has a new list that shows all the available models. Models that are downloaded have a green check mark. Models that are not downloaded have a download icon.
image

If the user clicks on a model with a green check mark, it becomes the current model. If the user clicks on a model with a download icon,  they are prompted to download it:

image

image

Models are stored in the users data directory in an "extensions" folder. Example on Linux:
`/home/brian/.local/share/Meltytech/Shotcut/extensions/Whisper Model/ggml-small.bin`

The available models are listed in the QML extension file: whispermodel.qml
The extension file is distributed with the Shotcut binary package.

Discussion: should Shotcut offer a way to delete the models? Perhaps a right-click context menu?

Comments are welcome.


ddennedy:
I think it looks good, and I agree there can be a context menu to **Remove** a model. (Recent and Recent Project list view have that menu item.)

I see how the model can be used for other sets of extensions and have different context-sensitive UIs where it makes sense instead of perhaps one central UI. Maybe we will change our mind and later have a common UI, but we can run with this for now.

bmatherly:
I see what you mean. Here is what I propose:

* Shotcut will still include the extension file with the package in the application directory
* I will add a button to the dialog called "Update Models"
* When the user clicks the button, an extension file will be downloaded from the Shotcut website and stored in the user directory
* When Shotcut starts up, it will first look in the user directory and if it finds an extension file, it will use that one
* If an extension file is not found in the user directory, it will use the one in the application directory that came with Shotcut

Thoughts?

bmatherly:
Thanks for your comments. I added the button:

image

When the new extension file is downloaded, I just refresh the model, rather than checking the version since the cost to refresh is negligible.

src/dialogs/transcribeaudiodialog.cpp
```diff
@@ -225,6 +232,62 @@ TranscribeAudioDialog::TranscribeAudioDialog(const QString &trackName, QWidget *
     });
     configLayout->addWidget(modelBrowseButton, 2, 2, Qt::AlignLeft);
 
+    // Update Model button
+    QPushButton *updateModelsButton = new QPushButton(tr("Refresh Models"), this);
+    connect(updateModelsButton, &QAbstractButton::clicked, this, [&] {
+        QString downloadUrl
+            = "https://raw.githubusercontent.com/mltframework/shotcut/"
```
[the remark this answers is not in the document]
ddennedy: Change the URL to https://check.shotcut.org/whispermodel.qml`, and upload it to the `shotcut_web` git repo, which is where `version.json` is maintained. Also, make the URL a `static const` at the top of this file.

src/settings.cpp
```diff
@@ -1527,8 +1527,7 @@ QString ShotcutSettings::whisperModel()
 {
     QDir dataPath = QmlApplication::dataDir();
     dataPath.cd("shotcut/whisper_models");
-    return settings.value("subtitles/whisperModel", dataPath.absoluteFilePath("ggml-base-q5_1.bin"))
-        .toString();
+    return settings.value("subtitles/whisperModel", "").toString();
```
ddennedy, line 1530: Is it your intention to not include a default model with the downloaded package any longer? It is still bundling but no longer recognized from the install. Existing users are in for a shock when they see the red text: Select a model or browse to choose one--especially if they are updating an existing install. That's somewhat fine, but I need to change the build scripts. Do you think we should make the old, bundled model bold in the list to make it easier to get back to what you had before? Otherwise, I worry people will not know what to do and simply choose the first.

bmatherly: > Is it your intention to not include a default model with the downloaded package any longer?

That is my intention. But I had not changed the build script yet to give more time for testing.

> Otherwise, I worry people will not know what to do and simply choose the first.

Well, that is a very good comment that I did not think of. I think it would be good to remove it from the build script.

To help the user, I can change the text in the list to say "Base Multilingual Quantized 5_1 (Recommended)"

bmatherly: I'm not sure that this will send the message we are looking for:
image

Another idea I have: When we open the dialog, we could detect if they do not have any models downloaded. In that case, we could pop up a prompt that says:

"You have not downloaded any models. Click a model to download one. Would you like to download a common model to get started?"

If they click "yes", we would start the download for them.

bmatherly: This is done. I also added a way to delete models.

PR #1697, opened 2025-07-06
Updating Swedish translation

yeager:


ddennedy:
Thank you! They needlessly break hundreds of thousands of URLs every few years.  :-(

Issue #1638, opened 2025-03-11
flatpak on linux aarch64 crashes


odlg:
flatpak version of shotcut 25.01.25 crashes on ARM 64-bit linux.

Running gdb on it reveals this:
```
Reading symbols from /app/bin/shotcut...
Reading symbols from /usr/lib/debug/app/bin/shotcut.debug...
(gdb) run
Starting program: /app/bin/shotcut 

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.fedoraproject.org/>
Enable debuginfod for this session? (y or [n]) y
Debuginfod has been enabled.
To make this setting permanent, add 'set debuginfod enabled on' to .gdbinit.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/aarch64-linux-gnu/libthread_db.so.1".
[New Thread 0xfffff02defe0 (LWP 15)]
[New Thread 0xffffee2defe0 (LWP 16)]
[New Thread 0xffffedacefe0 (LWP 17)]
[New Thread 0xffffed2befe0 (LWP 18)]
[New Thread 0xffffdfffefe0 (LWP 19)]
[New Thread 0xffffdf7eefe0 (LWP 20)]
[New Thread 0xffffdefdefe0 (LWP 21)]
[New Thread 0xffffdd1aafe0 (LWP 22)]
[New Thread 0xffffc1fbefe0 (LWP 23)]
libDeckLinkAPI.so: cannot open shared object file: No such file or directory
[New Thread 0xffffc17aefe0 (LWP 24)]
[New Thread 0xffffc0efafe0 (LWP 25)]
[New Thread 0xffffb7ffefe0 (LWP 26)]

Thread 1 "shotcut" received signal SIGSEGV, Segmentation fault.
0x0000fffff5fc261c in QtPrivate::QMetaTypeForType<QImage>::getDtor()::{lambda(QtPrivate::QMetaTypeInterface const*, void*)#1}::_FUN(QtPrivate::QMetaTypeInterface const*, void*) () at /run/build-runtime/qt6-qtbase/src/corelib/kernel/qmetatype.h:2470
2470	                reinterpret_cast<S *>(addr)->~S();
(gdb) bt
[...]

PR #1698, opened 2025-07-08
add github ci to build linux aarch64

li-nkSN:


.github/workflows/build-linux_arm64.yml
```diff
@@ -0,0 +1,102 @@
+name: build-shotcut-linux-aarch64
+
+on:
+  workflow_dispatch:
+  schedule:
+    # nightly
+    - cron: '0 8 * * *'
+
+env:
+  IMAGE_NAME: mltframework/shotcut-build:qt6.4.3-ubuntu22.04-arm64
```
ddennedy, line 10: Can you please use a newer version of Qt? The x86-64 version only uses version 6.4 for compatibility with glibc version in Ubuntu 20.04. I plan to upgrade it with Ubuntu 26.04 at the latest. The newer versions of Qt work better on Wayland.

li-nkSN: > Did you know there is a Flatpak of Shotcut for aarch64? Please try it. Some people have reported it as failing to launch. See https://github.com/mltframework/shotcut/issues/1638. It might be better if you can help fix that because I am not very confident these aarch64 portable/AppImage/Snap are going to work.

```
flatpak run org.shotcut.Shotcut


libDeckLinkAPI.so: cannot open shared object file: No such file or directory
```

li-nkSN: > The docker image is maintained at https://github.com/mltframework/mlt-scripts/tree/master/docker

There is no mltframework/shotcut-build:qt6.4.3-ubuntu22.04-arm64. So, this PR is not going to work, and I do not have an arm64 Linux system on which to help and verify a build. You will need to make a PR on mlt-scripts this builds the image that this PR will use. From https://doc.qt.io/qt-6/linux.html I see that might need to be based on Ubuntu 24.04.

https://github.com/mltframework/mlt-scripts/pull/36/files

PR #1700, opened 2025-07-08
Remove custom OpenCV build logic for M_PI workaround

thiru31:
This PR removes the custom `build_opencv` function and reverts the build command for OpenCV to its original form.

Previously, a workaround was required to manually patch `M_PI` into `carotene/definitions.hpp` when using an older OpenCV branch that lacked this definition. OpenCV 4.x has since addressed this issue in commit [ce1d840adf](https://github.com/opencv/opencv/commit/ce1d840adfe68504fd9b35f1568286e1911599c9), adding this custom build logic is unnecessary.

Changes:
- Removed `build_opencv` function.
- Reverted OpenCV build command to `ninja -C build -j $MAKEJ`.

This simplifies the build script by taking advantage of OpenCV repo's improvements.


thiru31:
That was real quick, thanks for the response & maintaining shotcut.

PR #1754, opened 2025-11-22
Fix a tiny error in the Catmull-Rom interpolation calculation

traprog:
A smooth keyframe curve in the Keyframes dock is slightly distorted due to an error in the Catmull-Rom interpolation calculation. It's not a huge deal because it's a hard-to-notice aesthetics issue.

curve2


ddennedy:
Thanks!

PR #1777, opened 2026-01-17
Improve Inno Setup script - Use GNU RTF license file

bovirus:
@ddennedy 

Please check if it's Ok for you.

Instead TXT file you can use GNU R3.0 RTF license file (with better use of GUI for license).

GNU RTF license - https://www.gnu.org/licenses/gpl-3.0.rtf

Note: if you can update to Inno Setup 6.7.0 you can improve changing in [Setup] section

WizardStyle=modern dynamic

Reference about WizardStyle - https://jrsoftware.org/ishelp/topic_setup_wizardstyle.htm

Thanks.


bovirus:
Thanks.

PR #1779, opened 2026-01-30
Add player zoom with mouse wheel

bmatherly:
Zoom is triggered with the mouse wheel and the Alt key

As suggested here:
https://forum.shotcut.org/t/pourrait-on-avoir-une-possibilitee-grossissement-plus-fin-detaille-dans-loption-zoom/47891/7

Posting this for discussion and to wait until after the next release.

For the player, wheel is already used to scale the SPR VUI size. And wheel + CTL is used to rotate the SPR VUI angle. So all that is left for zoom is wheel + ALT.

However, I wanted to mention that wheel with no modifiers is already used to zoom in the notes panel and the video zoom scope widget. So I wonder if we want to CHANGE the SPR VUI to use whell + ALT for scaling, and then use wheel with no modifiers to zoom the player.

Pro: I think this would be more intuitive for new users
Con: It would be a change of behavior for existing users.

### DIFF
--- a/src/widgets/screenselector.cpp
+++ b/src/widgets/screenselector.cpp
@@ -59,19 +59,23 @@ void ScreenSelector::setSelectedRect(const QRect &rect)
 void ScreenSelector::startSelection(QPoint initialPos)
 {
 #if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
-    const auto p = MAIN.geometry().center();
-    const auto id = MAIN.window()->winId();
-    for (auto screen : QGuiApplication::screens()) {
-        if (screen->geometry().contains(p)) {
-            m_useDBus = screen->grabWindow(id, p.x(), p.y(), 1, 1).isNull();
-            if (m_useDBus) {
-                emit screenSelected(m_selectionRect);
-                return;
+    if (m_selectionRect.isEmpty()) {
+        // Using as a single point selection.
+        const auto p = MAIN.geometry().center();
+        const auto id = MAIN.window()->winId();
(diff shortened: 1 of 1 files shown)


ddennedy:
Please refer to [this page](https://shotcut.org/howtos/keyboard-shortcuts/) for an overview of all existing wheel behaviors. I think the zoom modifier should be more consistent with Timeline and Keyframes rather than the lesser used Notes and Video Zoom Scope. As you noted, `Ctrl` is already used by the rectangle control to rotate. However, one can argue that `Shift` to change track height is a form of zoom, and this could use that.

Another thing to consider is that perhaps when the player zoom shows scroll bars people expect to be able to use the mouse wheel to scroll just like any other UI panel (or window outside of Shotcut) with scroll bars. Typically, no modifier = scroll vertically, and `Alt` to scroll horizontally. Only Timeline swaps those because it is somewhat typical for NLEs to do that since primary navigation is horizontal unlike say, a list of files or web page. Outside of middle mouse button drag, is scrolling the zoomed-in player something you want to add onto this? Consider that pressing and holding the wheel to drag is a foreign concept to a lot of people.

I do not yet know how to reconcile all of those.

bmatherly:
Thanks for the thoughtful comments. I took the conservative approach on the player and just used the SHIFT modifier for the new zoom function. There is not change to existing behaviors.

For the video zoom scope, I did make a change because I have been using that a lot lately and the usability is poor:
-  Scroll vertically (NEW)
-  Hold Alt to scroll horizontally (NEW)
-  Hold Ctrl to zoom in and out (Formerly no modifiers)

PR #1813, opened 2026-04-11
Add support for experimental add-on filters

bmatherly:
This includes a new selection dialog and generated UIs
The --experimental option is required to expose the selection
--experimental is only available when Shotcut is compiled for debug

When --experimental is enabled, the filter selection menu adds a hamburger menu before the search bar with a single action called "Manage Add-on Filters"
image

This action opens a new Add-on Filter selection dialog which lists ALL available filters in MLT:
image

The user can select and de-select MLT filters to be added to the main filter selection as Add-on Filters. After the dialog is closed, the main filter selection is updated with the selected (or de-selected) filters
image

When the user selects an Add-on filter that has been added to a clip, the UI for that add-on is auto-generated and cached in a temporary folder.
image

The help button (?) opens a non-modal dialog with a detailed view of all the properties for that add-on filter
image

I find this feature useful for exploring filters - especially the host of OpenFx filters. This has also exposed some opportunities for improving metadata for some MLT filters - which I will submit in different PRs.

Review comments are welcome.

I recommend that we wait until after the April release before we consider merging this.


src/controllers/filtercontroller.cpp
```diff
@@ -122,6 +166,154 @@ void FilterController::loadFilterMetadata()
...
+        }
+
+        addMetadata(meta);
+    }
+}
+
+bool FilterController::ensureAddOnTempDir()
+{
+    if (!m_addOnTempDir.isValid()) {
+        m_addOnTempDir = QTemporaryDir(QDir::tempPath() + "/shotcut-addon-XXXXXX");
```
bmatherly: A new Temporary directory is created for every Shotcut instance. If someone opens/closes Shotcut a lot, they can accumulate. But this handles the case of regenerating the UI for software updates or multiple versions installed at the same time. It also makes development easy since I don't have to delete the generated files every time I make a change to the code.

src/controllers/filtercontroller.cpp
```diff
@@ -256,6 +448,12 @@ void FilterController::setCurrentFilter(int attachedIndex)
     QmlMetadata *meta = m_attachedModel.getMetadata(m_currentFilterIndex);
     QmlFilter *filter = nullptr;
     if (meta) {
+        if (meta->objectName().startsWith(QStringLiteral("addOn.")) && !ensureAddOnFilterQml(meta)) {
```
bmatherly, line 482: The UI QML for an add-on filter is only generated when absolutely needed  - when a user has selected an add-on filter in the UI. If present, the cached version is used. Else, it is generated. So, in many cases, the temporary directory will never be created if the user opens shotcut, but never uses an add-on filter.

src/qml/views/filter/FilterMenu.qml
```diff
@@ -78,6 +78,26 @@ Rectangle {
 
             Layout.fillWidth: true
 
+            Shotcut.Button {
+                id: addOnMenuButton
+
+                visible: enableAddOns
```
bmatherly: This button only appears when --experimental is passed to Shotcut.

src/main.cpp
```diff
@@ -165,6 +165,13 @@ class Application : public QApplication
         QCommandLineOption gpuOption("gpu",
                                      QCoreApplication::translate("main", "Use GPU processing."));
         parser.addOption(gpuOption);
+#ifdef QT_DEBUG
```
bmatherly, line 168: The --experimental option is only available when Shotcut is compiled for debug.

src/mltcontroller.cpp
```diff
@@ -64,8 +64,13 @@ Controller::Controller()
...
 {
     LOG_DEBUG() << "begin";
-    if (!qEnvironmentVariableIsSet("MLT_REPOSITORY_DENY"))
-        ::qputenv("MLT_REPOSITORY_DENY", "libmltqt:libmltglaxnimate:libmltopenfx");
+    if (!qEnvironmentVariableIsSet("MLT_REPOSITORY_DENY")) {
+        const bool experimental = qApp && qApp->property("experimental").toBool();
+        if (experimental)
+            ::qputenv("MLT_REPOSITORY_DENY", "libmltqt:libmltglaxnimate");
+        else
+            ::qputenv("MLT_REPOSITORY_DENY", "libmltqt:libmltglaxnimate:libmltopenfx");
```
bmatherly, line 72: The experimental option also enables openfx to be loaded at startup.

src/settings.cpp
```diff
@@ -1069,11 +1069,22 @@ QString ShotcutSettings::filterFavorite(const QString &filterName)
...
 {
     settings.setValue("filter/favorite/" + filterName, value);
 }
 
+QStringList ShotcutSettings::addOnFilterServices() const
+{
+    return settings.value("filter/addOnServices").toStringList();
+}
+
+void ShotcutSettings::setAddOnFilterServices(const QStringList &services)
```
bmatherly, line 1083: When the user selects add-on services in the dialog, they are saved in the settings so they can be reloaded very time shotcut starts (only with --experimental)

bmatherly (commented):
Some general comments for understanding intent.

src/controllers/filtercontroller.cpp
```diff
@@ -27,18 +28,39 @@
...
+}
+
+QString addOnServiceFromObjectName(const QString &objectName)
+{
+    static const QString kPrefix = QStringLiteral("addOn.");
+    if (!objectName.startsWith(kPrefix))
+        return QString();
+
+    return objectName.mid(kPrefix.size());
+}
```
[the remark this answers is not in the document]
bmatherly: done

src/controllers/filtercontroller.cpp
```diff
@@ -27,18 +28,39 @@
...
+
 FilterController::FilterController(QObject *parent)
     : QObject(parent)
+    , m_addOnTempDir(QString())
```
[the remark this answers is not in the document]
bmatherly: done a different way to maintain lazy creation.

src/dialogs/addonfiltersdialog.cpp
```diff
@@ -0,0 +1,197 @@
...
+    m_listWidget->setSortingEnabled(true);
+    m_listWidget->setColumnCount(3);
+    m_listWidget->setHeaderLabels({tr("Title"), tr("Service"), tr("Type")});
+    m_listWidget->header()->setStretchLastSection(true);
+    m_listWidget->header()->setSectionResizeMode(0, QHeaderView::Interactive);
+    m_listWidget->header()->setSectionResizeMode(1, QHeaderView::ResizeToContents);
+    m_listWidget->header()->setSectionResizeMode(2, QHeaderView::Stretch);
+    m_listWidget->header()->setSortIndicatorShown(true);
+    m_listWidget->sortItems(0, Qt::AscendingOrder);
+    layout->addWidget(m_listWidget, 1);
```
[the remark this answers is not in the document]
bmatherly: done

src/controllers/addonqmlgenerator.cpp
```diff
@@ -0,0 +1,851 @@
...
+#include <QStringList>
+#include <QTextStream>
+
+QString quotedJsString(const QString &value)
+{
+    QString s = value;
+    s.replace('\\', "\\\\");
+    s.replace('\'', "\\'");
+    return QStringLiteral("'%1'").arg(s);
+}
```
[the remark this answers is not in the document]
bmatherly: done

src/controllers/addonmetadataparser.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include <QSet>
+
+bool parseYesNoBool(const char *value)
+{
+    if (!value)
+        return false;
+    const QString text = QString::fromUtf8(value).trimmed().toLower();
+    return text == QStringLiteral("yes") || text == QStringLiteral("true")
+           || text == QStringLiteral("1");
```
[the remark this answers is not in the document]
bmatherly: done

src/controllers/addonqmlgenerator.cpp
```diff
@@ -0,0 +1,851 @@
...
+           "        anchors.top: parent.top\n"
+           "        anchors.margins: 8\n"
+           "        columns: 4\n"
+           "        columnSpacing: 8\n"
+           "        rowSpacing: 6\n\n"
+           "        Label {\n"
+           "            Layout.columnSpan: 3\n"
+           "            Layout.fillWidth: true\n"
+           "            text: root.filterDescription.length > 0 ? qsTr('Add-on Filter: "
+           "%1').arg(root.filterDescription) : qsTr('Add-on Filter')\n"
```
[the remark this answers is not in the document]
bmatherly: This is intentional.

src/controllers/addonmetadataparser.cpp
```diff
@@ -0,0 +1,100 @@
...
+    descriptor.title = service;
+
+    if (!mltMetadata || !mltMetadata->is_valid())
+        return descriptor;
+
+    descriptor.title = QString::fromUtf8(mltMetadata->get("title"));
+    if (descriptor.title.isEmpty())
+        descriptor.title = service;
+
+    descriptor.description = QString::fromUtf8(mltMetadata->get("description"));
```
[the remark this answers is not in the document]
bmatherly: MLT is consistent. I removed unnecessary capital checks in AddOnServiceModel

src/qmltypes/qmlfiltermenu.h
```diff
@@ -0,0 +1,37 @@
...
+#define QMLFILTERMENU_H
+
+#include <QObject>
+
+class QmlFilterMenu : public QObject
+{
+    Q_OBJECT
+
+public:
+    explicit QmlFilterMenu(QObject *parent = 0);
```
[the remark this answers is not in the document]
bmatherly: Done

ddennedy:
Thank you; it is nice to have. The first filter I tried `avfilter.rotate` crashed (Windows). Not looked into it yet but a reminder this is full of landmines. I think we should filter out the ones that already have an installed QML if not too much trouble.

src/dialogs/addonfiltersdialog.cpp
```diff
@@ -0,0 +1,200 @@
...
+    auto *description = new QLabel(tr("Select filters to expose as add-on filters."), this);
+    description->setWordWrap(true);
+    layout->addWidget(description);
+
+    auto *searchLayout = new QHBoxLayout();
+    m_searchField->setPlaceholderText(tr("Search service or title"));
+    searchLayout->addWidget(m_searchField);
+    auto *clearButton = new QPushButton(tr("Clear"), this);
+    searchLayout->addWidget(clearButton);
+    layout->addLayout(searchLayout);
```
ddennedy, line 50: Use `LineEditClear`

src/qml/views/filter/FilterMenu.qml
```diff
@@ -78,6 +78,26 @@ Rectangle {
...
+                padding: 2
+                implicitWidth: 20
+                implicitHeight: 20
+                icon.name: 'show-menu'
+                icon.source: 'qrc:///icons/oxygen/32x32/actions/show-menu.png'
+                onClicked: addOnMenu.popup()
+
+                Shotcut.HoverTip {
+                    text: qsTr('Add-on filter options')
```
ddennedy: It is rather weird to have a menu button in an unusual location with this description and one menu item. Do you suspect there will be other menu items?

How about a button that does not open a menu and is on the `filtersview` alongside the other buttons? You can reuse the gear (Jobs) icon.

ddennedy:
> avfilter.rotate crashed

I found that the avformat module was incorrect about metadata and pushed a change. Now, it does not crash.

bmatherly:
> I think we should filter out the ones that already have an installed QML if not too much trouble.

I am on the fence about this. On the one hand, it would reduce redundancy. On the other hand, filters can be exposed/represented in many ways. Maybe we are exposing a service through our curated UI, but not exposing some of the parameters or doing it in a limited way. It might be good if the experimenter can work with the raw filter even if we have exposed it through our supported filters.

As a reference, the following services are used in multiple Shotcut filters:

- autofade (2)
- brightness (4)
- channelcopy (2)
- frei0r.alpha0ps (2)
- frei0r.IIRblur (3)
- ladspa.1204 (2)
- lift_gamma_gain (2)
- mask_start (5)
- movit.opacity (2)
- panner (2)
- qtcrop (2)
- qtext (2)
- volume (4)

As a variation on this comment, for each add-on filter, I could add a note somewhere to indicate which supported filters are using that MLT filter. Perhaps as a column in the manager dialog. Or maybe a tool tip in the generated UI.

ddennedy:
> I am on the fence about this.

OK I see your point.

src/dialogs/addonfiltersdialog.cpp
bmatherly: Done

src/qml/views/filter/FilterMenu.qml
bmatherly: Yeah. There is no great place for it. The filter panel does not have a menu - unlike most panels. I was thinking that maybe in the future we would add things to it. But I don't have anything in mind.

I feel like the button should not be in the FiltersView because the user might use the dialog and then expect that a filter they added should appear in the list of added filters. But the dialog really only affects what shows in the filters selection menu.

I tried putting the "cog" button ahead of the favorites button (still only visible when experimental is enabled). The button opens the dialog directly (rather than show a single menu). It has a tooltip that says "Manage Add on filters".

Thoughts on this?
image

ddennedy: OK for now

bmatherly:
This last commit adds columns to see what image formats each add-on filter supports. This is the last change I have in mind for this branch.

Depends on this PR: https://github.com/mltframework/mlt/pull/1240

image

src/mltcontroller.cpp
ddennedy: FYI, I am going to decouple the denial of the OpenFX module from `--experimental` because I want to be able to soft launch support for OpenFX and VST2 without the UI generator. The UI for NTSC-rs is already committed, and I have another ready for a VST2. I do not think the UI generator is ready for soft launch because it is rather basic and the number of filters and their usefulness is well, kind of like the twilight zone.
I will add a crash handler to make Shotcut restart in a "safe mode" with libmltopenfx denied (plus another env change to block vst2). I am thinking to add a menu item to Settings such as "Exit Safe Mode" or "Restart with OpenFX/VST2". We will mention `--experimental` in release notes, but keeping out of the UI and only in the CLI makes it out-of-reach for most Windows and macOS users, which is good. People should use `--experimental` to help qualify plugins as interesting & usable. Then, I can generate a good QML UI quickly with the custom Copilot prompt perhaps until the UI generator is more sophisticated if we decide to go that way.

bmatherly: That all sounds good to me. Let me know if there are any tasks that you have in mind that I can help with. I don't have any more work planned for this, but I am happy to help.

Do we want to package any OpenFX plugins with Shotcut? It wouldn't be hard to package the Natron plugins, for example.

ddennedy: No, I do not want to bundle any OpenFX plugins until we decide it is critical. I have not found many useful Natron OpenFX plugins yet. Yesterday, I found the UI generator does not handle a pair of numbers correctly such as point and sizes. That was causing some problems. I think you should work on that. It was a Crop filter; I do not have the full ID handy. I can get it later if you need.
There is a lot to debug to figure out why these things are not working. The GMic ones are very slow and silly. At this point the OpenFX and VST2 story is mostly about working with external plugins. Later, we might port some things from frei0r to openfx for rgba64 and threading.

PR #1845, opened 2026-06-06
Add graphics adapter (GPU) selection

naughtyGitCat:
## What

Adds a **Settings → Graphics Adapter** submenu to choose which physical GPU Shotcut renders and encodes with.

## Why

On a desktop where a low-power integrated GPU drives the display while a more capable discrete GPU is also installed (e.g. an NVIDIA card with no monitor attached), Shotcut renders the preview — and, depending on the chosen encoder, exports — on the integrated GPU, with no way to direct the work to the discrete GPU. This adds that choice.

## How

- **Enumeration** (`gpuinfo.{h,cpp}`, new): list physical adapters via DXGI. Some drivers (notably AMD integrated graphics) report the same physical GPU many times with different LUIDs, so adapters are de-duplicated by their stable hardware identity (vendor + device + subsys + revision).
- **Preview / UI**: the chosen GPU is applied by setting `QT_D3D_ADAPTER_INDEX` before the RHI is initialized, so the Direct3D scene graph (the default backend on Windows) creates its device on that adapter.
- **Stable selection**: the raw DXGI adapter index is **not stable across runs** on some systems (the number of duplicate integrated-GPU entries varies), so the selection is persisted as the GPU's vendor + device id and the live index is resolved at startup.
- **Export**: when a specific GPU is selected, export uses the hardware encoder family matching that vendor (NVIDIA → `*_nvenc`, AMD → `*_amf`, Intel → `*_qsv`), and hardware encoding defaults on.
- **UI**: `Settings > Graphics Adapter` with the usual restart prompt, following the existing "Display Method" menu pattern.

## Scope / notes

- Currently implemented for the **Windows Direct3D** backend; `enumerateGpuAdapters()` returns empty elsewhere and the menu is hidden, so other platforms are unaffected.
- New `gpuinfo.{h,cpp}` carry the Meltytech copyright header; formatted with clang-format 14.
- **UI translations are intentionally not included** (handled via Transifex).

## Tests

Adds a QtTest suite (`tests/test_gpuinfo.cpp`, 11 cases) covering encoder-family selection, codec-type matching, fallbacks, adapter de-duplication and live index resolution, behind a new off-by-default `SHOTCUT_BUILD_TESTS` option (default/official builds unchanged):

```
cmake -B build -DSHOTCUT_BUILD_TESTS=ON .
ninja -C build test_gpuinfo && ctest --test-dir build --output-on-failure
```

Manually verified on an AMD iGPU + NVIDIA RTX 3090 machine: selecting the 3090 moves preview rendering to it (confirmed via `nvidia-smi`) and routes hardware export through NVENC.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ddennedy:
Thank you for your contribution. It is a good idea, but it would nice to have some cross-platform support, particularly on Linux. I do not think it matters much on macOS any more since the advent of Apple Silicon. In any case, I just had to push a fix for the next release that created a merge conflict. Please fix that. Then, I will not be able to merge this until after the next release due by end of the month plus a quiet week. I require that I am able to test and support nearly everything in Shotcut because I cannot trust any outsider to be there to help. Currently, I do not have any multi-GPU system, but I have a couple that I can swap a PCIe card if the power supply supports it. So, in the meantime, I will try to get there in position to test and support it and possibly extend to Linux.

src/docks/encodedock.cpp
```diff
@@ -174,8 +175,14 @@ EncodeDock::EncodeDock(QWidget *parent)
     ui->videoCodecCombo->model()->sort(0);
     ui->videoCodecCombo->insertItem(0, tr("Default for format"));
 
-    ui->hwencodeCheckBox->setChecked(Settings.encodeUseHardware()
-                                     && !Settings.encodeHardware().isEmpty());
+    bool useHardwareEncode = Settings.encodeUseHardware() && !Settings.encodeHardware().isEmpty();
+    // If the user selected a specific GPU and never explicitly configured hardware
+    // encoding, default it on so the chosen GPU is actually used for export. Once the
+    // user toggles the checkbox, that saved preference is honored from then on.
+    if (!Settings.encodeUseHardwareWasSet() && Settings.gpuAdapterVendorId() != 0
```
ddennedy: I do not like this. The rationale in the comment conflates hardware usage in the UI (`QT_D3D_ADAPTER_INDEX`) with the encoder, which is optional. A lot of users do not want hardware encoding considering it inferior quality. I realize there is some relationship and value, however, in automatic switching of the hardware encoding vendor in the configured hardware vendors. It is the enabling of the opt-in hardware encoding option that is a problem. This change effectively makes it opt-out.

src/gpuinfo.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+// Platform-independent: pick the hardware encoder matching the GPU vendor, with a
+// fallback to the first type-compatible encoder. Kept out of the Windows-only block
+// so it builds and is unit-testable everywhere.
+QString preferredHardwareVcodec(const QStringList &hardwareCodecs,
+                                const QString &softwareVcodec,
+                                uint vendorId)
+{
+    auto matchesType = [&softwareVcodec](const QString &hw) {
+        return (softwareVcodec == QLatin1String("libx264") && hw.startsWith(QLatin1String("h264")))
```
ddennedy: This needs to block using a H.264 hardware encoder for 10-bit video because that is inherently incompatible. I do not know any hardware encoder that supports it. That is the merge conflict.
`const bool is10bit = QString::fromLatin1(preset.get("pix_fmt")).contains("p10le");`
plus
`&& !is10bit`

src/docks/encodedock.cpp
ddennedy: Upon closer inspection, I think it is the comment is a little misleading or something misunderstood about the contributor's expectation of how things actually work. This will effectively remain opt-in until available hardware encoders are configured either through automatic probing or manually in the configuration dialog, resulting in `Settings.encodeHardware().isEmpty()` false. Thus, it will be extremely rare that is non-empty and  `Settings.encodeUseHardwareWasSet()` is false. So, why bother adding this complication and more code to read, understand, and maintain?

src/main.cpp
```diff
@@ -423,6 +424,24 @@ int main(int argc, char **argv)
...
+    // the environment by the user takes precedence. The adapter index is resolved live
+    // from the GPU's stable vendor+device identity because some drivers enumerate
+    // adapters in an unstable order across runs. See gpuinfo.cpp and Settings.
+    {
+        const uint gpuVendorId = Settings.gpuAdapterVendorId();
+        if (gpuVendorId != 0 && !qEnvironmentVariableIsSet("QT_D3D_ADAPTER_INDEX")) {
+            const int gpuAdapterIndex = gpuAdapterIndexFor(gpuVendorId,
+                                                           Settings.gpuAdapterDeviceId());
+            if (gpuAdapterIndex >= 0)
+                ::qputenv("QT_D3D_ADAPTER_INDEX", QByteArray::number(gpuAdapterIndex));
```
ddennedy: I think a conceptual problem with this contribution is that it only affects GPU usage of the user interface including video display and encoding. However, there are two other big components that are not affected by this resulting in confusion leading to support questions and bug reports: hardware decoding and the GPU processing mode (runs on OpenGL currently). Today, on Windows, it is simple: there is code at the beginning of `main.cpp` to prefer a discrete NVIDIA or AMD GPU and no choice.

ddennedy (changes requested):

naughtyGitCat:
Thanks for the detailed review — much appreciated. I've pushed an update (rebased onto master):

1. **Merge conflict** — rebased onto the latest master, so it now includes your "Fix 10-bit H.264 hardware encode fails" change.

2. **Auto-enabling hardware encoding** — removed. You're right that it effectively turned the opt-in hardware encoding into opt-out, and as you noted it would rarely trigger anyway. The checkbox is back to being driven purely by the user's saved preference; only the vendor matching among already-configured hardware encoders remains.

3. **10-bit H.264** — folded your `is10bit` guard into the encoder-selection helper, so a 10-bit `pix_fmt` skips H.264 hardware encoders (HEVC/AV1 hardware are unaffected).

On the broader point: you're right that this currently only affects the UI/display adapter and the optional encoder, not hardware decoding or the GPU (Movit) processing mode. I scoped it conservatively for that reason — `enumerateGpuAdapters()` returns empty on non-Windows / non-D3D backends, so the menu is simply hidden and other configurations are unaffected. I'm happy to extend it to cover hardware decode and the GPU processing mode, and to add a Vulkan path for Linux (`QT_VK_PHYSICAL_DEVICE_INDEX`) — though real multi-GPU testing there needs the hardware. No rush on my end; I understand this is post-release, and I'm glad to iterate once you're set up to test.

src/main.cpp
ddennedy: Also set `MLT_AVFORMAT_HWACCEL_DEVICE` to the same GPU index. That will target the hardware decoder.

naughtyGitCat:
Thanks — good call. I've set `MLT_AVFORMAT_HWACCEL_DEVICE` to the same resolved adapter index alongside `QT_D3D_ADAPTER_INDEX` in `main.cpp`, so the hardware decoder now targets the selected GPU as well. Since the melt export subprocess inherits the environment, it applies to both preview and export.

One note on the Linux CI: the `build` job failed at `git checkout origin/gpu-adapter-selection-pr` (`pathspec ... did not match`), which I believe is because the branch only exists on the fork rather than the upstream remote the build script checks out — not a compile error. Happy to adjust anything on my end if that's not expected.

ddennedy (approved):
I successfully tested this and will merge it a few days after the next release.

ddennedy:
merged manually in bc511a869aa461b1faa526d74c5e5af3a265957d since I had to cleanup some line endings in settings.cpp and clang-format gpuinfo.cpp

PR #1847, opened 2026-06-08
Addon generator

bmatherly:
Depends on the metadata changes in here: https://github.com/mltframework/mlt/pull/1254


ddennedy:
I think it would be nice to have 2 branches: one dependent upon the MLT change and another not. I like the Channel Mask and limited description height. I myself was thinking of eliding to one line with ellipses and a tooltip for all of it.

Plus, for release I still want to get in proper support for numeric pair fields, and I ran into a single numeric field using a simple text edit instead of SpinBox. Should I work on those?

ddennedy:
Or if you create 2 MLT branches this is acceptable in whole.

bmatherly:
> Or if you create 2 MLT branches this is acceptable in whole.

I will do that

> Plus, for release I still want to get in proper support for numeric pair fields, and I ran into a single numeric field using a simple text edit instead of SpinBox. Should I work on those?

I have been working on 2d and 3d types for size and position types. I have it working but I still need to sort out absolute vs canonical vs normalized values. I have been storing 2d and 3d values in a rect structure. But I wonder if we should just create new MLT types for 2d and 3d? Or types for point and size? I am open to discussion and happy to collaborate. I was just going to throw something out there and then let us debate it. my testing has been  focused on "CropOfx".

bmatherly:
I think this is safe to merge with our without the two MLT changes (Nuke extensions and animation). It will still work the old way without those and we would still get the channel mask and description scrolling changes.

I'll try to break things up more in the future. It just that every plugin I want to improve is a rabbit hole of issues (as you have also been finding).

ddennedy:
> I have been storing 2d and 3d values in a rect structure. But I wonder if we should just create new MLT types for 2d and 3d? Or types for point and size?

Currently, the OpenFX code in MLT is using property lists for these tuples. At first, I thought OK it seems to be working, and Shotcut can then manage space-delimited strings of 2 numbers. However, upon reflection, this cannot realistically support animation and keyframes. If it is possible today, then is it not obvious how with respect to (de)serialization, and it lacks API like we have for mlt_rect. We should use mlt_rect for these; there is no need to make new types. Not all uses of mlt_rect utilize the opacity field in mlt_rect. It is fine for any sort of numeric tuple <= 5 items except I think there is no explicit signal about subsets of the type. Something like Shotcut UI generate needs to use a hint like the `widget` in metadata.

ddennedy:
> Make the filter description area scrollable with capped height for long descriptions.

Not a deal-breaker, but there is no scroll bar or some other indication. Consider my idea since a scroll bar is kind of ugly (esp. nested scroll bars), or with no tooltip and see the Help button. BlurCImg is interesting: the first line is brief and there is a new line before the lengthy description. I would rather not see an ellipses in that case, only the first line and the rest from Help.

ddennedy:
> Use Shotcut ChannelMask control for ladspa channel_mask param

It should be for vst2 as well. Can it match on audio and "channel_mask"?

bmatherly:
> We should use mlt_rect for these; there is no need to make new types. Not all uses of mlt_rect utilize the opacity field in mlt_rect. It is fine for any sort of numeric tuple <= 5 items except I think there is no explicit signal about subsets of the type. Something like Shotcut UI generate needs to use a hint like the widget in metadata.

Thanks for validating my approach. I have a working copy in test that maps 2D and 3D types to mlt_rect and uses widget hints to inform size vs position widgets. It is working - but i haven't got normalization working yet.

> Not a deal-breaker, but there is no scroll bar or some other indication. Consider my idea since a scroll bar is kind of ugly (esp. nested scroll bars), or with no tooltip and see the Help button. BlurCImg is interesting: the first line is brief and there is a new line before the lengthy description. I would rather not see an ellipses in that case, only the first line and the rest from Help.

I tried very hard but was not able to get a scrollbar to appear - even AI couldn't figure it out. I could try to add a heuristic to try to find a way to take the first part of the description. For example, only display the first sentence and show the rest in a tooltip.

> It should be for vst2 as well. Can it match on audio and "channel_mask"?

I don't have vst2 installed. Is it a prefix on the service name? Could we add it to the prefix logic in this function?

```
static bool isLadspaChannelMaskParameter(const AddOnFilterDescriptor &descriptor,
                                         const AddOnParameterDescriptor &parameter)
{
    return descriptor.service.startsWith(QStringLiteral("ladspa"), Qt::CaseInsensitive)
           && parameter.name.trimmed().toLower() == QStringLiteral("channel_mask");
}
```

src/controllers/addonqmlgenerator.cpp
```diff
@@ -83,6 +83,18 @@ static bool isGroupHeadingParameter(const AddOnParameterDescriptor &parameter)
...
+static bool hasDividerLayoutHint(const AddOnParameterDescriptor &parameter)
+{
+    return parameter.layoutHint.trimmed().toLower() == QStringLiteral("divider");
+}
+
+static bool isLadspaChannelMaskParameter(const AddOnFilterDescriptor &descriptor,
+                                         const AddOnParameterDescriptor &parameter)
+{
+    return descriptor.service.startsWith(QStringLiteral("ladspa"), Qt::CaseInsensitive)
+           && parameter.name.trimmed().toLower() == QStringLiteral("channel_mask");
```
ddennedy:
```suggestion
static bool isChannelMaskParameter(const AddOnFilterDescriptor &descriptor,
                                   const AddOnParameterDescriptor &parameter)
{
    return (descriptor.service.startsWith(QStringLiteral("ladspa."), Qt::CaseSensitive)
           || descriptor.service.startsWith(QStringLiteral("vst2."), Qt::CaseSensitive))
           && parameter.name.trimmed().toLower() == QStringLiteral("channel_mask");
```

src/controllers/addonqmlgenerator.cpp
```diff
@@ -175,11 +187,14 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
         const QString editorId = parameterId + QStringLiteral("_editor");
         const QString keyframesId = parameterId + QStringLiteral("_keyframes");
         const QString nameLiteral = quotedJsString(parameter.name);
+        const bool useLadspaChannelMask = isLadspaChannelMaskParameter(descriptor, parameter);
```
ddennedy:
```suggestion
        const bool useChannelMask = isChannelMaskParameter(descriptor, parameter);
```

src/controllers/addonqmlgenerator.cpp
```diff
@@ -175,11 +187,14 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
...
         if (supportsGeneratedKeyframes) {
             quotedKeyframeProperties << nameLiteral;
             quotedKeyframeMapEntries << QStringLiteral("%1: true").arg(nameLiteral);
         }
-        if (parameter.isReadOnly) {
+        if (useLadspaChannelMask) {
```
ddennedy:
```suggestion
        if (useChannelMask) {
```

src/controllers/addonqmlgenerator.cpp
```diff
@@ -438,12 +453,34 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
            "        columns: 4\n"
            "        columnSpacing: 8\n"
            "        rowSpacing: 6\n\n"
-           "        Label {\n"
```
ddennedy, line 441: Label has an `elide` property

PR #1849, opened 2026-06-11
Add Number2D QML control type

bmatherly:
This enables size/position controls for OFX addon filters

Tested with CropOFX

This shows some progress with supporting 2D types. There are still some issues when testing with CropOFX
* The size is correctly initialized to the project resolution. But when you click on the "default" button, it goes to 1x1 because that is the default signaled by the plugin.
* In Shotcut, you have to click "Black Outside" to see the effect. But in Natron, you do not have to - it just works
* The format control is not hidden/disabled when the "Extent" is not "Format" - in Natron it is by way of the "secret" property
* VUI does not work yet.

Depends on https://github.com/mltframework/mlt/pull/1257

### DIFF
--- a/src/controllers/addonqmlgenerator.cpp
+++ b/src/controllers/addonqmlgenerator.cpp
@@ -125,6 +125,7 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
     QStringList quotedDescriptions;
     QStringList quotedValueLists;
     QStringList quotedNormalizedCoordinates;
+    QStringList quotedNormalizedDefault;
     QStringList quotedKeyframeProperties;
     QStringList quotedKeyframeMapEntries;
     QStringList setControlsLines;
@@ -164,6 +165,10 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
         }
         if (parameter.normalizedCoordinates) {
             quotedNormalizedCoordinates
+                << QStringLiteral("%1: 'yes'").arg(quotedJsString(parameter.name));
+        }
+        if (parameter.normalizedDefault) {
+            quotedNormalizedDefault
                 << QStringLiteral("%1: 'yes'").arg(quotedJsString(parameter.name));
         }
         if (!parameter.description.isEmpty()) {
@@ -290,6 +295,9 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
         << "}\n\n"
            "    property var propertyNormalizedCoordinates: {"
         << quotedNormalizedCoordinates.join(QStringLiteral(", "))
+        << "}\n\n"
+           "    property var propertyNormalizedDefault: {"
+        << quotedNormalizedDefault.join(QStringLiteral(", "))
         << "}\n\n"
            "    keyframableParameters: ["
         << quotedKeyframeProperties.join(QStringLiteral(", "))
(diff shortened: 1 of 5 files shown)


src/controllers/addonqmlgenerator.cpp
```diff
@@ -434,10 +450,27 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
...
+           "                if (d !== undefined && d !== null && d !== '') {\n"
+           "                    if (type === 'rect' && root.propertyNormalizedCoordinates\n"
+           "                            && root.propertyNormalizedCoordinates[p] === 'yes') {\n"
+           "                        var _np = String(d).trim().split(/\\s+/);\n"
+           "                        var _nx = _np.length > 0 ? parseFloat(_np[0]) : 0;\n"
+           "                        var _ny = _np.length > 1 ? parseFloat(_np[1]) : 0;\n"
+           "                        filter.set(p, (isNaN(_nx) ? 0 : _nx * profile.width) + ' ' + "
+           "(isNaN(_ny) ? 0 : _ny * profile.height) + ' 0 0 1');\n"
+           "                    } else if (type === 'rect' && widget === 'size') {\n"
+           "                        // OFX plugins that use kParamDoubleTypeXY (size/extent)\n"
```
bmatherly: This is an unfortunate workaround. I have not figured out how Natron knows when to use normalized dimensions and when not to. The CropOFX filter only works if we use absolute dimensions - but it isn't signaling that?

ddennedy:
> But when you click on the "default" button, it goes to 1x1 because that is the default signaled by the plugin.

That might be normalized?

> "Black Outside" to see the effect. But in Natron, you do not have to - it just works

That sounds like Natron is doing some automatic alpha visualization or implicit composing with a black background. This is not important to address now.

> The format control is not hidden

low priority

> VUI does not work yet

That is not surprising since it is not generated. It seems a big effort to map our rectangle control to this combination of params.

For OpenFX, I do not expect a generated UI to be as good as an authored one without months of continual improvements. I just want the experimentation to be less mysterious in the near term. I added the ability to load  filter UIs from AppData so if someone running `--experimental` vets an add-on they can share the UI with others. They can copy the generated one and manually refine it.  When it becomes refined, proven, and popular it can be added to the Shotcut source for everyone out-of-the-box and translated. So, 3 levels if you see what I mean: experimental -> shared -> official. Even for external ones like I have done for NTSC-rs and Supermassive (VST2). Eventually, there might be an extension installer and support for UIs embedded in the plugin.

bmatherly:
I agree with all our comments. I intend to keep iterating on these plugin UIs - using Natron as a host implementation reference. I wouldn't mind if this PR were merged as-is and then I can keep making improvements over time.

src/controllers/addonqmlgenerator.cpp
bmatherly: I implemented the workaround in MLT for the specific plugin.

bmatherly:
I think it would be good to merge this and the MLT PR  before the test release.

src/controllers/addonqmlgenerator.cpp
```diff
@@ -451,19 +459,16 @@ bool AddOnQmlGenerator::generate(const AddOnFilterDescriptor &descriptor,
...
-           "                        // OFX plugins that use kParamDoubleTypeXY (size/extent)\n"
-           "                        // but do not declare kOfxParamPropDefaultCoordinateSystem\n"
-           "                        // (e.g. CropOFX in filter context). Default to full frame.\n"
-           "                        filter.set(p, profile.width + ' ' + profile.height + ' 0 0 "
-           "1');\n"
+           "                    if (type === 'rect' && root.propertyNormalizedDefault\n"
+           "                            && root.propertyNormalizedDefault[p] === 'yes') {\n"
+           "                        var _dp = String(d).trim().split(/\\s+/);\n"
+           "                        var _dx = _dp.length > 0 ? parseFloat(_dp[0]) : 0;\n"
+           "                        var _dy = _dp.length > 1 ? parseFloat(_dp[1]) : 0;\n"
```
ddennedy, line 466: I am not sure you are aware of this, but JavaScript `let` and `const` are preferred in new/changed QML/JS code:
https://github.com/mltframework/shotcut/blob/master/.github/instructions/qml.instructions.md

PR #1854, opened 2026-06-27
Fix Motion Tracker keyframes not applying after save/reload

skuznetsov:
## Problem
After saving and reopening a project, **Load Keyframes from Motion Tracker** silently applies zero keyframes — the masked rectangle stays static instead of following the tracked motion.

## Root cause
`MotionTrackerModel::trackingData()` parses each keyframe's time key with `QString::toInt`, which only accepts integer frame numbers (e.g. `5`). When a project is saved, Shotcut serializes animated properties — including the `opencv.tracker` `results` — in clock time format (e.g. `00:00:00.167`). After a save + reload, `toInt` then fails on every entry, so `trackingData` returns an empty list and `applyTracking` lays down no keyframes.

This is why tracking "works" right after Analyze (in-memory `results` are frame-numbered) but stops working once the project has been saved and reopened.

## Fix
Accept both frame-number and clock/timecode time keys. The parsed frame value is not used downstream (only the rectangles are consumed by `applyTracking`/`reset`), so fall back to the running index when `toInt` fails.

## Verification
On a real project whose tracker `results` were stored as clock timecodes, the parser yielded **0** rectangles before the change and all **46** after, and the masked filter's keyframes are applied across the whole clip as expected.

### DIFF
--- a/src/models/motiontrackermodel.cpp
+++ b/src/models/motiontrackermodel.cpp
@@ -270,12 +270,18 @@ QList<MotionTrackerModel::TrackingItem> MotionTrackerModel::trackingData(const Q
     for (const auto &i : l) {
         auto pair = i.split("~=");
         if (pair.size() == 2) {
-            auto frame = pair.at(0).toInt(&ok);
+            // The keyframe time may be serialized either as a frame number
+            // ("5") or as a clock value ("00:00:00.167"), depending on the MLT
+            // time format used when the project was written. toInt() only
+            // parses the frame-number form; for the clock form it fails, so we
+            // fall back to the running index. Only the rectangles are consumed
+            // downstream, so the exact frame value is not significant here.
+            int frame = pair.at(0).toInt(&ok);
+            if (!ok)
+                frame = int(result.size());
             props.set("", pair.at(1).toLatin1().constData());
             auto rect = props.get_rect("");
-            if (ok) {
-                result << TrackingItem{frame, QRectF(rect.x, rect.y, rect.w, rect.h)};
-            }
+            result << TrackingItem{frame, QRectF(rect.x, rect.y, rect.w, rect.h)};
         }
     }
     return result;
--- a/src/models/motiontrackermodel.cpp
+++ b/src/models/motiontrackermodel.cpp
@@ -264,21 +264,19 @@ QList<MotionTrackerModel::TrackingItem> MotionTrackerModel::trackingData(const Q
     QList<TrackingItem> result;
     auto s = m_data.value(key, {}).trackingData;
     auto l = s.split(';');
-    bool ok{false};
     Mlt::Properties props;
+    auto consumer = MLT.consumer();
 
     for (const auto &i : l) {
         auto pair = i.split("~=");
         if (pair.size() == 2) {
-            // The keyframe time may be serialized either as a frame number
-            // ("5") or as a clock value ("00:00:00.167"), depending on the MLT
-            // time format used when the project was written. toInt() only
-            // parses the frame-number form; for the clock form it fails, so we
-            // fall back to the running index. Only the rectangles are consumed
-            // downstream, so the exact frame value is not significant here.
-            int frame = pair.at(0).toInt(&ok);
-            if (!ok)
-                frame = int(result.size());
+            // The keyframe time is serialized either as a frame number ("5") or
+            // as a clock value ("00:00:00.167"), depending on the time format
+            // MLT used when the project was written (animated properties are
+            // saved as time so they adapt to a frame-rate change). time_to_frames()
+            // parses both forms to the correct frame using the project frame rate.
+            int frame = consumer ? consumer->time_to_frames(pair.at(0).toUtf8().constData())
+                                 : pair.at(0).toInt();
             props.set("", pair.at(1).toLatin1().constData());
             auto rect = props.get_rect("");
             result << TrackingItem{frame, QRectF(rect.x, rect.y, rect.w, rect.h)};


src/models/motiontrackermodel.cpp
```diff
@@ -270,12 +270,18 @@ QList<MotionTrackerModel::TrackingItem> MotionTrackerModel::trackingData(const Q
...
-            auto frame = pair.at(0).toInt(&ok);
+            // The keyframe time may be serialized either as a frame number
+            // ("5") or as a clock value ("00:00:00.167"), depending on the MLT
+            // time format used when the project was written. toInt() only
+            // parses the frame-number form; for the clock form it fails, so we
+            // fall back to the running index. Only the rectangles are consumed
+            // downstream, so the exact frame value is not significant here.
+            int frame = pair.at(0).toInt(&ok);
+            if (!ok)
+                frame = int(result.size());
```
ddennedy: You should provide clear steps to reproduce the bug since I did not reproduce it. I think there was a change in the past year or so to ensure property animations would save as time values instead of frame numbers to ensure they adapt to changing the project video mode's frame rate. So, it would not surprise me if this bug occurred as a result. However, I must be able to confirm a fix.

I added a debug log line inside this method to log each pair. The MLT XML contains
```xml
<property name="results">00:00:00.000~=1333 671 192 108 0;00:00:00.167~=1334 669 192 108 0;00:00:00.334~=1332 664 196 110 0;00:00:00.501~=1335 661 192 108 0;00:00:00.667~=1335 657 196 110 0;00:00:00.834~=1338 655 192 108 0;00:00:01.001~=1337 652 196 110 0;00:00:01.168~=1339 649 196 110 0;00:00:01.335~=1342 649 192 108 0;00:00:01.502~=1342 646 196 110 0;00:00:01.668~=1343 646 196 110 0;00:00:01.835~=1342 646 196 110 0;00:00:02.002~=1345 647 192 108 0;00:00:02.169~=1345 648 192 108 0;00:00:02.336~=1345 650 192 108 0;00:00:02.503~=1344 650 192 108 0;00:00:02.669~=1343 652 192 108 0;00:00:02.836~=1342 653 192 108 0;00:00:03.003~=1340 654 192 108 0;00:00:03.170~=1339 654 192 108 0;00:00:03.203~=1339 654 192 108 0</property>
```
And the debug log shows:
```
[Debug  ] <MotionTrackerModel::trackingData> QList("0", "1333 671 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("5", "1334 669 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("10", "1332 664 196 110 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("15", "1335 661 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("20", "1335 657 196 110 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("25", "1338 655 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("30", "1337 652 196 110 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("35", "1339 649 196 110 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("40", "1342 649 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("45", "1342 646 196 110 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("50", "1343 646 196 110 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("55", "1342 646 196 110 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("60", "1345 647 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("65", "1345 648 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("70", "1345 650 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("75", "1344 650 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("80", "1343 652 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("85", "1342 653 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("90", "1340 654 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("95", "1339 654 192 108 0") 
[Debug  ] <MotionTrackerModel::trackingData> QList("96", "1339 654 192 108 0") 
```
The first pair item is already a numeric string. It appears something already converted them from time values. 

Secondly, your change is wrong; it cannot simply use the index as the frame number because it currently makes a keyframe every 5 frames. It needs to use `MLT->consumer()->time_to_frames()`.

ddennedy (changes requested):

src/models/motiontrackermodel.cpp
skuznetsov: Thanks for the review.
 Reproduction (it only triggers once the results are read back as clock time, which is what happens here after a save + reload):

1. Add a Motion Tracker filter to a clip, set the rect, and Analyze.
2. Add a Mask: Simple Shape filter and use "Load Keyframes from Motion Tracker" - this works.
3. Save the project, close it, and reopen it.
4. Select the mask and "Load Keyframes from Motion Tracker" again → zero keyframes are applied (the masked rect stays static).

On reopen the saved results are in clock form, e.g. `00:00:00.000~=789 457 154 13 0;00:00:00.167~=...`, and trackingData() receives those clock strings. The original code did `pair.at(0).toInt(&ok)` and only appended the item `if (ok)`. `toInt()` fails on **00:00:00.167**, so every entry was skipped and trackingData() returned an empty list, so no keyframes applied. That `if (ok)` gate is the real bug.

ddennedy: Those are the steps I am doing but not reproducing it.

ddennedy: image

ddennedy:
Thank you. Even though I did not reproduce it, I do believe there are ways it can get to that part of the code with time values and this is a good, valuable fix that does not regress.

skuznetsov:
My apologies for not answering earlier.
I found it hard way, when I downloaded ShotCut, and could not understand why the positions are not saving. So I had to clone the repo and go through the code with a debugger to see what is going on.

PR #1856, opened 2026-06-27
[codex] Add file rename action

vitaly-zdanevich:
- enable Rename only when exactly one file or folder is selected across all file view modes
- reject path separators, rename in place, and reselect the renamed item after the filesystem model updates

image

### DIFF
--- a/src/docks/filesdock.cpp
+++ b/src/docks/filesdock.cpp
@@ -890,6 +890,14 @@ void FilesDock::setupActions()
     });
     Actions.add("filesUpdateThumbnailsAction", action);
 
+    action = new QAction(tr("Rename"), this);
+    action->setEnabled(false);
+    connect(action, &QAction::triggered, this, &FilesDock::onRenameActionTriggered);
+    connect(m_selectionModel, &QItemSelectionModel::selectionChanged, action, [=]() {
+        action->setEnabled(selectedFileIndex().isValid());
+    });
+    Actions.add("filesRenameAction", action);
+
     action = new QAction(tr("Select All"), this);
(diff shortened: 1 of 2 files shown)


ddennedy:
Files does not have Delete, Copy, Paste, Move or any change-related functions for various reasons:
- It is easier to say "not Shotcut's fault" because Shotcut generally does not do general file modifications anywhere except for the files it makes.
- These things need Undo and Redo.
- People expect these to handle multiple selection (except perhaps Rename).
- Combinations of Undo and multiple is very tricky and require progress handling, adding more work, undoing and redoing batches,.
- On Windows, if the file is open in Shotcut it is locked, and this new feature will fail leading to confusion. You do not care about Windows, but the project maintainers and directly handle the user complaints and problem reports, not you.

src/docks/filesdock.cpp
```diff
@@ -1255,6 +1284,57 @@ void FilesDock::onSelectAllActionTriggered()
...
+        return;
+
+    // Keep rename scoped to the current directory instead of moving the item.
+    if (newName.contains(QLatin1Char('/')) || newName.contains(QLatin1Char('\\'))) {
+        QMessageBox::warning(this, tr("Rename"), tr("A file name cannot contain path separators."));
+        return;
+    }
+
+    QDir dir(info.absolutePath());
+    if (!dir.rename(oldName, newName)) {
```
ddennedy, line 1319: This must be inside a class derived from `QUndoCommand` and pushed onto `MainWindow::undoStack()`.

ddennedy (changes requested):

ddennedy:
It also needs to update an opened project that is using the file being renamed.

PR #1868, opened 2026-07-19
Fix magnet snap leaving a gap at low timeline zoom

Proggerdogger:
## Summary
At low timeline zoom, magnet snap can adjust a clip by less than one pixel while changing its frame. The drop path only committed moves when `abs(pixelDelta) >= 1`, so a one-frame gap could remain after an apparent snap.

This snaps neighboring clip edges in frame space and commits the drop when the rounded frame changes.

## Test plan
- [x] Zoom timeline out (`scaleFactor` around 0.2), magnet on, close a one-frame gap between neighboring clips on the same track; after drop they are frame-exact contiguous (no blank).
- [ ] Magnet off: existing non-snap drag/drop behavior unchanged.
- [ ] Playhead/marker snaps still work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,14 @@ The fastest way to build and try Shotcut development version is through [Qt Crea
 
 First, check dependencies are satisfied and various paths are correctly set to find different libraries and include files (Qt, MLT, frei0r and so forth).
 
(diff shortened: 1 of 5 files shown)


Proggerdogger:
Splitting into focused PRs: snap fix and preflight separately.

PR #1869, opened 2026-07-19
Fix magnet snap leaving a gap at low timeline zoom

Proggerdogger:
## Summary
At low timeline zoom, magnet snap can adjust a clip by less than one pixel while changing its frame. The drop path only committed moves when the pixel delta was at least 1, so a one-frame gap could remain after an apparent snap.

This snaps neighboring clip edges in frame space and commits the drop when the rounded frame changes.

## Test plan
- [x] Zoom timeline out (around `scaleFactor` 0.2), magnet on, close a one-frame gap between neighboring clips on the same track; after drop they are frame-exact contiguous (no blank).
- [ ] Magnet off: existing non-snap drag/drop behavior unchanged.
- [ ] Playhead/marker snaps still work.

### DIFF
--- a/src/qml/views/timeline/Clip.qml
+++ b/src/qml/views/timeline/Clip.qml
@@ -259,7 +259,11 @@ Rectangle {
                 doubleClickTimer.isFirstRelease = false;
                 parent.y = 0;
                 var delta = parent.x - startX;
-                if (Math.abs(delta) >= 1 || trackIndex !== originalTrackIndex) {
+                // At low zoom, magnet snap may move by <1px while changing the
+                // frame. Commit when the frame changes, not only when abs(delta)>=1.
+                var startFrame = Math.round(startX / multitrack.scaleFactor);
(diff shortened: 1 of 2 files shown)


ddennedy:
Thank you. I reproduced the bug, and this fixes it.

Proggerdogger:
Happy to help :)

ddennedy:
I am sorry to report that I have reverted this because it caused a much worse regression: at the default zoom level (and around it) clicking a clip to select it creates a 2-frame transition right of the clip and shifts all clips on the track following the transition increasing the track (sometimes total) duration by 2 frames. Zoomed out a few steps from default, the damage can be difficult to see.

Proggerdogger:
No worries thanks for telling me.

Proggerdogger:
Problem is fixed here- https://github.com/mltframework/shotcut/pull/1874

Also, it doesn't pass all checks, but that's just because of the github branch I was using, there's no problems with the code.

PR #1870, opened 2026-07-19
Add Debian/Ubuntu preflight dependency check for build-shotcut.sh

Proggerdogger:
## Summary
`scripts/build-shotcut.sh` can fail late in a long build when a host tool is missing. This adds `scripts/preflight-deps.sh` for Debian/Ubuntu-family systems to report all missing packages in one sorted `sudo apt install` command, and runs it early from `build-shotcut.sh` by default.

- Exit `1` aborts the build with the install hint
- Exit `2` (unsupported OS) warns and continues
- Disable with `ACTION_PREFLIGHT=0` in `build-shotcut.conf`
- Does not install packages or run `apt`

## Test plan
- [x] `scripts/preflight-deps.sh` exits 0 on a stock Debian/Ubuntu build host with deps present
- [x] With tools removed from `PATH`, exits 1 and prints one sorted `sudo apt install -y …` line
- [x] `ACTION_PREFLIGHT=0` skips the check
- [ ] Unsupported OS path exits 2 and build continues with a warning

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,14 @@ The fastest way to build and try Shotcut development version is through [Qt Crea
 
 First, check dependencies are satisfied and various paths are correctly set to find different libraries and include files (Qt, MLT, frei0r and so forth).
 
+If you build with `scripts/build-shotcut.sh`, run the Debian/Ubuntu preflight checker to verify host build tools up front:
+
+```
+scripts/preflight-deps.sh
+```
+
+`build-shotcut.sh` runs this check by default before fetching or compiling. Disable it with `ACTION_PREFLIGHT=0` in `build-shotcut.conf`.
+
 #### Configure
 
 In a new directory in which to make the build (separate from the source):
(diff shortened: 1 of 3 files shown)


ddennedy:
This is not a change I want to include because `build-shotcut.sh` is primarily used by our CI and maintained by the main developers who maintain and support the builds generated by CI .

ddennedy:
I can accept the new script and mention it in the readme.

Proggerdogger:
Sounds good. I can understand you not wanting to maintain a new script when it's not generally useful. Putting it in the readme means someone can make the install easier if they want to. It'll be interesting to see if anyone uses it.

ddennedy:
Rejected because this script is used for CI including macOS. Also, it only works with Debian-derived distributions. People developing on Linux should use a distro that has the latest release of MLT (rolling, unstable, etc.) and simply use the hard, direct dependencies mentioned in the README.

PR #1873, opened 2026-07-23
Add Timeline > Generate > Adjustment Clip

ddennedy:
This pull request adds support for "Adjustment Clips" to the timeline, allowing users to add a special type of clip that applies effects to everything below it. The feature is only available when using libmlt version 7.41.0 or newer. The implementation includes UI changes to add the new generator, logic to create and insert the adjustment clip, and support for displaying it in the main window.

**Adjustment Clip Feature:**

* Added a menu option to insert an "Adjustment Clip" in the timeline when libmlt >= 7.41.0. (`src/docks/timelinedock.cpp`, `src/docks/timelinedock.h`) [[1]](diffhunk://#diff-f68df8db1fd177eef1313a1db87917a1f55cca11cd7d8474f227a25162845a18R1487-R1493) [[2]](diffhunk://#diff-ddb896addd53739957e7ec6971138cc519b45acb3cfb84ca42cbdda5192e2085R289-R291)
* Implemented the `addAdjustmentClip()` method to create and insert a color producer configured as an adjustment clip. (`src/docks/timelinedock.cpp`)
* Updated `loadProducerWidget()` to recognize adjustment clips and set up the filter controller appropriately. (`src/mainwindow.cpp`)

### DIFF
--- a/src/docks/timelinedock.cpp
+++ b/src/docks/timelinedock.cpp
@@ -1484,6 +1484,13 @@ void TimelineDock::setupActions()
         if (mltProducers->get_data("blipflash")) {
             menu->addAction(tr("Blip Flash"), this, SLOT(addGenerator()))->setObjectName("blipflash");
         }
+#if LIBMLT_VERSION_INT >= ((7 << 16) + (41 << 8))
+        if (mltProducers->get_data("color")) {
+            menu->addSeparator();
+            menu->addAction(tr("Adjustment Clip"), this, SLOT(addGenerator()))
+                ->setObjectName("adjustment");
+        }
+#endif
         action->setMenu(menu);
     }
     Actions.add("timelineNewGenerator", action, windowTitle());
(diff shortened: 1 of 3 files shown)


bmatherly:
I think this is going to be a good feature. And implementing it as a producer from the generator menu makes the implementation simple. From a user perspective, I wonder if this would be easier to understand as an "adjustment track". Under the hood, it would just be a regular timeline track with an adjustment producer added to it. But  we could show it with a different color. Hmmm. but then users might try to add clips to the track itself instead of the producer on the track. But maybe that will happen anyway.

Alternately, we could add a convenience action in the timeline menu to create a track with an adjustment clip that fills it.

Maybe I'm over thinking it.

ddennedy:
An adjustment clip is far more flexible with respect to timing.

bmatherly:
I agree - the clip paradigm is more flexible overall. Sometimes an adjustment is needed for the entire project, and sometimes it is only needed temporarily. This PR made me think about times that I use a generator on a track and I  want the generated clip to endure for the entire duration of the timeline. But as I work, the timeline length keeps changing. So I have to keep adjusting the clip. Maybe a future convenience (not part of this PR) would be a property on a timeline generator clip that would be called "match duration to timeline". If checked, the clip would be automatically resized when the timeline duration changes.

ddennedy:
I agree it can be useful to generalize the ability to automatically adjust a clip to the entire timeline. Another example use case is a brand watermark. In that case, it is unlikely to be a generator but rather an image or glaxnimate animation. We can even add the ability to loop a video (add a checkbox to Properties that sets `eof=loop`). However, at this time I need to turn my attention to improve Timeline performance and undo.

PR #1874, opened 2026-07-24
Click is now safe

Proggerdogger:
This should fix the error I introduced with my previous fix. The problem was that by making shotcut accept <1 pixel movements, it would also take even small movements like that caused by clicking and insert additional frames and the like. This was fixed by making it so that acceptance is also based on a change being greater than a certain value of x. Feel free to check this on whatever codebase you found out this bug existed on, and see if this fixes it, I am quite confident it's good.

The build-shotcut-linux failure is not from this change. CI clones mltframework/shotcut and runs git checkout origin/<branch>, but the branch only exists on my fork, so checkout fails before compiling.


ddennedy:
Is this supposed to apply atop #1869 or replace it? I get a merge conflict if I reapply that and then merge this, but easy to resolve. However, #1869 also has changes to `Track.js`.

ddennedy:
I tested the combination of that plus this. I confirm this fixes the bug that #1869 introduced. However, there is a minor problem that it *appears* that a clip has moved when you click to select it: it often nudges slightly to the left. But when you zoom in to check details nothing changed; rather it undermines confidence causing wasted effort to check it. It is caused by `parent.x = startX;`. Do you think it is necessary?

Proggerdogger:
I could probably change things so that clicks don't appear to move the timeline at all unless you're more serious about moving them ( a real drag). I'll work on it later. Thanks for accepting the commit :)

ddennedy:
I am not sure we are on the same page. Please answer my questions.

Proggerdogger:
I think the problem is that the >x change is stopping the change from actually being committed on a click, but it still looks like it moves because it displays the change then decides whether to accept it. That change is never accepted because it isn't a large enough percentage width of the screen (the x is pixels btw). To fix that you would probably have to put in a check before displaying a change as well.

Proggerdogger:
Oh to be clear, yes it is necessary, but a new piece of code will have to be added to stop the unpleasant nudge you're talking about.

ddennedy:
I disagree that it is required. Why do you claim that it is? More code change increases the risk.

Proggerdogger:
I'll do some more testing without the x thing, but I think the problem is that without it small movements are going to cause the timeline to change.

Proggerdogger:
By the way, I will come back to this and test the things I said I would test + improve the code a bit, but it's probably going to be about two weeks.

ddennedy:
I merged it with my change that removes `parent.x = startX;`

PR #1875, opened 2026-07-24
Frame accurate snap

Proggerdogger:
On my quest to fix shotcut bugs I noticed that often snapping would not snap to the edge of a clip but slightly inside of it. I fixed this by rounding snap targets to whole frames (as every clip is a certain number of whole frames you never want to snap onto a decimal place).

## Reproduce

1. New project, magnet/snap on.
2. Put two clips on the same track with a 1-frame gap (or a project where a tiny blank sits between them).
3. Zoom the timeline all the way out (very small scale / minimum zoom).
4. Drag the right clip left until it snaps to the left clip.
5. Zoom all the way in at the join.

Before: the join is 1 frame short or overlapped, or a transition appears on drop.
After: the clips abut on the same frame (no gap, no 1-frame overlap).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/qml/views/timeline/Track.js
+++ b/src/qml/views/timeline/Track.js
@@ -36,53 +36,57 @@ function snapClip(clip, repeater) {
             // Snap to blank
             if (right > itemLeft - SNAP && right < itemLeft + SNAP) {
                 // Snap right edge to left edge.
-                clip.x = itemLeft - clip.width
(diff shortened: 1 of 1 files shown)


Proggerdogger:
Superseded by #1876 (false magnets PR already includes frame-accurate snap). Closing this duplicate Track.js change to avoid merge conflicts.

PR #1876, opened 2026-07-24
False magnets when closing a gap

Proggerdogger:
Two changes, first on my quest to fix shotcut bugs I noticed that often snapping would not snap to the edge of a clip but slightly inside of it. I fixed this by rounding snap targets to whole frames (as every clip is a certain number of whole frames you never want to snap onto a decimal place).

## Reproduce
1. New project, magnet/snap on.
2. Put two clips on the same track with a 1-frame gap (or a project where a tiny blank sits between them).
3. Zoom the timeline all the way out (very small scale / minimum zoom).
4. Drag the right clip left until it snaps to the left clip.
5. Zoom all the way in at the join.
Before: the join is 1 frame short or overlapped, or a transition appears on drop.
After: the clips abut on the same frame (no gap, no 1-frame overlap).

Second, when dragging a clip to join with another clip it feels like the join fails for no reason. I made it so that instead of stopping part way through joining to the next clip (this was due to hitting a magnet that really shouldn't be present), the join completes, whilst also keeping snapping to playhead working and joining to clips in other sections (e.g. V1 can still snap to V2 clips). I will note that if the playhead and the edge of a clip are nearby, snapping now preferences the edge of the clip which is probably fine but tell me if it creates issues.

## Reproduce

1. Same track: two clips with a 2-frame blank between them.
2. Magnet/snap on. Set zoom so about 1 frame ≈ SNAP (roughly 10 px per frame / scaleFactor ≈ 10).
3. Drag the right clip left to close the gap.

Before: the clip sticks about halfway across the gap (~0.016s on a ~0.033s gap) or snaps to the blank/playhead instead of the real join.
After: the clip joins the left clip’s edge in one magnet action.
Still works: snap to playhead when not closing a same-track gap; V1↔V2 cross-track snap when home isn’t mid-join.

Change 2 partially relies on change 1, so I am bundling these together.

The build-shotcut-linux failure is not from this change. CI clones mltframework/shotcut and runs git checkout origin/<branch>, but the branch only exists on my fork, so checkout fails before compiling.

### DIFF
--- a/src/qml/views/timeline/Track.js
+++ b/src/qml/views/timeline/Track.js
@@ -18,72 +18,161 @@
 var SNAP = 10
 var SNAP_TRIM = 4
 
+function framesOf(item) {
+    if (typeof item.clipDuration === 'number' && item.clipDuration > 0)
+        return item.clipDuration
+    return Math.round(item.width / timeScale)
+}
+
+// Exclusive end frame of an item — never prefer a value that would leave a
+// 1-frame overlap (that becomes a transition on drop).
+function exclusiveEndFrame(item) {
+    var itemLeft = item.x
+    var itemRight = itemLeft + item.width
+    var byDuration = Math.round(itemLeft / timeScale) + framesOf(item)
+    var byWidth = Math.round(itemRight / timeScale)
+    return Math.max(byDuration, byWidth)
+}
+
+// Gaps must not be magnet targets. Snapping to a blank's edge that coincides
+// with the dragged clip's current edge holds the clip still until the cursor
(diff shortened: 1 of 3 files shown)


ddennedy:
I’m not going to accept all these changes for something so minor especially after the regression. Just move the playhead out of the way.

Proggerdogger:
I will remove the playhead changes and do another commit. I will point out that you're being somewhat dishonest about "just move the playhead out of the way" as what this actually fixes is a bug that occurs every time you move two clips together at low zoom, which does get quite irritating.

PR #1877, opened 2026-07-26
Auto duck

bmatherly:
depends on https://github.com/mltframework/mlt/pull/1272

One downside is that the ducking can not look ahead at the audio. So it can't start fading until after the upper tracks have already crossed the threshold.

image

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -4429,11 +4429,10 @@ QWidget *MainWindow::loadProducerWidget(Mlt::Producer *producer)
                                  ->data(m_timelineDock->model()->index(trackIndex),
                                         MultitrackModel::IsBottomVideoRole)
                                  .toBool();
-        if (!isBottomVideo) {
-            w = new TrackPropertiesWidget(*producer, this);
-            scrollArea->setWidget(w);
-            return w;
-        }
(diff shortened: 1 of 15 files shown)


src/mainwindow.cpp
```diff
@@ -4429,11 +4429,10 @@ QWidget *MainWindow::loadProducerWidget(Mlt::Producer *producer)
                                  ->data(m_timelineDock->model()->index(trackIndex),
                                         MultitrackModel::IsBottomVideoRole)
                                  .toBool();
-        if (!isBottomVideo) {
-            w = new TrackPropertiesWidget(*producer, this);
-            scrollArea->setWidget(w);
-            return w;
-        }
+        bool isTopTrack = trackIndex == 0;
+        w = new TrackPropertiesWidget(*producer, !isBottomVideo, !isTopTrack, this);
```
ddennedy: Add a check for MLT version >= 7.41.0 or `mix` \version >=4 to allow showDucking to be true

ddennedy (changes requested):

src/mainwindow.cpp
bmatherly: done

PR #1883, opened 2026-08-03
Add track volume control and audio level indicator in timeline

bmatherly:
Track volume control is now available in each timeline track header and updates the track gain via a hidden volume filter.

A per-track audio level indicator shows the audio level for each track in the timeline UI. The indicator color mapping follows the audio meter gradient and updates in real time with incoming levels.

The the audiolevel and volume filters are hidden and are added after all the user filters. So some refactoring was necessary to keep track of normalizing filters (hidden), pre-user filters (subtitle feed - hidden), user filters (not hidden), and post-user filters (track volumen and level - hidden)

Depends on https://github.com/mltframework/mlt/pull/1278

Volume control:
image

Level indicators:
image

### DIFF
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -1013,8 +1013,9 @@ bool MultitrackModel::trimClipOutValid(int trackIndex, int clipIndex, int delta,
 
 int MultitrackModel::trackHeight() const
 {
-    int result = m_tractor ? m_tractor->get_int(kTrackHeightProperty)
-                           : Settings.timelineTrackHeight();
+    int result = (m_tractor && m_tractor->property_exists(kTrackHeightProperty))
+                     ? m_tractor->get_int(kTrackHeightProperty)
+                     : Settings.timelineTrackHeight();
     return qBound(10, result ? result : Settings.timelineTrackHeight(), 150);
 }
 
--- a/src/models/multitrackmodel.cpp
+++ b/src/models/multitrackmodel.cpp
@@ -28,6 +28,7 @@
 #include "proxymanager.h"
 #include "qmltypes/qmlmetadata.h"
 #include "settings.h"
+#include "sharedframe.h"
 #include "shotcut_mlt_properties.h"
 #include "util.h"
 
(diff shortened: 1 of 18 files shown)


ddennedy:
> Depends on https://github.com/mltframework/mlt/pull/1278

What happens when you don't have that MLT version?

src/qml/views/timeline/TrackHead.qml
```diff
@@ -167,6 +289,40 @@ Rectangle {
         RowLayout {
             spacing: 8
 
+            Item {
+                Layout.alignment: Qt.AlignVCenter
+                width: 14
+                height: 14
+                visible: trackHeadRoot.trackAudioLevelSupported
```
bmatherly: The audio level indicator is not visible if MLT does not support it.

src/models/multitrackmodel.cpp
```diff
@@ -36,9 +37,87 @@
...
+    if (linear <= 0.0)
+        return -100.0;
+
+    double level = 20.0 * log10(qMin(linear, 1.0));
+    return qRound(level * 10.0) / 10.0;
+}
+
+static bool hasTrackLevelIndicatorSupport()
+{
+    if (LIBMLT_VERSION_INT >= ((7 << 16) + (41 << 8)))
```
bmatherly, line 108: Level indicator version.

bmatherly:
> What happens when you don't have that MLT version?

The level indicator is hidden. The volume control works with any version.

ddennedy:
The volume control obstructs the function of the audio mute button. That is not longer convenient. It works on the player because the mute button inside the popup volume controls at the same screen position as the volume button. Therefore, a double-click can toggle mute. That does not work here because the buttons are not at the same place. Align them such that double-click can toggle mute.

ddennedy:
Another UI idea for you to consider: when the track height is tall enough, show the volume slider without mute toggle and full horizontal meter. Make audio track height same as a video track to accommodate this. Fall back to this current UI when the track height is short. This would go a step further in convenience and making it more like a mixing console because a little dot is more difficult to read and a click to popup the volume slider is an extra click that might needed when trying to balance multiple tracks. No graticule, but possibly make the meter segmented or draw line over itself every 5 dB.

bmatherly:
> The volume control obstructs the function of the audio mute button. Align them such that double-click can toggle mute.

Done

> Another UI idea for you to consider

That is a really great idea. Here is my first attempt:

When tracks are all short:
image


When tracks are all tall:
image


When some are tall and some are short:
image


I see that the track height is designed so that the audio waveform is always the same size. I kind of wish there were a  view where all the tracks are as short as possible but still display the inline meter and volume. I wonder what you think about that.

ddennedy:
There are still merge conflicts.



> track height is designed so that the audio waveform is always the same size

Ignore that and just make all tracks the same height. 
Regarding the inline meter and slider, the graticule (lines with numbers) requires more space. You could possibly remove it. 
Those combined and you should be able to get a somewhat short track height that displays the controls for all track types.

bmatherly:
> Ignore that and just make all tracks the same height.

I didn't do that yet. But, inspired by your comment, I removed the graticule and tweaked some default dimensions and wrapping thresholds. I think I found a good balance. But I am still open to suggestions.

Smallest size to show the inline controls (and new default):
image

smallest size possible:
image

Taller size showing the label on its own line:
image

I could still make them all the same size if we prefer that.

ddennedy:
Close, but a problem is that it is all tracks get the expanded volume control or not:
Screenshot_2026-08-08_17-52-55
So, only part of the meter shows and volume slider inaccessible on the audio track. Use shift+wheel a couple steps to try to increase it to access it now makes it further out of reach:
Screenshot_2026-08-08_17-55-39
You have to go very tall on video to be reachable on audio.

 But one little change to make the tracks the same height is much better
Screenshot_2026-08-08_17-59-07

```js
diff --git a/src/qml/views/timeline/Timeline.js b/src/qml/views/timeline/Timeline.js
index 2b6996a1..d96ee784 100644
--- a/src/qml/views/timeline/Timeline.js
+++ b/src/qml/views/timeline/Timeline.js
@@ -113,7 +113,7 @@ function acceptDrop(xml) {
 }
 
 function trackHeight(isAudio) {
-    return isAudio? Math.max(20, multitrack.trackHeight) : multitrack.trackHeight * 2
+    return multitrack.trackHeight * 2;
 }
 
 function clamp(x, minimum, maximum) {
```

ddennedy:
One more little thing is needed: a way to reset the volume adjustment to 0. I suggest double-click. on the slider itself, not the button, which correctly toggles mute.

bmatherly:
> But one little change to make the tracks the same height is much better

Done. I agree - that is way better.

> I suggest double-click. on the slider itself, not the button, which correctly toggles mute.

Done

PR #1893, opened 2026-08-15
Validate SA3D channel map bounds

acts-1631:
Shotcut can exhaust memory when rewriting a crafted MP4 containing an SA3D
box with an oversized channel count.

SA3DBox::load trusted the declared channel count and appended that many channel
map entries without checking the box payload. This patch validates extended
box sizes, avoids overflowing position arithmetic, and requires the channel
map to fit within the box before parsing it.

Verification:

- A standalone parser harness rejected a malformed 20-byte SA3D box without
  allocating unbounded memory.
- A valid one-channel SA3D box still parsed successfully.

### DIFF
--- a/src/spatialmedia/sa3d.cpp
+++ b/src/spatialmedia/sa3d.cpp
@@ -28,6 +28,7 @@
 #include <sstream>
 #include <iterator>
 #include <cmath>
+#include <limits>
 
 #include "constants.h"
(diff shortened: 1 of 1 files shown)


ddennedy:
Are you having fun driving your AI to spam projects?

This code was copied from https://github.com/VarolOkan/spatial-media. Fix it there please

PR #1894, opened 2026-08-15
Bound MP4 index table rewrites

acts-1631:
Shotcut can generate unbounded output and consume CPU when rewriting a crafted
MP4 containing an stco or co64 box with an oversized entry count.

Box::index_copy trusted the declared table count and rewrote entries without
checking the containing box. This patch validates the table header and bounds
the declared count against the available payload for both file-backed and
in-memory box contents.

Verification:

- Malformed stco and co64 tables were rejected without unbounded output.
- A valid one-entry stco table was rewritten successfully.

### DIFF
--- a/src/spatialmedia/box.cpp
+++ b/src/spatialmedia/box.cpp
@@ -23,6 +23,23 @@
 #include "constants.h"
 #include "box.h"
 
+namespace {
+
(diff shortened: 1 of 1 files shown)


ddennedy:
This code was copied from https://github.com/VarolOkan/spatial-media. Fix it there please

Issue #1898, opened 2026-08-20
Undo crashes the app in large projects


peteraba:
**OS and Version (please complete the following information):**
 - Operating System: Linux (Manjaro)
 - Shotcut Version: 26.8.1 (Installed via the built-in app store, likely via pacman under the hood.)

**Describe the bug**
Working on a compilation project with hundreds of clips on the timeline results in an immediate jump in CPU load and memory usage when using the undo button. The jump is not minor; it completely occupies the system until the app is killed. This doesn't happen every time, but more often than not.

**Steps To Reproduce**
1. Start Shotcut.
2. Import ~50 clips, add them 10x to the timeline.
3. Split one of the clips.
4. Undo.

**Expected behavior**
The two sides of the clip should be joined as quickly as they were split, or at least be comparable.

**Screenshots / Other**
No screenshots, but I do have a fix: [peteraba/shotcut/pull/1](https://github.com/peteraba/shotcut/pull/1). The reason I didn't create a Pull Request for you is that I respect your time and knowledge, and the fix was primarily done via AI. It went through multiple rounds of AI reviews, and I have also read every single line in that PR. That said, even though I've been a programmer for 25+ years, and I do understand why my changes improve the situation, I don't have the domain expertise; I don't know what existing features my changes might be destroying, hence I can't say with confidence that "I know what I'm doing".

Beyond reviews, I'm also using my changes locally, and in my experience, together with the zoom-fix, it is a game-changer for large projects.

Say the word, and I'm happy to create a PR against the main project.

**Note:** I've read in the forums that others work around this issue by splitting projects up and importing other projects; I find this quite unacceptable. 

**Note 2:** I'm not claiming that my changes solve this problem completely; there may be many cases not covered, but it's likely a decent first step.

ddennedy:
Hi Peter, thanks for helping! I have been working on this the past couple of weeks in #1897, also with help of AI. So, this is good timing. My branch also has the same caveat as your "Note 2." I was just starting to do some deeper testing of regressions on my branch before merging. This involves going through the git log for the timeline commands and undohelper to look at fixes and hopefully find the reproduction steps.
Are you available to help compare our branches while I review and test your zoom improvements? I leave that open-ended, but some ideas:
- which one solves a gap in the other (e.g. from a brief scan, I think yours improves moving clips while mine does not)
- compare high-level approaches
- simply test my branch and give a rough high level comparison of which has better speed and coverage

[...]

PR #1901, opened 2026-08-22
Undo reliability fix

peteraba:
This is a PR for https://github.com/mltframework/shotcut/issues/1898.

**Note:** This is likely a partial duplicate of #1897 by @ddennedy . The two change sets are to be compared to find the best and most complete solution.

## Summary

On large timelines, several undo paths walked or rebuilt whole tracks through `UndoHelper`. That froze the UI (OS “force quit”) or crashed the app, even when CPU and RAM stayed modest.

Those operations now reverse directly instead of snapshot-diffing the playlist:

- **Split** — restore the original clip from stored XML (`joinClips` dropped keyframes and the original UUID)
- **Same-track move** — lift from the new position, overwrite at the original start
- **Cross-track move** — same reverse, back onto the source track; dest clips overwritten by the drop are snapshotted and put back
- **Transition grow by dragging a clip** — apply the opposite trim (`-delta`) on the existing mix. XML-restoring a mix leaves `mix_in`/`mix_out` pointing at freed cuts and the consumer segfaults in `transition_get_frame`
- **Insert / paste** — `removeClip` the stamped UUID
- **Overwrite / drop onto a track** — lift the stamped UUID and restore only the dest clips that occupied that span

Remaining `UndoHelper` snapshots are track-scoped, unused clip XML is skipped, split halves get distinct UUIDs, restore XML is parsed before any track is wiped, and cached waveforms / elided thumbnails no longer storm the GUI thread.

## Test plan

Verified on a large project (~8h, ~1000 clips on the main video track):

- [x] Split a clip, undo — immediate, clip is whole again
- [x] Split into 4, open/close gaps, rearrange pieces, ripple-delete empty space, undo through the stack
- [x] Drag a clip to another track, undo / redo
- [x] Drag a clip to grow a transition, undo
- [x] Cut and paste clips, undo — no hang
- [x] Drop a clip onto an almost-empty track, undo
- [x] Drop a clip onto a busy track (including on top of existing clips), undo / redo — dest clips come back, no hang

Also previously verified on a small project: lift, same-track move, trim, filters, and split.

Not included: local CMake/qdoc kit tweaks for the Linux debug build.

<h3>Greptile Summary</h3>

The PR replaces expensive timeline-wide undo snapshots with targeted inverse operations and track-scoped restoration.
- Restores overwritten destination clips and reconnects neighboring transitions.
- Restores split clips from saved XML to preserve clip state and identity.
- Moves waveform cache processing off the GUI thread and reduces redundant thumbnail work.
- Limits remaining `UndoHelper` snapshots to affected tracks and necessary clip XML.

<h3>Confidence Score: 5/5</h3>

The PR appears safe to merge.

No blocking failure remains.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| src/commands/timelinecommands.cpp | Replaces broad undo snapshots with targeted inverse operations for split, insert, overwrite, move, trim, and transition paths; the previously reported overwrite and transition issues are addressed. |
| src/commands/timelinecommands.h | Adds the saved clip state and helper declarations required by targeted timeline undo. |
| src/commands/undohelper.cpp | Adds track restrictions, selective XML restoration, and explicit reconstruction of transition references. |
| src/commands/undohelper.h | Exposes track-scoping and transition-repair interfaces used by the optimized commands. |
| src/models/multitrackmodel.cpp | Updates clip identity and timeline refresh behavior supporting direct undo restoration. |
| src/docks/timelinedock.cpp | Adjusts timeline task scheduling and refresh paths to avoid unnecessary GUI-thread work. |
| src/qml/views/timeline/Clip.qml | Reduces redundant thumbnail and waveform work for elided timeline clips. |
| src/commands/filtercommands.cpp | Replaces producer validity assertions with logged, recoverable command failures. |
| MinimalMediaBackend/CMakeLists.txt | Copies the minimal multimedia plugin into the build-tree plugin location needed by local execution. |

</details>


MinimalMediaBackend/CMakeLists.txt
```diff
@@ -17,3 +17,15 @@ target_link_libraries(minimalmediaplugin PRIVATE
   Qt6::Multimedia
   Qt6::MultimediaPrivate
 )
+
+# Running from the build tree (build/cc-debug/src/shotcut) adds
+# <build>/lib/qt6 to QT_PLUGIN_PATH. Official installs copy this plugin
+# there; without it QT_MEDIA_BACKEND=minimal cannot create QVideoSink
```
ddennedy, line 23: I do not want to include this. It is not normal that you need it. Is that because you are using the cmake preset or build script? You should probably turn it off locally. It is only to deal with Qt SDK binary linking to a different FFmpeg than the CI build makes.

ddennedy (commented):
I asked Claude Opus:

> Compare and contrast the changes made for this branch (my `optimized-timeline-undo)  with the branch proposed here: https://github.com/mltframework/shotcut/pull/1901/changes
Which one should be used as a starting point to merge them based on best price (risk) / performance ratio? And outline which has an improvement for an undocommand that the other does not.

---

## The two philosophies

**This branch (`optimizied-timeline-undo`, the #1897 lineage) — optimize *inside* `UndoHelper`.**
Commands stay almost unchanged; they just call `setHints()`, `storeXmlForClip()`, and pass a track scope to `recordBeforeState()`. One generic engine does the work:
- Three strategies: `RestoreTracks` (whole-track XML rebuild), `NoHints` (fine-grained diff + reconcile with per-clip XML), `SkipXML` (resize-only).
- A **self-healing fallback**: survivingClipsInOrder() validates the reconciliation's core assumption before mutating, and drops to the always-correct whole-track restore if it doesn't hold.
- The threading fix (parse snapshot XML off the GUI thread, mutate the model on the GUI thread) and `LongUiTask` progress UX.

**PR #1901 (peteraba) — replace `UndoHelper` with *per-command inverse operations*.**
`UndoHelper` is removed entirely from `SplitCommand`, `InsertCommand`, `OverwriteCommand`, and special-cased out of `MoveClipCommand`. Each writes its own reverse:
- Split → restore original clip from its stored XML (new `ClipState`, mints a fresh UUID for the split half in `splitClip`).
- Insert/paste → `removeClip` the stamped UUID.
- Overwrite / drop → lift the stamped UUID and restore only the destination clips in the span (`Overwritten` struct).
- Transition-grow-by-drag → apply the **opposite trim (`-delta`)**, deliberately *not* XML-restoring the mix (XML restore leaves `mix_in`/`mix_out` dangling → `transition_get_frame` segfault).
- `UndoHelper` that remains gets: bitwise `Q_DECLARE_FLAGS` hints, `restrictToTrack(s)` scoping, `failUndo`/`m_undoFailed` reliability, `fixTransitions` made static/public, and mix relink.
- Plus orthogonal reliability: `filtercommands.cpp` swaps `Q_ASSERT(producer.is_valid())` for a graceful `requireProducer()` early-return, and `Clip.qml` guards thumbnail loading with `visible ? imagePath : ''`.

## Contrast on risk vs. performance

| | This branch (A) | PR #1901 (B) |
|---|---|---|
| Blast radius | Small — commands barely change | Large — bespoke inverse per command |
| Correctness floor | **Self-healing**: worst case = slow-but-correct, never corrupt | No generic net; each inverse must be exhaustively right |
| Capture cost | Still serializes whole-track XML for `NoHints` | Avoids the snapshot entirely for inverse-able commands (cheapest) |
| Peak performance | Fast (~30 ms typical); ripple-all still `RestoreTracks` | Potentially fastest |
| Evidence of edge-case density | One shared, guarded path | 14 commits of incremental crash/data-loss fixes |

## Recommendation: start from **this branch (A)**, cherry-pick from B

Best price(risk)/performance ratio is **A as the foundation**, because its generic engine plus the self-healing fallback gives you *correctness by construction* — the worst outcome degrades to the old slow path, never to a corrupted timeline. B reaches a slightly higher performance ceiling but pays for it with many hand-written inverses and no safety net, which is exactly where undo historically crashed (transition-grow segfault, `mix_in`/`mix_out`, split identity). Merging B wholesale would import that risk you don't need, since A is already &quot;fast enough&quot; for the common cases.

Then port the **isolated, low-risk wins** from B on top:
- `filtercommands.cpp` `requireProducer()` — pure reliability, fully orthogonal, take it as-is.
- `splitClip` UUID minting in `multitrackmodel.cpp` — small identity fix.
- `Clip.qml` thumbnail visibility guard — trivial GUI-thread relief.
- The transition-grow &quot;opposite trim, not XML restore&quot; insight — adopt the *principle* (A currently routes `MoveClip` through `RestoreTracks`, so verify whether A already sidesteps that segfault).
- Adopt B's bespoke inverses (Split/Insert/Overwrite) **only where profiling shows A's snapshot-capture cost still hurts**, keeping A's fallback underneath them.

## The undo-command improvement each has that the other lacks

- **#1901 has, this branch does not:** graceful failure in the **Filter undo commands** — `Filter::AddCommand`, `RemoveCommand`, `MoveCommand`, `DisableCommand`, `PasteCommand`, `UndoParameterCommand` all replace `Q_ASSERT(producer.is_valid())` with `requireProducer()` that logs and returns instead of aborting. This branch still has the raw `Q_ASSERT`s at `filtercommands.cpp:123` (12 of them), so a missing producer during filter undo crashes here but not in #1901.
- **This branch has, #1901 does not:** fine-grained / self-healing undo for `AddTransitionCommand`, `DetachAudioCommand`, `ReplaceCommand`, and `MergeCommand` — plus the generic order-validation fallback that protects *every* command using `UndoHelper`. #1901 optimizes Split/Move/Insert/Overwrite/transition-grow but leaves these on the heavier path and has no equivalent universal safety net.

---

Also, I asked about `MoveClipCommand`, and it said there is a big improvement but also high risk of regression and recommends some merging of the two.

ddennedy:
Closing as too risky to take in whole (and found some bugs). Good ideas merged into #1897 with co-authorship. That PR is merging into master later today.

PR #1905, opened 2026-08-29
Add Audio Mixing Assistant

bmatherly:
Add an Audio Mixing Assistant tool.

Looking for feedback on this proof of concept. I'm happy to abandon it, or go a wildly different direction if that is where the discussion goes.

The purpose of this tool is to give the user a jump start to an audio mix on their timeline. It also implements audio ducking as keyframes on a gain filter if they choose to accept that recommendation. 

The tool is launched from the Edit menu in the timeline.
image

This opens a dialog. To start, the user assigns an audio role to each track: dialog, background music, ambiance, etc. They also specify the target program loudness for the main mix.
image

The tool scans the audio on all the tracks.
image

Based on the track audio role, the tool recommends filters to be applied to each track.
image

Here is the full list of possible recommendations:

**Dialogue**
- Adjust gain to match the target loudness level.
- Remove low-frequency noise to improve clarity.
- Even out volume fluctuations for consistent intelligibility.

**Background Music**
- Set base level to sit beneath other audio elements.
- Lower volume when dialogue is present.
- Cut mid-range frequencies to make space for dialogue.

**Ambience**
- Set a subtle level that supports without competing.
- Remove low-frequency rumble for a cleaner background.
- Soften high frequencies to reduce harshness.

**Sound Effects**
- Balance impact while keeping dialogue and music clear.
- Eliminate low-frequency rumble to prevent masking.
- Lower volume when dialogue is active for clarity.

**Premixed Program**
- Adjust gain to match the target loudness level.

The user can read each recommendation, and then select the ones they want to apply.  Clicking Apply will cause the filters to be added. to the tracks.

image

Even if multiple filters are created, the operation makes one undo operation
image

Filters that are created are tagged as being applied by the Assistant so the user can be reminded later why the filter is there: (The yellow box is a  tooltip popup, not a fixed banner. My mouse is not shown in this screenshot)
image

If the assistant is run a second time, any audio assistant filters are removed and new ones are added. So they do not accumulate after multiple runs.

This tool is based on simple rules and creates static filters (some use keyframes for ducking).

Ideas that I did not incorporate:
* It could also add a loudness filter on the main output to ensure the final program loudness. Maybe that would require a second pass.
* It could apply an audio loudness filter on every clip on a dialog track to ensure that each dialog clip is pre-normalized in case some are louder than others.

Please provide feedback on the concept before I dive too deep into the code review.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,6 +20,7 @@ add_executable(shotcut WIN32 MACOSX_BUNDLE
   dialogs/addencodepresetdialog.ui
   dialogs/alignaudiodialog.cpp dialogs/alignaudiodialog.h
   dialogs/alignmentarray.cpp dialogs/alignmentarray.h
+  dialogs/audiomixingassistantdialog.cpp dialogs/audiomixingassistantdialog.h
   dialogs/addonfiltersdialog.cpp dialogs/addonfiltersdialog.h
   dialogs/addonmetadatahelpdialog.cpp dialogs/addonmetadatahelpdialog.h
   dialogs/bitratedialog.h dialogs/bitratedialog.cpp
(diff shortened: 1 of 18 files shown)


ddennedy:
I will not be able to get to this for a while.

PR #1906, opened 2026-08-29
Allow scopes to request any audio format

bmatherly:
The scopes all assumed that the shared frame is rendered in S16. This change will cause the audio to be converted and cached if a format is requested that is not the rendered format.


bmatherly:
Nevermind. I see that Dan already did this
https://github.com/mltframework/shotcut/commit/e1bbcae6b98ce44da39df613239ddb92ef2b90b7

bmatherly:
Actually, I prefer my solution. I will merge the implementations.

